PR #4, opened 2017-09-13
Add close button to license dialog

peterNordin:


HopsanGUI/Dialogs/LicenseDialog.cpp
```diff
@@ -44,7 +44,6 @@ LicenseDialog::LicenseDialog(QWidget *pParent) :
     resize(640,480);
     setWindowTitle(tr("Hopsan License"));
     setAttribute(Qt::WA_DeleteOnClose, true);
-    setPalette(QPalette(QColor("gray"), QColor("whitesmoke")));
```
robbr48: Should this line really be removed?

peterNordin:
Yes, I should have made a separate commit. It was required before for some reason to make the license text visible. But I have rewritten the dialog now and it looks strange if I force that style. Now it looks OK with both native and custom hopsan style

HopsanGUI/Dialogs/LicenseDialog.cpp
peterNordin: Yes, I should have made a separate commit. It was required before for some reason to make the license text visible. But I have rewritten the dialog now and it looks strange if I force that style. Now it looks OK with both native and custom hopsan style

peterNordin: I can rebase, lets try that before merge :)

PR #5, opened 2017-09-13
Make it possible to build HopsanGUI even when QtWebKit is not available

peterNordin:
In Qt 5.6 QtWebKit has been removed and QWebEngine wont build with MinGW, so there is no longer a webview.
It might be possible to build QtWebKit manually but for now lets fall back to opening html in external browser.


peterNordin:
this is a test comment

PR #13, opened 2017-10-06
Automatic install liu

robbr48:
Silent installers shall not launch program or release notes when finished.


robbr48:
> Actually you need to add the gs path to the buildDocumentation.bat script also

Done!

PR #18, opened 2017-10-29
Add and improve error messages in library handler

peterNordin:


peterNordin:
Yes, a warning message is shown if a component typename was already present. But it does not say from what library the component comes. It should. I will add this information.

PR #1696, opened 2018-03-14
Fix initialization error in HydraulicTLMlossless

peterNordin:


componentLibraries/defaultLibrary/Hydraulic/Volumes&Lines/HydraulicTLMlossless.hpp
```diff
@@ -68,14 +68,14 @@ namespace hopsan {
             mpP2 = addPowerPort("P2", "NodeHydraulic");
 
             addInputVariable("alpha", "Low pass coefficient", "-", 0.0, &mpAlpha);
-            addInputVariable("Z_c", "Impedance", "Ns/m^5",  1.0e9, &mpZc);
+            addInputVariable("Z_c", "Impedance", "Pa s/m^3",  1.0e9, &mpZc);
```
robbr48: This is of course the same unit, but I agree that it makes more sense to use Pa in hydraulics.

PR #1714, opened 2018-05-02
Improve core log handling

peterNordin:
Make it possible to choose if the log should be active and where to put it in the cli.
Default log location for hopsan gui is now the temp directory.
The log is disabled if not explicitly opened by the user code.
Various related cleanup.


peterNordin:
The main pupose of this PR is to make it easier to debug the reason for the strange slowdowns befor or after simulation that affect some users. It should now be easier to rule out this log since it does not go into the working directory.

PR #1729, opened 2018-05-21
List unconnected ports in HCOM. Resolves #1727.

### LINKED ISSUE #1727: Make it easier to identify all unconnected ports
Perhaps HCOM command to list them

robbr48:


HopsanGUI/HcomHandler.cpp
```diff
@@ -4718,6 +4726,40 @@ void HcomHandler::executeConnectCommand(const QString cmd)
...
+void HcomHandler::executeListUnconnectedCommand(const QString cmd)
+{
+    QStringList args = splitCommandArguments(cmd);
+    QString wildcard = "*";
+    if(args.size() == 1)
+    {
+        wildcard = args[0];
+    }
+    else if(args.size() > 1)
+    {
```
peterNordin: It would be helpful if this text mentioned how many are supported, or maybe list the description.

HopsanGUI/HcomHandler.cpp
```diff
@@ -4718,6 +4726,40 @@ void HcomHandler::executeConnectCommand(const QString cmd)
...
+    }
+
+    QList<Port*> ports;
+    getPorts(wildcard,ports);
+
+    QStringList portNames;
+    for(Port* pPort : ports)
+    {
+        if(!pPort->isConnected())
+        {
```
peterNordin: Maybe we could print the name directly here instead of temp storing in portNames

PR #1730, opened 2018-05-21
Filtering DIPA output by value. Resolves #1726.

### LINKED ISSUE #1726: HCOM: Filter DIPA output by value
To for example print all parameters greater than a certain value

robbr48:


robbr48:
I added help text explaining the command with 0, 1 or 2 arguments. Also added hint for correct number of arguments in error message.

PR #1731, opened 2018-05-22
Issue 1542 fmu export report errors

### LINKED ISSUE #1441: FMU locallogger should log to hopsan core
---

Original Redmine Issue: 1441, https://flumes.iei.liu.se/redmine/issues/1441
Original Date: 2015/03/09

---

FMU locallogger should log to hopsan core
To begin with std::out is enough

But hopsan core should be able to give a function to external libraries that they can use to write to message handler

### LINKED ISSUE #1542: An exported FMU that internally runs Hopsan will not signal failure if initialize (and maybe simulate) fails
---

Original Redmine Issue: 1542, https://flumes.iei.liu.se/redmine/issues/1542
Original Date: 2016/02/05

---

I tried to export the NumHopInterpreter component and it could obviously not find the script file set as a a parameter.
The result was that the FMU is simulated successfully but only output 0 at every timestep

An other test is to build a model with a C or  Q component that is not connected. 
The FMU will export and after import it is possible to start simulation even though the model is incomplete. This simulation results in a crash.

peterNordin:
Resolves #1542 
Resolves #1441


peterNordin:
The first one (reloading) works for me, so this is odd. It was the import code that needlessly loaded the fmu dll and never released it, and I belive that I have fixed that. Or is this a different problem.

peterNordin:
I need to test one more thing before merging. I gott gui freeze if debug messages were enabled on linux. Not sure if it was related to this or if it was a new problem. May have been fixed by, the last commits to this PR

robbr48:
The reloading issue was not related to loading a DLL. I first got a warning that a folder with same name existed. When I clicked yes, I got an error saying something like "target file exists" in the compilation dialog. This was on Windows, so it could of course be related to some file indexing or similar.

peterNordin:
This is similar yo the problem I solved (hopefully). The dll would be locked by the hopsan process, so the import function would not be allowed to clear the import directory, before importing. (It deletes everything) but this dir should not be user selectable, so should be safe. 

If directory is not cleared then the target file may already exist and import is aborted. So thid could be the same problem. 
But as you say, windows tend to lock files for no apperant reason, so could be that as well.

PR #1733, opened 2018-05-24
Don't clear optimization script when closing dialog

### LINKED ISSUE #1701: Optimization script should not be cleared when optimization dialog is closed
In fact, I think it should be saved along with the model. It must however be clear for the user that the "Generate Script" button must be used to generate a new script.

robbr48:
Resolves #1701 

Drawback is that users may forget to click "Regenerate Script" if the previous script is already there. But this is still a better solution, since you no longer risk to lose manual changes to the script.


peterNordin:
But what happens if you forget to regenerate, will it run the old script or not have a script at all.
The mScript variable was cleared.

PR #1735, opened 2018-05-24
Update total progressbar during optimization

### LINKED ISSUE #1699: Optimization: Total progress bar is not updated for PSO algorithm
This may concern other algorithms as well.

robbr48:
Resolves #1699 

Progress bar was disabled when removing Qt stuff from Ops library. It could no longer use signals and slots. I fixed it by re-implementing the stepComplete() function in HopsanGUI.

### DIFF
--- a/HopsanGUI/OptimizationHandler.h
+++ b/HopsanGUI/OptimizationHandler.h
@@ -77,6 +77,7 @@ public:
     void objectiveChanged(size_t idx);
     void candidatesChanged();
     void candidateChanged(size_t idx);
+    void stepCompleted(size_t steps) override;
 
 public slots:
     void abort();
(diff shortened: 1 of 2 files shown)


HopsanGUI/OptimizationHandler.h
```diff
@@ -77,6 +77,7 @@ class OptimizationMessageHandler : public QObject, public Ops::MessageHandler
     void objectiveChanged(size_t idx);
     void candidatesChanged();
     void candidateChanged(size_t idx);
+    void stepCompleted(size_t);
```
peterNordin: The variable 'value' is missing here
Also if this method (and others) overload virtual methods in the base class, they should be marked override. 
That would lead to less risk for mistakes.

robbr48: I didn't even know there were such a keyword, but now it's there. You learn something new every day :)

peterNordin: Yes it came with c++11

PR #1736, opened 2018-05-25
Moved optimization script wizard to new dialog

### LINKED ISSUE #1734: Optimization: Put script generation in separate wizard
It is confusing when the script generator is opened by default in the optimization dialog. It should open the script page by default, and there should be three buttons: "Generate new script", "load script" and "save script". The first button shall open a new wizard.

robbr48:
Resolves #1734 

Script generation is now completely separated from the optimization dialog. This should give a much more intuitive user experience, since you don't need to step through the generator pages before starting the script.


HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -0,0 +1,1611 @@
...
+#include "Widgets/HcomWidget.h"
+#include "Utilities/GUIUtilities.h"
+#include "OptimizationScriptWizard.h"
+
+//! @brief Constructor
+OptimizationScriptWizard::OptimizationScriptWizard(SystemContainer* pSystem, QWidget *parent)
+    : QWizard(parent)
+{
+        //Set the name and size of the main window
+    this->resize(800,600);
```
peterNordin: Maybe scale to percentage of curent resolution if someone havs super high resolution

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -0,0 +1,1611 @@
...
+    this->setOption(QWizard::HaveHelpButton);
+    pHelpButton->setObjectName("optimizationHelpButton");
+
+
+
+    this->addPage(pSettingsWidget);
+    this->addPage(pParametersWidget);
+    this->addPage(pObjectiveWidget);
+    //this->addPage(pOutputWidget);
+    //this->addPage(pRunWidget);
```
peterNordin: leftover comments

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -0,0 +1,1611 @@
...
+    pSettingsLayout->addWidget(mpPlotParticlesCheckBox,row++, 0, 1, 2);
+    pSettingsLayout->addWidget(mpPlotEntropyCheckBox,  row++, 0, 1, 2);
+    pSettingsLayout->addWidget(mpExport2CSVBox,        row++, 0, 1, 2);
+    pSettingsLayout->addWidget(mpFinalEvalCheckBox,    row++, 0, 1, 2);
+    pSettingsLayout->addWidget(new QWidget(this),      row++, 0, 1, 2);    //Dummy widget for stretching the layout
+    pSettingsLayout->setRowStretch(row++, 1);
+    QWizardPage *pSettingsWidget = new QWizardPage(this);
+    pSettingsWidget->setLayout(pSettingsLayout);
+    pSettingsWidget->setPalette(gpConfig->getPalette());
+    setAlgorithm(0);
```
peterNordin: What algorithm is 0, 
ok saw that its undefined, maybe should have that in the Ops enum.
'NoAlgorithm' or similar

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -0,0 +1,1611 @@
...
+    {
+
+    //Parameter is not in list (should not really happen), so remove it here instead
+    mpParametersLayout->removeWidget(mpParameterLabels.at(i));
+    mpParametersLayout->removeWidget(mpParameterMinLineEdits.at(i));
+    mpParametersLayout->removeWidget(mpParameterMaxLineEdits.at(i));
+    mpParametersLayout->removeWidget(mpParameterRemoveButtons.at(i));
+
+    delete(mpParameterLabels.at(i));
+    delete(mpParameterMinLineEdits.at(i));
```
peterNordin: Here and in other places, maybe should use ->deleteLater(), in case events or signals are queued to these widgets

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -0,0 +1,1611 @@
...
+{
+    mObjectiveFunctionDescriptions.clear();
+    mObjectiveFunctionCalls.clear();
+    mObjectiveFunctionDataLists.clear();
+    mObjectiveFunctionNumberOfVariables.clear();
+    mObjectiveFunctionUsesTimeVector.clear();
+
+    // Look in both local and global scripts directory in case they are different
+
+    QDir scriptsDir(gpDesktopHandler->getScriptsPath()+"HCOM/objFuncTemplates");
```
peterNordin: We should not assume that paths are returned with trailing /, add /HCOM

HopsanGUI/HcomHandler.cpp
```diff
@@ -2297,9 +2297,9 @@ void HcomHandler::executeRunScriptCommand(const QString cmd)
         gpOptimizationDialog->open();
         gpOptimizationDialog->setCode(code);
         if(code.contains("opt autostart")) {
-            gpOptimizationDialog->next();
-            gpOptimizationDialog->next();
-            gpOptimizationDialog->next();
+//            gpOptimizationDialog->next();
```
peterNordin: Leftover comments

HopsanGUI/OptimizationHandler.h
```diff
@@ -95,6 +95,7 @@ class OptimizationHandler : public QObject
 
     friend class HcomHandler;           //! @todo Should probably not be friends
     friend class OptimizationDialog;    //! @todo Should probably not be friends
+    friend class OptimizationScriptWizard;    //DEBUG! REMOVE LATER /robbr
```
peterNordin: Leftover ?

HopsanGUI/Dialogs/OptimizationDialog.cpp
```diff
@@ -61,1742 +62,237 @@
...
     this->setPalette(gpConfig->getPalette());
 
-    //Settings tab
-    QLabel *pSettingsLabel = new QLabel("Please choose general settings for optimization algorithm.");
-    QFont boldFont = pSettingsLabel->font();
-    boldFont.setBold(true);
-    pSettingsLabel->setFont(boldFont);
+    // Main layout
+    QVBoxLayout *pLayout = new QVBoxLayout(this);
+    this->setLayout(pLayout);
```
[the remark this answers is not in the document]
peterNordin: Actually it does matter a little. In debug mode a warning is printed to stderr
QLayout: Attempting to add QLayout "" to OptimizationDialog "", which already has a layout

peterNordin:
The help button in the optimization dialog is not connected to anything

peterNordin:
The buttons in the optimization dialog is missing tool-tips (they have setText) but that is not the same

peterNordin:
On KUbuntu, the script wizard does not work for me, I do the typical Position servo example, choose complex rf, check all the plot boxes, choose P-gain set min max range. But then in the opjective function generator, I choose Step.out.value and PostionSensor.out.value -> minimize -> PROBLEM

The function selction combo box only show me two blank choices, that take up all of may vertical screen space, I can click one of them, to get either Overshoot or Lowest value, but clicking Add Function does nothing and I can get no further

HopsanGUI/Dialogs/OptimizationDialog.cpp
```diff
@@ -61,1742 +62,237 @@
...
+        int i = mParameterOutputIndexes[x];
+        if(i >= (int)objectives.size()) continue;
 
-        OptimizationSettings optSettings;
-        mpSystem->getOptimizationSettings(optSettings);
-        QString min, max;
-        for(int i=0; i<optSettings.mParamters.size(); ++i)
+        QString output = "obj: ";
+        QString objStr;
+        //! @todo This is an error in the code, this must be solved, previous code assumed that i is in range of objects -> ASSERT failed in QVector
```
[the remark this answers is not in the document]
robbr48: The check is actually redundant already, since it both has the error check and this line above:
if(i >= (int)objectives.size()) continue;
This error will only occur if someone calls the function with objectives and values vectors of different size, which is of course not allowed. And it will not cause a crash anyway, it would just ignore the excessive objectives.

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
robbr48: It does not mean undefined (there is an "++i" in setAlgorithm). 0 simple means use the first algorithm in the drop-down menu (which is currently Nelder-Mead).

peterNordin: ok, I did not mean undefined as undefined variable, I saw that i++, I just mean that the magic value 0 is unclear.
I would be nice if there would be a enum value or named value that represent, "nothing chosen" or similar. 
But not such a big deal

robbr48:
I cannot recreate the function selection error on Ubuntu 17.10. It searches for optimization scripts in two directories:
/home/username/Documents/Hopsan/Scripts/HCOM/objFuncTemplates
/home/username/Documents/Git/hopsan/bin//../Scripts/HCOM/objFuncTemplates
If both of these are empty or if the paths are interpreted wrong, the box will be empty. I'm not sure what will happen if there are erroneous HCOM files in the directory though.

However, I think this is a separate issue. I did not change this code in any way in this PR, so it most likely also existed before. I think we should create a new issue about it.

peterNordin:
I think the commit made tody is incorrect, it seemt to belong to some other branch and contain lots of stuff.

peterNordin:
I will test this on windows and see if that works better.
You are right that the problems I was having are old problems. (Will report issue)

peterNordin:
There is something odd with the last commit, it adds changes to the tools submodule, (fmu checkers).
But that submodule has no new commits in the main repo ?

peterNordin:
Seems like it has been reverted to an older commit, unintentional I assume

peterNordin:
Seems to work on Windows, much better usability now.

Found these issues, but likely separate issue
* When selecting overshoot having two output variables selected, addFunction silently does nothing. (should show some error, that I cant choose two variables)
* When enabling plots, they lack usefull information like parameter names and legend names. It mostly show the number 1 everywhere, but I think it did that before as well.

robbr48:
I reverted the accidental subproject commit, and also deleted a leftover debug output.

Yes, the plots should contain more information and there should be better error handling when adding functions. I will create separate issues for this (#1737 and #1738), since these problems already existed before.

PR #1740, opened 2018-05-29
Model setting for disabling animation mode

### LINKED ISSUE #1703: Disable animation model setting
Useful when teaching mostly. Students tend to use animation instead of plotting.

robbr48:
Restriction for educational use. Resolves #1703.

### DIFF
--- a/HopsanGUI/GUIObjects/GUIContainerObject.h
+++ b/HopsanGUI/GUIObjects/GUIContainerObject.h
@@ -187,6 +187,10 @@ public:
     bool areLossesVisible();
     QRadioButton *mpAvgPwrRadioButton;
 
+    //Animation methods
+    bool isAnimationDisabled();
+    void setAnimationDisabled(bool disabled);
+
 public slots:
 
     //Selection slots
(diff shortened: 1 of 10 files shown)


HopsanGUI/GUIObjects/GUIModelObject.cpp
```diff
@@ -636,6 +636,17 @@ bool ModelObject::isDisabled()
     }
 }
 
+bool ModelObject::isAnimationDisabled()
+{
+    // Does nothing by default, should be overloaded
```
peterNordin: I am not sure, but a return may be needed here to prevent warning ?

Maybe the function (and others) should be pure virtual. I dont think that we will ever have an instance of ModelObject directly

HopsanGUI/GUIObjects/GUIContainerObject.h
```diff
@@ -374,6 +378,9 @@ public slots:
     //Time measurement dialog
     QTableView *mpComponentTable;
     QTableView *mpTypeTable;
+
+    //Animation members
```
peterNordin: This one is not initialized, or?

HopsanGUI/GUIObjects/GUIModelObject.cpp
robbr48: I removed it from base class completely. It is only relevant for container objects anyway.

HopsanGUI/GUIObjects/GUIContainerObject.h
robbr48: Added initialization (false)

PR #1741, opened 2018-05-29
Check for Inf and NaN in animation

### LINKED ISSUE #1695: Animation crashes when variable is NaN or inf
Only solution is to check all variables for illegal data. It will reduce performance, but stability is more important.

Suggestion:
- Replay animation: Check all varaibles for NaN or Inf before playing
- Realtime animation: Check all variables for NaN or Inf at every step
- Indicate somehow to the user that something is wrong (highlighting component, stopping animation or similar)

robbr48:
Prevents crashes when animation data contains on-finite values. Resolves #1695.
- Disable playback buttons if a non-finite value is detected in animation data
- Abort real-time playback if a non-finite value is detected
- No noticable slowdown detected on my machine

I am actually not sure if Inf or NaN causes crashes, or it if just takes a very long time to compute the position. I have seen cases where the GUI freezes for some minutes and then starts working again. In either case, this check is useful.

### DIFF
--- a/HopsanGUI/GUIObjects/AnimatedComponent.cpp
+++ b/HopsanGUI/GUIObjects/AnimatedComponent.cpp
@@ -49,6 +49,7 @@
 #include "Utilities/GUIUtilities.h"
 #include "Widgets/AnimationWidget.h"
 #include "Widgets/ModelWidget.h"
+#include "MessageHandler.h"
 
 #include "PlotHandler.h"
 #include "PlotWindow.h"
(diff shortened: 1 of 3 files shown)


HopsanGUI/GUIObjects/AnimatedComponent.cpp
```diff
@@ -106,6 +107,17 @@ AnimatedComponent::AnimatedComponent(ModelObject* unanimatedComponent, Animation
                             tempPortName = mpModelObject->getPort(portName)->getConnectedPorts().first()->getName();
                         }
                         mpData->last().append(mpAnimationWidget->getPlotDataPtr()->copyVariableDataVector(makeFullVariableName(mpModelObject->getParentSystemNameHieararchy(), tempComponentName, tempPortName, dataName),generation));
+
+                        const auto& dataValues = mpData->last().last();
```
peterNordin: In this case there should be an & after auto, to ensure that dataValue is not a copy, we want to use a const reference to the actual value.
This would be equivalent to the [i] call that was used before

peterNordin: Hmm my app seem to add the comments to the wrong line.
I mean the one in the for loop

PR #1742, opened 2018-06-01
Improved error handling in optimization script wizard

### LINKED ISSUE #1737: Optimization: Better error handling in script wizard
- [x] Error message when trying to add function with wrong number of variables
- [x] Do not allow generating script with no variables selected
- [x] Do not allow min parameter limit to be greater than max

robbr48:
Resolves #1737.

### DIFF
--- a/HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
+++ b/HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
@@ -58,6 +58,29 @@
 #include "Utilities/GUIUtilities.h"
 #include "OptimizationScriptWizard.h"
 
+
+
+OptimizationScriptWizardPage::OptimizationScriptWizardPage(OptimizationScriptWizard *pParent)
+    : QWizardPage(pParent)
+{
+    mpWizard = pParent;
+
+    connect(mpWizard, SIGNAL(contentsChanged()), this, SIGNAL(completeChanged()));
+}
+
+bool OptimizationScriptWizardPage::isComplete() const
+{
+    if(isFinalPage())
+    {
+        return (!mpWizard->mSelectedParameters.empty() && !mpWizard->mSelectedFunctions.empty());
+    }
+    else
+    {
+        return QWizardPage::isComplete();
+    }
+}
+
+
 //! @brief Constructor
 OptimizationScriptWizard::OptimizationScriptWizard(SystemContainer* pSystem, QWidget *parent)
     : QWizard(parent)
@@ -242,7 +265,7 @@ OptimizationScriptWizard::OptimizationScriptWizard(SystemContainer* pSystem, QWi
     pSettingsLayout->addWidget(mpFinalEvalCheckBox,    row++, 0, 1, 2);
     pSettingsLayout->addWidget(new QWidget(this),      row++, 0, 1, 2);    //Dummy widget for stretching the layout
     pSettingsLayout->setRowStretch(row++, 1);
-    QWizardPage *pSettingsWidget = new QWizardPage(this);
+    auto *pSettingsWidget = new OptimizationScriptWizardPage(this);
     pSettingsWidget->setLayout(pSettingsLayout);
     pSettingsWidget->setPalette(gpConfig->getPalette());
     setAlgorithm(0);
@@ -269,7 +292,7 @@ OptimizationScriptWizard::OptimizationScriptWizard(SystemContainer* pSystem, QWi
     mpParametersLayout->addWidget(pParameterMinLabel,      3, 0, 1, 1);
     mpParametersLayout->addWidget(pParameterNameLabel,     3, 1, 1, 1);
     mpParametersLayout->addWidget(pParameterMaxLabel,      3, 2, 1, 1);
-    QWizardPage *pParametersWidget = new QWizardPage(this);
+    OptimizationScriptWizardPage *pParametersWidget = new OptimizationScriptWizardPage(this);
     pParametersWidget->setLayout(mpParametersLayout);
 
     //Objective function tab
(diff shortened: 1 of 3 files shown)


HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -970,38 +995,24 @@ void OptimizationScriptWizard::saveConfiguration()
...
-                itsOk = false;
-            }
+            QMessageBox::warning(this, "Error", "All maximum parameter values must be greater than their minimum parameter values.");
+            return false;
         }
-        if(!itsOk)
+        if(minVal <= 0 && mpParametersLogCheckBox->isChecked())
         {
-            QMessageBox::warning(this, "Warning", "Logarithmic scaling is selected, but parameters are allowed to be negative or zero. This will probably not work.");
+            QMessageBox::warning(this, "Error", "Logarithmic scaling does not work with negative parameters.");
```
peterNordin: negative or zero ?

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -1356,20 +1364,20 @@ QString OptimizationScriptWizard::generateFunctionCode(int i)
...
     {
-        //This error makes no sense, since it is showed on the last page only, so user will see it too late
-        //mpMessageHandler->addErrorMessage("Too few variables selected for this function.");
+        if(!noWarning)
+        {
+            QMessageBox::warning(this, QString("Wrong number of arguments"),
+                                 QString::fromStdString("Selected function requires exactly "+
+                                 std::to_string(nVar)+
+                                 " variables, but you selected "+
+                                 std::to_string(nSelVar)+"."),QMessageBox::Ok);
```
peterNordin: This would be easier to read if something like: QString("text %1 more text %2").arg(v1).arg(v2) was used instead of mixing QString and std:string

Also I think that there should be a "The" before "selected"

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -1496,15 +1504,15 @@ void OptimizationScriptWizard::loadConfiguration()
         if(idx > -1) //found!
         {//Lägg till variabel i XML -> compname, portname, varname, ska vara i mSelectedVariables
             mpFunctionsComboBox->setCurrentIndex(idx);
-            addObjectiveFunction(idx, optSettings.mObjectives.at(i).mWeight, optSettings.mObjectives.at(i).mNorm, optSettings.mObjectives.at(i).mExp, optSettings.mObjectives.at(i).mVariableInfo, optSettings.mObjectives.at(i).mData);
+            addObjectiveFunction(idx, optSettings.mObjectives.at(i).mWeight, optSettings.mObjectives.at(i).mNorm, optSettings.mObjectives.at(i).mExp, optSettings.mObjectives.at(i).mVariableInfo, optSettings.mObjectives.at(i).mData, true);
```
peterNordin: I realize that this code was already here, but I would prefer to get the mSettings.mObjectives.at(i) once, and create a reference to it. Then use rhe reference instead of the above call multiple times

HopsanGUI/Dialogs/OptimizationScriptWizard.h
```diff
@@ -90,12 +102,12 @@ private slots:
...
     void generateCommonOptions(QString &templateCode);
     QString generateFunctionCode(int i);
-    bool verifyNumberOfVariables(int i, int nSelVar);
+    bool verifyNumberOfVariables(int i, int nSelVar, bool noWarning);
     bool loadObjectiveFunctions();
 
     void loadConfiguration();
 
-    void addObjectiveFunction(int idx, double weight, double norm, double exp, QList<QStringList> selectedVariables, QStringList objData);
+    void addObjectiveFunction(int idx, double weight, double norm, double exp, QList<QStringList> selectedVariables, QStringList objData, bool noWarning=false);
```
peterNordin: Double negation is confusing
I have done this sometimes as well, but we should stop,
printWarning=true would feel better.

peterNordin: Also in other places

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -846,6 +869,7 @@ void OptimizationScriptWizard::addFunction()
 {
     int idx = mpFunctionsComboBox->currentIndex();
     addObjectiveFunction(idx, 1.0, 1.0, 2.0, mSelectedVariables, QStringList());
+    emit(currentPage()->completeChanged());
```
peterNordin: This fail to build with Qt4 since the signal is protected. Maybe it can be emitted directly from the new page subclass somehow.
It might also be possible to chain link a signal from the wizard to the signal inside the page. Signal to signal connection.
In that case the connection must be made inside the page class and you need a new signal in the wizard class

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -242,7 +263,7 @@ OptimizationScriptWizard::OptimizationScriptWizard(SystemContainer* pSystem, QWi
     pSettingsLayout->addWidget(mpFinalEvalCheckBox,    row++, 0, 1, 2);
     pSettingsLayout->addWidget(new QWidget(this),      row++, 0, 1, 2);    //Dummy widget for stretching the layout
     pSettingsLayout->setRowStretch(row++, 1);
-    QWizardPage *pSettingsWidget = new QWizardPage(this);
+    OptimizationScriptWizardPage *pSettingsWidget = new OptimizationScriptWizardPage(this);
```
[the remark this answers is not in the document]
robbr48: In this case it doesn't really matter to me. I like the code to be self-documenting, which is sometimes made harder by the "auto" keyword. But in this case it is obvious what it means.

peterNordin:
It would be nice if the merge commit could be removed, a merge with your self. It looks strang. Not sure if its easy because the changes are obviously needed. If its to complicated, let it be, but if possible remove it.

robbr48:
Not sure where the merge commit comes from. There are also two regular commits with same name, but on different branches. It must be related to me doing something wrong while working on both Linux and Windows. I believe the only solution is to redo all the changes manually and make a new commit. I tried to make a patch from the diff against master, but it didn't apply.

robbr48:
Not sure where the merge commit comes from. There are also two regular commits with same name, but on different branches. It must be related to me doing something wrong while working on both Linux and Windows. I believe the only solution is to redo all the changes manually and make a new commit. I tried to make a patch from the diff against master, but it didn't apply.

robbr48:
I solved it in a slightly different way, but now there is only one commit :)

PR #1743, opened 2018-06-03
Issue 1723 auto close dialog on success

### LINKED ISSUE #1138: Compiler utility always thinks compilation was successful if previous target file exist
---

Original Redmine Issue: 1138, https://flumes.iei.liu.se/redmine/issues/1138
Original Date: 2014/03/31
Original Assignee: Robert Braun

---

None

### LINKED ISSUE #1723: Recompilation dialog is annoying
Especially if a large number of libraries are recompiled successfully at startup. User then need to close each dialog individually. Also enter button does not seem to work, so user needs to click on "close" in every dialog.

Suggestion: Use one single new dialog for displaying all recompilation. Output can be opened by clicking a button. Dialog should close automatically after X seconds if all recompilations were successful.

peterNordin:
* Add boolean return values from generator functions to notify success or failure
* Make generator dialog responsive (multi thread)
* Auto close import dialogs on success 
Resolves #1723 
Resolves #1138

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ output/
 moc_*.cpp
 Makefile
 compileWithMinGW.bat
+GPATH
+GTAGS
+GRTAGS
+GSYMS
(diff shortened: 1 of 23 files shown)


peterNordin:
I forgot to change to the new utility functions in all places. The one you found was the most important one as its the most annoying one. I added a new commit

PR #1745, opened 2018-06-04
Issue 1738 improve optimization plots

### LINKED ISSUE #1738: Optimization: Improve variable names in plots
All variables are now labelled "(1)" for some reason.
- [x] Points should be called x1, x2, x3...
- [x] Candidates should be called c1, c2, c3...
- [x] Entropy should be called "Entropy"
- [x] Objectives should be called "Best", "Worst" and "Latest"
- [x] Parameters should be called par1, par2, par3...

Unfortunately there is no mapping between parameter names and optimization parameters, so it is not possible to print the model parameter names currently.

### LINKED ISSUE #1744: Optimization plot windows should be reusable
If you close a plot window during optimization and restart the script, the plot window does not open again as it should.

robbr48:
Show correct plot variable names and axis labels. Resolves #1738. 

Plot windows are now re-opened when starting a new optimization, even if they were closed in the previous one. Resolves #1744.

### DIFF
--- a/HopsanGUI/OptimizationHandler.cpp
+++ b/HopsanGUI/OptimizationHandler.cpp
@@ -89,6 +89,8 @@ void OptimizationHandler::startOptimization(ModelWidget *pModel, QString &modelP
 {
     if(mpWorker)
     {
+        clearPlotVariables();
+
         mModelPath = modelPath;
 
         mEvaluations = 0;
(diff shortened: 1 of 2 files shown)


robbr48:
I replaced the "auto&" with just "auto" for the variable description objects.

PR #1746, opened 2018-06-04
Add plot export button that closes the dialog

### LINKED ISSUE #1676: Export plot to image dialog should probably close after successful export
---

Original Redmine Issue: 1676, https://flumes.iei.liu.se/redmine/issues/1676
Original Date: 2018/01/22

---

Users rarely want to export the same plot multiple times. It feels unintuitive that the dialog stays open after the export is finished.

peterNordin:
But keep a button for apply style export
Add error message if exported file is missing
Resolves #1676


peterNordin:
Hmm that sounds really annoying, I will check if some flag might be helpful. The problem is likely that the dialog signals the plottab to open the export file select dialog. This dialog will be a sibling to the export dialog itself. Needs redesign but that will have to wait until some other time.

peterNordin:
Could not reproduce in Kubuntu but set export dialog as parent to file dialog (instead of null). Will likely help.

PR #1748, opened 2018-06-08
File handling for optimization scripts. Resolves #1739.

### LINKED ISSUE #1739: Optimization: Correct script file handling
- [x] Store script file info when loading
- [x] Track modifications in script text edit
- [x] Ask for saving script file if modified when starting an optimization or closing the dialog

robbr48:


HopsanGUI/Dialogs/OptimizationDialog.h
```diff
@@ -76,16 +77,18 @@ public slots:
 private slots:
     void generateScriptSkeleton();
     void openScriptWizard();
-    void saveScriptFile();
```
peterNordin: I dont think that the saveScript function taking bool is needed. The one tanking qstring and the one savingAs should be enough.

If the one taking string gets an empty string and mScriptFileInfo is not set (or points to file that does not exist) then the saveAs should be called. 
All file selction code could be contained in that function, then it would call the saveScript function taking qstring with the selected path. 
Later calling saveScript would then save to the existing mScriptFile.

HopsanGUI/Dialogs/OptimizationDialog.cpp
```diff
@@ -568,16 +620,53 @@ void OptimizationDialog::applyParameters()
 }
```
peterNordin: Why is this comparison needed, is it not enough that the signal is emitted, (textChanged)?
Or is that signal emitted to often?

HopsanGUI/Dialogs/OptimizationDialog.cpp
```diff
@@ -568,16 +620,53 @@ void OptimizationDialog::applyParameters()
...
```
peterNordin: The previous comment should have gone here

HopsanGUI/Dialogs/OptimizationDialog.cpp
```diff
@@ -568,16 +620,53 @@ void OptimizationDialog::applyParameters()
...
 
 
-//! @brief Asks user whether or not to overwrite optimization script
-bool OptimizationDialog::checkIfScriptIsEmpty()
+//! @brief Slot for flagging that script text has changed
+void OptimizationDialog::scriptTextChanged()
 {
-    if(!mpScriptBox->document()->isEmpty())
+    mScriptTextChanged = (mpScriptBox->toPlainText() != mSavedScriptText);
```
peterNordin: Why is this comparison needed, is it not enough that the signal is emitted, (textChanged)?
Or is that signal emitted to often?

robbr48: It is needed because the slot is called also when user presses ctrl-z. If all changes have been undone, the asterisk shall be removed from the tab text.

HopsanGUI/Dialogs/OptimizationDialog.h
robbr48: But there are two QToolButtons, which each need to be connected to a different slot. We could use a QSignalMapper to provide arguments to the slots, but that would be over-complicated.

It is the same problem in ModelWidget class, which has three functions: save(), saveAs() and saveModel().

I will move all save code to one of the functions though, to make things clearer.

peterNordin: Yes two buttons save and save as, then only two slots should be needed. Or maybe I am missing some implementation detail. I have only looked at the PR diff, not the surrounding code.

HopsanGUI/Dialogs/OptimizationDialog.cpp
peterNordin: Ok

HopsanGUI/Dialogs/OptimizationDialog.h
robbr48: We also use it for saving a backup copy of the script prior to starting the optimization. Therefore we need a save function that can take an arbitrary filepath as a string argument.

peterNordin: Yes the one with QString is needed, the one taking bool should not be needed. I think that you can signal the slot taking qstring without argument. Then if the qstring defaults to empty string, that could be used to decide if the saveAs slot should be called. Depends on if the current mSaveFileInfo (or whatever its name is) is set (and what it is set to)

peterNordin: The save(qstring) can call saveAs() (if qstring empty and current file empty).
SaveAs opens the dialog and then calls save(qstring=validpath) which saves to the intended file.

Anyway not super important...

robbr48: Ok, I think I managed to solve it with only two functions. Check for current file being empty needs to be done in the save() function before calling saveAs(), but that should not be a problem I think.

peterNordin: Have not tested, but looks good now

Issue #1711, opened 2018-04-18
Installation on OS X 


mjfwest:
Hello,
Has anyone recently built the application for OS X?
The scripts in buildMacApp seem out of date.   I have followed most of the instructions for debian/unix build but I have the following error when building in QTcreator. 

```
:-1: error: No rule to make target `/Users/martinwest/Documents/repositories/hopsan/HopsanCore/dependencies/libNumHop/src/Expression.cc', needed by `Expression.o'.  Stop.
```

peterNordin:
No, sorry, it is outdated, i recently changed a lot of things. But I do not have a Mac so I could not test this. But in this particular case, have you run git submodule update --init ?

Can you check if there is actually code in libNumHop, I have had problems where the submodule failed to download.

mjfwest:
Ok. I'm happy to put some effort in to updating the build instructions and testing if I can get it to compile successfully. 

Yes, the submodules were the issue there.  running ` git submodule update --init` updated `libNumHop`, but I believe some build script had also deleted the code from `IndexingCSVParser`.  Manually resetting and checking out the code fixed that.  I now have the following error:
```
compiling ../../hopsan/HopsanGenerator/src/generators/HopsanFMIGenerator.cc
In file included from ../../hopsan/HopsanGenerator/src/generators/HopsanFMIGenerator.cc:25:
In file included from ../../hopsan/HopsanGenerator/include/generators/HopsanFMIGenerator.h:29:
In file included from ../../hopsan/HopsanGenerator/include/generators/HopsanGenerator.h:56:
In file included from ../../hopsan/HopsanGenerator/include/GeneratorUtilities.h:49:
In file included from ../../hopsan/Dependencies/FMILibrary/include/JM/jm_portability.h:19:
../../hopsan/Dependencies/FMILibrary/include/JM/jm_callbacks.h:209:42: warning: unused parameter 'cb' [-Wunused-parameter]
static void jm_log_debug_v(jm_callbacks* cb, const char* module, const char* fmt, va_list ap) {}
                                         ^
../../hopsan/Dependencies/FMILibrary/include/JM/jm_callbacks.h:209:58: warning: unused parameter 'module' [-Wunused-parameter]
static void jm_log_debug_v(jm_callbacks* cb, const char* module, const char* fmt, va_list ap) {}
                                                         ^
[...]

PR #1751, opened 2018-06-13
MacOS no app bundle

peterNordin:
* Do not build Hopsan GUI apps as bundles
* Copy local dependencies into bin directory 

Related to #1711


peterNordin:
Yes I know, but since I do not have a mac computer, getting it to work is not so easy. I will try to do this properly in the future.

PR #1752, opened 2018-06-13
Separate local and system parameter names.

### LINKED ISSUE #1694: Separate local parameter names from system parameter names
When a system parameter and a local parameter in a component have the same name, it is unclear which one will be used. 

Suggestions:
- Use system parameters by default
- Use local parameter if system parameter with same name do not exist
- Add special syntax for force using local parameters

robbr48:
Resolves #1694.

### DIFF
--- a/HopsanCore/include/HopsanTypes.h
+++ b/HopsanCore/include/HopsanTypes.h
@@ -69,6 +69,7 @@ public:
     bool empty() const;
     bool compare(const char* other) const;
     bool compare(const HString &rOther) const;
+    bool startsWith(const HString& rOther) const;
     bool isNummeric() const;
     bool isBool() const;
     double toDouble(bool *isOK) const;
--- a/HopsanCore/src/HopsanTypes.cpp
+++ b/HopsanCore/src/HopsanTypes.cpp
@@ -235,6 +235,19 @@ bool HString::compare(const HString &rOther) const
     }
 }
 
+bool HString::startsWith(const HString &rOther) const
+{
+    if (mSize < rOther.size()) {
+        return false;
+    }
+    else if (empty() && rOther.empty()) {
+        return true;
+    }
+    else {
+        return (strncmp(mpDataBuffer, rOther.c_str(), rOther.size()) == 0);
+    }
+}
+
 //! @brief Check if the string can be interpreted as a number
 bool HString::isNummeric() const
 {
--- a/HopsanCore/src/Parameters.cpp
+++ b/HopsanCore/src/Parameters.cpp
@@ -706,21 +706,15 @@ bool ParameterEvaluatorHandler::setParameterValue(const HString &rName, const HS
 //! @param [out] rEvaluatedParameterValue The result of the evaluation
 //! @param [in] rType The type of how the parameter should be interpreted
 //! @return true if success, otherwise false
-bool ParameterEvaluatorHandler::evaluateParameter(const HString &rName, HString &rEvaluatedParameterValue, const HString &rType)
+bool ParameterEvaluatorHandler::evaluateInLocalComponent(const HString &rName, HString &rEvaluatedParameterValue, const HString &rType)
 {
     bool success = false;
-    // Try our own parameters
     for(size_t i = 0; i < mParameters.size(); ++i)
     {
         if ( (mParameters[i]->getName() == rName) && (mParameters[i]->getType() == rType) )
         {
             success = mParameters[i]->evaluate(rEvaluatedParameterValue);
         }
-    }
-    if(!success)
-    {
-        // Try one of our system parent component parameters
-        success = evaluateInSystemParent(rName, rEvaluatedParameterValue, rType);
     }
     return success;
 }
@@ -753,7 +747,10 @@ bool ParameterEvaluatorHandler::refreshParameterValueText(const HString &rParame
 bool ParameterEvaluatorHandler::evaluateInSystemParent(const HString &rName, HString &rEvaluatedParameterValue, const HString &rType)
 {
     // Try one of our system parent component parameters
-    if(mParentComponent && mParentComponent->getSystemParent())
+    if(mParentComponent && mParentComponent->isComponentSystem()) {
+        return mParentComponent->evaluateParameter(rName, rEvaluatedParameterValue, rType);
+    }
+    else if(mParentComponent && mParentComponent->getSystemParent())
     {
         return mParentComponent->getSystemParent()->evaluateParameter(rName, rEvaluatedParameterValue , rType);
     }
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -7784,6 +7784,11 @@ QString HcomHandler::getParameterValue(QString parameterName, QString &rParamete
             CoreParameterData par;
             pMO->getParameter(parName, par);
             rParameterType = par.mType;
+            if(par.mValue.startsWith("self."))
+            {
+                QString otherPar = par.mValue.remove(0,5);
+                return getParameterValue(compName+"."+otherPar);
+            }
             return par.mValue;
         }
 
@@ -8143,7 +8148,7 @@ bool HcomHandler::evaluateArithmeticExpression(QString cmd)
         evaluateExpression(right);
 
         QStringList pars;
-        if (mAnsType==Scalar)
+        if (mAnsType==Scalar || mAnsType==Wildcard)
         {
             getParameters(left, pars);
             SharedVectorVariableT data = getLogVariable(left);
@@ -8173,7 +8178,12 @@ bool HcomHandler::evaluateArithmeticExpression(QString cmd)
             return false;
         }
 
-        if(mAnsType==Scalar)
+        if(mAnsType == Wildcard && !pars.isEmpty())
+        {
+            executeCommand("chpa "+left+" "+mAnsWildcard);
+            return true;
+        }
+        else if(mAnsType==Scalar)
         {
             if(!pars.isEmpty())
             {
--- a/HopsanGUI/loadFunctions.cpp
+++ b/HopsanGUI/loadFunctions.cpp
@@ -230,6 +230,26 @@ void loadParameterValue(QDomElement &rDomElement, ModelObject* pObject, UndoStat
         return;
     }
 
+    if(existingNames.contains(parameterValue) &&
+       !pObject->getParentContainerObject()->hasParameter(parameterValue))
+    {
+        gpMessageHandler->addWarningMessage(QString("%1: Updating parameter value from \"%2\" to \"self.%2\"")
+                                            .arg(pObject->getName())
+                                            .arg(parameterValue));
+        pObject->setParameterValue(parameterName, "self."+parameterValue, true);
+        return;
+    }
+
+    if(existingNames.contains(parameterValue) &&
+       !pObject->getParentContainerObject()->hasParameter(parameterValue))
+    {
+        gpMessageHandler->addWarningMessage(QString("%1: Updating parameter value from \"%2\" to \"self.%2\"")
+                                            .arg(pObject->getName())
+                                            .arg(parameterValue));
+        pObject->setParameterValue(parameterName, "self."+parameterValue, true);
+        return;
+    }
+
     if(!(pObject->getTypeName() == MODELICATYPENAME &&
        (parameterName == "ports" || parameterName == "parameters" || parameterName == "defaults")))
     {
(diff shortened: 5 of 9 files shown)


HopsanCore/src/Parameters.cpp
```diff
@@ -710,17 +710,22 @@ bool ParameterEvaluatorHandler::evaluateParameter(const HString &rName, HString
 {
     bool success = false;
     // Try our own parameters
-    for(size_t i = 0; i < mParameters.size(); ++i)
+    if(rName.startsWith("self#"))
```
peterNordin: I am not sure if I like this, now "self./#" is required to self reference, this would feel odd when you write code and do something like pComponent->evaluateParameter("self#kp"), Now there are two "this"  pComponent and self#.

Also I am not sure if "self" belongs in here, the parameter evaluator should not need to know about "this", but I am not sure since it already know about "parent system". 

Also we should maybe try to use . instead of # for new things even in the core. I do not remember why I used # before.

HopsanCore/src/Parameters.cpp
```diff
@@ -710,17 +710,22 @@ bool ParameterEvaluatorHandler::evaluateParameter(const HString &rName, HString
...
+                success = mParameters[i]->evaluate(rEvaluatedParameterValue);
+            }
         }
     }
-    if(!success)
+    else
     {
         // Try one of our system parent component parameters
-        success = evaluateInSystemParent(rName, rEvaluatedParameterValue, rType);
+        HString name = "self#"+rName;   //Need to add self# for lookup in parent system
```
peterNordin: I think that this break the previous behavior. You could then recurse upwards in the model hierarchy (I think) until the desired value was found.

Now only the immediate parent will be visited. 

Also I think that it is kind of ugly to to this by force perpending "self", I would prefer class member function evaluateOwnParameter (or something similar). Need to think about this....

HopsanGUI/loadFunctions.cpp
```diff
@@ -230,6 +230,16 @@ void loadParameterValue(QDomElement &rDomElement, ModelObject* pObject, UndoStat
...
+    if(existingNames.contains(parameterValue) &&
+       !pObject->getParentContainerObject()->hasParameter(parameterValue))
+    {
+        gpMessageHandler->addWarningMessage(QString("%1: Updating parameter value from \"%2\" to \"self.%2\"")
+                                            .arg(pObject->getName())
+                                            .arg(parameterValue));
+        pObject->setParameterValue(parameterName, "self."+parameterValue, true);
+        return;
+    }
+
```
peterNordin: This code will not update NumHop expressions containing parameter names. Only if the value is exactly a parameter name.

Also this code makes the model loose backwards compatibility. You wont be able to load it in an older version of Hopsan (unless you manually remove the self.). But I know this is quit difficult to solve with a reasonable amount of work.

The same update must be made in the Core HMF loader otherwise the models will fail when run through the CLI and remote. (TEsts are currently failing) likely this is the cause. (Not sure)

SymHop/src/SymHop.cpp
```diff
@@ -365,7 +365,7 @@ void Expression::commonConstructorCode(QStringList symbols, bool &ok, const Expr
         int start=0;        //Start index of each symbol
         for(int i=0; i<str.size(); ++i)
         {
-            if(!var && parBal==0 && (str.at(i).isLetterOrNumber() || str.at(i) == '_' || str.at(i) == '-' || str.at(i) == '.' || str.at(i) == ':')) //New variable or function string
+            if(!var && parBal==0 && (str.at(i).isLetterOrNumber() || str.at(i) == '_' || str.at(i) == '-' || str.at(i) == '.' || str.at(i) == ':' || str.at(i) == '"')) //New variable or function string
             {
```
peterNordin: It is unclear to me why these changes are needed

HopsanGUI/HcomHandler.cpp
```diff
@@ -7988,7 +7993,7 @@ bool HcomHandler::evaluateArithmeticExpression(QString cmd)
         evaluateExpression(right);
 
         QStringList pars;
-        if (mAnsType==Scalar)
+        if (mAnsType==Scalar || mAnsType==Wildcard)
```
peterNordin: It is unclear to me why this change is needed, what relation does it have with self.
If this is a "Drive by fix" thats fine, I am just curious,

HopsanGUI/HcomHandler.cpp
```diff
@@ -8018,7 +8023,12 @@ bool HcomHandler::evaluateArithmeticExpression(QString cmd)
             return false;
         }
 
-        if(mAnsType==Scalar)
+        if(mAnsType == Wildcard && !pars.isEmpty())
+        {
```
peterNordin: Also here, I assume that this is some sort of assignment to parameters like Gain.p = Gain2.p

HopsanCore/src/HopsanTypes.cpp
```diff
@@ -235,6 +235,22 @@ bool HString::compare(const HString &rOther) const
...
 }
 
+bool HString::startsWith(const HString &rOther) const
+{
+    if (mSize < rOther.size())
+    {
+        return false;
+    }
+    else if ( (mSize == rOther.size()) && (mSize == 0) )
+    {
```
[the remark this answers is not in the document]
robbr48: I agree. This line was copied from the existing HString::compare() function, so I guess we should change both then.

HopsanCore/include/HopsanTypes.h
```diff
@@ -69,6 +69,7 @@ class HOPSANCORE_DLLAPI HString
     bool empty() const;
     bool compare(const char* other) const;
     bool compare(const HString &rOther) const;
+    bool startsWith(const HString& rOther) const;
```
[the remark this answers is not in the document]
robbr48: Good point, I will add one.

HopsanCore/src/Parameters.cpp
robbr48: I thought this was what we agreed on; if you want to reference a parameter in the same component you should prepend "self." to avoid confusion. 

The . was replaced by # automatically somewhere. Perhaps it always replaces all dots with hashes before evaluating the parameters?

HopsanCore/src/Parameters.cpp
robbr48: That's true, perhaps the "self." should be parsed at a higher level in the evaluateParameter hierarchy (which is a very long chain of functions right now). Then we can just add an ignoreSystemParameters argument which is false by default and set to true when "self." is encountered.

HopsanGUI/loadFunctions.cpp
robbr48: I think updating NumHop scripts would be really complicated. Or is there some function for easily replacing symbol names?

Loading new models with old versions of Hopsan is never guaranteed to work, and I see no solution for this really. It would require a "Hopsan version" setting in the save dialog, to save in older HMF formats. Old Hopsan versions of course have no idea about the meaning of "self."...

SymHop/src/SymHop.cpp
robbr48: It is needed so that string parameters can be set through HCOM. Consider the following commands:
> chpa Mass.B.y self.m
> chpa Mass.B.y "self.m"
The first will set Mass.B.y to the evaluated value of "self.m". The second will set Mass.B.y to reference "self.m". But this requires SymHop to support quoted strings.

HopsanGUI/HcomHandler.cpp
robbr48: Save as above. We need to also evaluate wildcards, in case you want to set a parameter to a string value. Perhaps we should make a separate PR for this, since it can be used for other purposes as well.

HopsanCore/src/HopsanTypes.cpp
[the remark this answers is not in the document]
peterNordin: Yes please change the other code as well

HopsanGUI/loadFunctions.cpp
peterNordin: If the parameter value is a script, just go through the text and replace parName with self.parName. Look for an occurance of any local parameter name.

HopsanCore/src/Parameters.cpp
peterNordin: I was thinking that self. should only be necessary if a name collission (ambiguity) is detected. Similarly to how the this pointer may be needed in a c++ class function. Say you have a member variable 'monkey' and a function parameter 'monkey'. Then you must place this-> infront of the member. But if the function parameter (argument) is called 'dog' then a this is not needed.
But maybe it is more cleare if the user must allways write self. explicitly. There is less risk for missinterpretation.

peterNordin: The hash thing is fot backward compatibility. Maybe we can try tp solve this later. In a release that breaks this compatibility.

HopsanCore/src/Parameters.cpp
peterNordin: Yes the function calling here is a nightmare.

HopsanGUI/HcomHandler.cpp
peterNordin: Ok, keep it

HopsanGUI/loadFunctions.cpp
robbr48: @peterNordin Will that really work? Consider the case that the expression is "x+xp", where x is a local parameter and xp is a system parameter. The correct replacement will be "self.x+xp". But since it will replace all "x" with "self.x", it will end up with "self.x+self.xp". It would require a NumHop function that will rename all occurrences of a variable.

peterNordin: Yes, sure we would need to parse the expression and extract variable names and check which are not system parameters.  Maybe its not worth the effort really.

robbr48:
@peterNordin This should now work with hierarchical system parameters. I also added unit test for `HString::startsWith` function. Can it be merged, or does it require more work?

peterNordin:
I will test this again

HopsanCore/src/Parameters.cpp
```diff
@@ -753,7 +747,10 @@ bool ParameterEvaluatorHandler::refreshParameterValueText(const HString &rParame
 bool ParameterEvaluatorHandler::evaluateInSystemParent(const HString &rName, HString &rEvaluatedParameterValue, const HString &rType)
 {
     // Try one of our system parent component parameters
-    if(mParentComponent && mParentComponent->getSystemParent())
+    if(mParentComponent && mParentComponent->isComponentSystem()) {
```
peterNordin: I am not sure this code does what you would think when reading the variable names, or possibly it did the wrong thing before, or possible your changes require this new behavior. 

Does the following make sense:  The Parameter evaluator is attached to a component, this is the badly named mParentComponent (I am fairly sure about this, but not completely).

evaluateInSystemParent, should evaluate only in the system that the current component is inside, in its system-parent. This is what the original code did. But now it first checks if the attached component (badly named mParentComponent) is itself a SystemComponent, and if that is the case, it evaluates the parameter locally. And this would be the same as self.paramname, so this may not work in case the actual component is a system. This could happen with programmed subsystems, or maybe from NumHop script evaluation.

But we should really write some kind of tests for this to make sure. I will try to make one.
But right now I think that the above code is not doing what it is supposed to be doing.

HopsanGUI/loadFunctions.cpp
```diff
@@ -230,6 +230,26 @@ void loadParameterValue(QDomElement &rDomElement, ModelObject* pObject, UndoStat
         return;
     }
 
+    if(existingNames.contains(parameterValue) &&
```
peterNordin: The previous behavior of setParameter first tried to evaluated the parameter as if was self.parameter (inside the local component) then it would try with a system variable.

This replacement fix will correctly replace paramname with self.paramname in that case, but in the case where the paramname exist both in self and in parent it will not replace the name and we will now choose the parent version instead, which is not the same as before.  

Does this make sense ?
If so, maybe just drop the check that it is not in the parent container.
The same problem (if it is one) exists in the other place in this file with the same code.

peterNordin:
Also I tried to do  self.B+100 to assign mass inside a Mass component, but it did not work. I guess this is because numhop support for self. is still missing ?

peterNordin:
I made a PR #1899 with a bunch of tests that capture various usage of system parameters, if you re-base on this branch (now or after it is merged) we can see what things will break.

Preferably nothing should break, but for edge-cases a warning message may be sufficient, but who looks at those. If I have time I can look into fixing some of these issues, especially those regarding numhop scripts.

peterNordin:
Also, since the tests do not test GUI specific parts of the functionality. It is a good idea to open the "unittestmodel" in hopsangui manually and try to simulate it. It really does not catch all possible errors, but at least it will catch problems with up-converting of the model.

PR #1754, opened 2018-06-22
Release 2.9

peterNordin:
* Code fixes
* Release notes with new style


peterNordin:
This one may need updating when the two remaining issues have been solved (unless the are postponed)

Hopsan-release-notes.txt
```diff
@@ -26,15 +31,77 @@
...
+  - Script generation is now shown in a separate wizard.
+  - Reduced terminal output during optimization.
+  - New algorithm, Genetic Algorithm.
+
+- FMU:
+  - Exported Hopsan FMUs will now report Warning and Error messages. Info messages will be reported as Debug type.
+  - Imported FMUs will now have their messages redirected to the Hopsan core logger instead of standard output.
+    messages should now be visible inside HopsanGUI. Enable Debug messages to see FMU Info messages.
+
+- HopsanGUI:
```
robbr48: There are also some new HCOM functions which should be mentioned (ceil, floor, round and unco). Also possibility to filter parameters by values with dipa command.

Hopsan-release-notes.txt
```diff
@@ -26,15 +31,77 @@
...
+  - Results saved by HopsanCLI now use scientific notation, this may increase precision for small values.
+
+--- Bug Fixes ---
+
+- Optimization:
+  - Plot windows should be reusable.
+  - Progress bars do not show with DE algorithm.
+  - Excessive progress bars are not removed when reducing number of models.
+  - Total progress bar is not updated with PSO algorithm.
+  - Abort button does only work once.
```
robbr48: Actually this still happens, but less often than before. I think we can remove or reformulate this line.

Hopsan-release-notes.txt
peterNordin: There were some issues that were not set to mileatone 2.9.0, so I did not see then. I will add this information.

Hopsan-release-notes.txt
peterNordin: If it does not work then that issue should be reopened, i will remove "abort buton only work once" from the releas notes

PR #1758, opened 2018-06-29
Fixes to piston accumulator

robbr48:
Modifications by @petkr14


peterNordin:
We can add those later, merging now to prepare release

PR #1760, opened 2018-07-16
Make models know what libraries they require

peterNordin:
Also contains various cleanup in related code

### DIFF
--- a/HopsanGUI/GUIObjects/GUIContainerObject.cpp
+++ b/HopsanGUI/GUIObjects/GUIContainerObject.cpp
@@ -593,72 +593,65 @@ ModelObject* ContainerObject::addModelObject(QString fullTypeName, QPointF posit
 //! @todo only modelobjects for now
 ModelObject* ContainerObject::addModelObject(ModelObjectAppearance *pAppearanceData, QPointF position, double rotation, SelectionStatusEnumT startSelected, NameVisibilityEnumT nameStatus, UndoStatusEnumT undoSettings)
 {
-        //Deselect all other components and connectors
+    // Deselect all other components and connectors
     emit deselectAllGUIObjects();
     emit deselectAllConnectors();
 
+    ModelObject* pNewModelObject = nullptr;
     QString componentTypeName = pAppearanceData->getTypeName();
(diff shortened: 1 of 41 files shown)


HopsanGUI/GUIObjects/GUIContainerObject.cpp
```diff
@@ -3499,3 +3491,22 @@ LogDataHandler2 *ContainerObject::getLogDataHandler()
     return mpModelWidget->getLogDataHandler();
     //return mpLogDataHandler;
 }
+
+QStringList ContainerObject::getRequiredComponentLibraries() const
+{
+    QStringList requiredLibraryPaths;
```
robbr48: Why is it called "Paths"? Shouldn't it be "requiredLibraryIds"?

Models/Animation Validation/MachinesValidation.hmf
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <hopsanmodelfile hmfversion="0.4" hopsanguiversion="0.6.x_r7435" hopsancoreversion="0.6.x_r7431">
   <requirements>
-    <componentlibrary>DigitalHydraulicsLib</componentlibrary>
     <componentlibrary>HopsanDefaultComponentLibrary</componentlibrary>
```
robbr48: Why not use the library ID here, now that the default lib has one?

HopsanGUI/GUIObjects/GUIContainerObject.cpp
peterNordin: Yes this is leftover from before introducing id. I forgot to change name. I will fix before merging

Models/Animation Validation/MachinesValidation.hmf
peterNordin: I made this change before adding ID. Then I considered updating the models again. But decided not to (to much  work). This informatin is still valid for models saved with an older version of hopsan. The load code will show a debug message if the library name is not pressent. (Like before). Only difference is that now it wont warn for non needed libraries. 
The next time the models are updated (resaved) the new ID info will be added.

PR #1775, opened 2018-09-28
Autocompleter in HCOM terminal

robbr48:
Based on QCompleter. Autocompletes the following entries:
- Commands
- Functions
- Functionoids
- Model parameters
- Log variables
- Local variables

### DIFF
--- a/HopsanGUI/Widgets/HcomWidget.cpp
+++ b/HopsanGUI/Widgets/HcomWidget.cpp
@@ -39,7 +39,8 @@
 #include <QGridLayout>
 #include <QTextBlock>
 #include <QScrollBar>
-
+#include <QAbstractItemView>
+#include <QStringListModel>
 
 #include <cmath>
 
@@ -232,7 +233,12 @@ TerminalConsole::TerminalConsole(TerminalWidget *pParent)
     mCurrentHistoryItem=-1;
     this->append(">> ");
 
-    //mpErrorSound = new QSound(QString(SOUNDSPATH)+"error.wav");
+    mpCompleter = new QCompleter(this);
+    mpCompleter->setWidget(this);
+    mpCompleter->setCompletionMode(QCompleter::PopupCompletion);
+    mpCompleter->setCaseSensitivity(Qt::CaseInsensitive);
+    QObject::connect(mpCompleter, SIGNAL(activated(QString)),
+                     this, SLOT(insertCompletion(QString)));
 }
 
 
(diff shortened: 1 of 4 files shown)


peterNordin:
/HcomWidget.cpp: In member function ‘void TerminalConsole::updateAutoCompleteList()’:
../../HopsanGUI/Widgets/HcomWidget.cpp:618:31: error: expected type-specifier before ‘QStringListModel’
     mpCompleter->setModel(new QStringListModel(mpTerminal->mpHandler->getAutoCompleteWords(), this));

HopsanGUI/Widgets/HcomWidget.cpp
```diff
@@ -551,6 +595,30 @@ void TerminalConsole::keyPressEvent(QKeyEvent *event)
     }
 }
 
+//! @brief Inserts a selected completion from autocompleter
+void TerminalConsole::insertCompletion(const QString& completion)
+{
+    QString text = completion;
```
peterNordin: If the first thing we do is to take a copy of the const reference "completion", and then never use it again. We could instead let the argument be a non-const Qstring, not a big deal but removes the need for an extra text variable.

HopsanGUI/Widgets/HcomWidget.cpp
```diff
@@ -551,6 +595,30 @@ void TerminalConsole::keyPressEvent(QKeyEvent *event)
...
+    tc.movePosition(QTextCursor::Left);
+    tc.movePosition(QTextCursor::EndOfWord);
+    tc.insertText(text.right(extra));
+    setTextCursor(tc);
+}
+
+//! @brief Updates list of auto complete words
+void TerminalConsole::updateAutoCompleteList()
+{
+    mpCompleter->setModel(new QStringListModel(mpTerminal->mpHandler->getAutoCompleteWords(), this));
```
peterNordin: I think that maybe the parent should be the mpCompleter, otherwise the previous model is not deleted when setting a new.

HopsanGUI/Widgets/HcomWidget.cpp
```diff
@@ -39,7 +39,7 @@
 #include <QGridLayout>
 #include <QTextBlock>
 #include <QScrollBar>
-
+#include <QAbstractItemView>
```
peterNordin: Is this the correct include?

robbr48: Yes, it is needed for the QCompleter for some reason.

PR #1776, opened 2018-10-03
Hcom script editor

robbr48:
**Summary**
Built-in script editor for HCOM scripts

**Features** 
- Load and save HCOM scripts
- Syntax highlighting
- Auto-complete function
- Mouse wheel zooming
- Block comment / uncomment
- Printer support


HopsanGUI/Widgets/ScriptEditor.cpp
```diff
@@ -0,0 +1,407 @@
...
+        }
+        mSavedText = scriptFile.readAll();
+        mpEditor->setText(mSavedText);
+        scriptFile.close();
+    }
+
+    QGridLayout *pLayout = new QGridLayout(this);
+    pLayout->addWidget(mpEditor,         0,0,1,2);
+    this->setLayout(pLayout);
+
```
peterNordin: I dont think that you need to set the layout. Since this was the parent. In debug mode annoying warnings are shown that "widget already have a layout"

HopsanGUI/Widgets/ScriptEditor.h
```diff
@@ -0,0 +1,100 @@
...
+ the Hopsan source code root directory.
+
+-----------------------------------------------------------------------------*/
+
+//!
+//! @file   ScriptEditor.h
+//! @author Robert Braun <robert.braun@liu.se>
+//! @date   2018-10-01
+//! @version $Id$
+//!
```
peterNordin: Thr $Id$ was a subversion thing, dont know of there is an alternative in git. commit hash, is an alternative but not sure if really needed.

HopsanGUI/Widgets/ScriptEditor.h
```diff
@@ -0,0 +1,100 @@
...
+    QFileInfo getScriptFileInfo() const { return mScriptFileInfo; }
+    bool isSaved() const { return mIsSaved; }
+
+protected:
+    void wheelEvent(QWheelEvent* event);
+
+signals:
+
+private:
+    HcomEditor *mpEditor;
```
peterNordin: Later we should be able to use this for emedded script (numhop) editing as well, which is nice. We can do that later.

peterNordin:
The svg files for the icons are missing. (I assume that you made them in inkscape)

HopsanGUI/ModelHandler.cpp
```diff
@@ -244,6 +246,33 @@ void ModelHandler::loadModel()
...
 
 
+//! @brief Loads a HCOM script from a file and opens it in a new project tab.
+//! @see loadScriptFile(QString scriptFileName)
+void ModelHandler::loadScriptFile()
+{
+    QStringList scriptFileNames = QFileDialog::getOpenFileNames(gpMainWindowWidget, tr("Choose Script File"),
+                                                                gpConfig->getStringSetting(CFG_LOADSCRIPTDIR),
+                                                                tr("HCOM Script Files (*.hcom)"));
+    foreach(const QString &scriptFileName, scriptFileNames)
```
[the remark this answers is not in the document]
robbr48: I agree. This was copied from existing load function. There are a lot of "foreach", "Q_FOREACH" and "for(int i=0..." in the code which can be replaced by C++11 range based loops.

robbr48:
There are no svg icons. I modified the existing png files for "new" and "open" directly.

HopsanGUI/Widgets/ScriptEditor.h
robbr48: Yes, that should be possible. We can rename the class in this case.

PR #1779, opened 2018-10-08
Quickly add components in graphics view

robbr48:
Quickly add components by clicking anywhere in graphics view and typing component typename in a line edit. 

Needs more testing, especially for when line edit should be shown and not.

### DIFF
--- a/HopsanGUI/GraphicsView.cpp
+++ b/HopsanGUI/GraphicsView.cpp
@@ -720,7 +720,7 @@ void GraphicsView::mouseReleaseEvent(QMouseEvent *event)
         hideAddComponentLineEdit();
     }
 #if QT_VERSION >= 0x050000
-    else if(!mpContainerObject->isCreatingConnector() &&
+    else if(event->button() == Qt::LeftButton && !mpContainerObject->isCreatingConnector() &&
             mpContainerObject->getSelectedModelObjectPtrs().isEmpty() &&
             mpContainerObject->getSelectedGUIWidgetPtrs().isEmpty() &&
             !mpContainerObject->isConnectorSelected() &&
--- a/HopsanGUI/GraphicsView.cpp
+++ b/HopsanGUI/GraphicsView.cpp
@@ -143,8 +143,13 @@ void GraphicsView::contextMenuEvent ( QContextMenuEvent * event )
 
 void GraphicsView::insertComponentFromLineEdit()
 {
-    gpMessageHandler->addInfoMessage("Adding: "+mpAddComponentLineEdit->text());
-    mpContainerObject->addModelObject(mpAddComponentLineEdit->text(), mapToScene(mpAddComponentLineEdit->pos()));
+    QString displayName = mpAddComponentLineEdit->text();
+    if(!mDisplayNames.contains(displayName)) {
+        return;
+    }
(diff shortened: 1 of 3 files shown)


HopsanGUI/GraphicsView.cpp
```diff
@@ -660,6 +703,30 @@ void GraphicsView::mousePressEvent(QMouseEvent *event)
...
+            mpContainerObject->getSelectedModelObjectPtrs().isEmpty() &&
+            this->rubberBandRect().isEmpty())
+    {
+        QCursor cursor;
+        QPointF pos = mapFromGlobal(cursor.pos());
+        QRect tempRect(pos.x(), pos.y(),300,1);
+
+        QStringList typeNames = QStringList() << gpLibraryHandler->getLoadedTypeNames();
+        mpAddComponentLineEdit->completer()->setModel(new QStringListModel(typeNames,mpAddComponentLineEdit));
+        mpAddComponentLineEdit->setGeometry(pos.x(), pos.y(), 200,30);
```
peterNordin: I have no idea how to solve this, but hard-coding sizes in pixels may not work well on monitors with some other resolution. No need to fix this now, maybe deserve an issue of its own. Problem exist in many palces in Hopsan

robbr48: I agree, but I'm not sure if it's possible. Perhaps you could look at the size of the default font or something.

robbr48:
I will wait a little more with merging this, since there is a risk of annoyances if the pop-up appears too often or does not disappear when it should.

peterNordin:
I tried it now and mostly it works fine, but I have two comments.
1. If I write some text and then click somewhere else, the box will close, bu the next time i open it, the old text remains. Not sure what the intended behavior is. Abort with Esc, clears the text.
2. I am not sure if single click is the best way to activate (it works OK), but maybe double-click or an option "add component" in the right-click context menu would be better. Not sure though.

robbr48:
1. Yes, Text should be cleared when closing the dialog. I will fix it.
2. I think people are used to single left-clicking from the newer versions of Simulink.

peterNordin:
OK In this case I guess left click makes sense

peterNordin:
wait a minute... how is it that I am using this feature on the master branch !?. This PR has not yet been merged!?

robbr48:
@peterNordin Interesting. Somehow the commit was pushed in #1787. Perhaps I can make a new PR to undo the changes.

peterNordin:
No leave it. use this PR to fix any remaining issues, then we can pretend that the function came with this PR.
But this needs to be rebased on master.

robbr48:
Hmm, seems like the PR was automatically closed when I rebased, since there were no changes left.

peterNordin:
I think you can reopen it when you push new stuff

peterNordin:
@robbr48 what should we do about this one. Its easy to forgett now that it seem to be closed. Maybe push some dummy change?

robbr48:
I managed to re-open the PR by pushing a new commit. But now it is showing a lot of already merged commits for some reason.

robbr48:
@peterNordin I guess we can merge this for the 2.11.0 release, and then create a new PR if we encounter more problems?

PR #1780, opened 2018-10-08
Add trace variables from multiple ports in debugger

robbr48:
Quickly add trace variables from multiple selected ports.


HopsanGUI/Widgets/DebuggerWidget.cpp
```diff
@@ -218,29 +219,51 @@ void DebuggerWidget::updateVariablesList(QString port)
...
+        for(int i=0; i<items.size(); ++i)
+        {
+            variables.push_back(component+"#"+port+"#"+items[i]->text());
+        }
+    }
+
+    //Add all selected variables to table
+    for(const QString &variable : variables)
+    {
+        if(mVariables.contains(variable)) return;
```
peterNordin: Maybe continue instead of return

PR #1787, opened 2018-10-23
Improve find widget

### LINKED ISSUE #1783: Improve find widget in HopsanGUI
Find widget is quite different compared to find widgets other programs, and therefore feels unintuitive.

- Use a docket widget beneath the workspace instead of a popup dialog
- Open with Ctrl-F instead of Ctrl-Shift-F
- Have one line edit, and use radio buttons to choose what to search for
- Add checkbox for case-sensitive matching
- Use "contains" matching instead of wildcards (searching for "pump" should find "MyPump")
- Add "find next" and "find previous" buttons

robbr48:
Resolves #1783.

Improvements to HopsanGUI find widget:
- [x] Show find widget beneath the workspace instead of a popup dialog
- [x] Open with Ctrl-F instead of Ctrl-Shift-F
- [x] Have one line edit, for all searches
- [x] Use combo box for search targets (components, aliases or system parameters)
- [x] Checkbox for case-sensitive matching
- [x] Checkbox for wildcard matching
- [x] Use "contains" matching by default

Note: "Find next" and "Find previous" buttons are not implemented. They are not really needed with the highlighting features anyway.

### DIFF
--- a/HopsanGUI/Widgets/FindWidget.cpp
+++ b/HopsanGUI/Widgets/FindWidget.cpp
@@ -72,7 +72,7 @@ void FindHelper::doFind()
 
 
 FindWidget::FindWidget(QWidget *parent) :
-    QDialog(parent)
+    QWidget(parent)
 {
     FindHelper *pComponentFinder = new FindHelper(this);
     FindHelper *pAliasFinder = new FindHelper(this);
--- a/HopsanGUI/Widgets/FindWidget.cpp
+++ b/HopsanGUI/Widgets/FindWidget.cpp
@@ -36,64 +36,31 @@
 #include <QLabel>
 #include <QToolButton>
 #include <QPushButton>
-#include <QVBoxLayout>
-#include <QHBoxLayout>
-
-
-FindHelper::FindHelper(QWidget *pParent) :
-    QWidget(pParent)
-{
-    mpLineEdit = new QLineEdit(this);
-
-    QToolButton *pClearButton = new QToolButton(this);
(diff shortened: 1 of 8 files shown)


HopsanGUI/Widgets/FindWidget.cpp
```diff
@@ -36,64 +36,43 @@
...
-    pLayout->addWidget(new QLabel("Find System Parameter:", this));
-    pLayout->addWidget(pSystemparFinder);
-    pLayout->addWidget(pCloseButton,0,Qt::AlignRight);
-
-    connect(pComponentFinder, SIGNAL(find(QString)), this, SLOT(findComponent(QString)));
-    connect(pAliasFinder, SIGNAL(find(QString)), this, SLOT(findAlias(QString)));
-    connect(pSystemparFinder, SIGNAL(find(QString)), this, SLOT(findSystemParameter(QString)));
+    //FindHelper *pComponentFinder = new FindHelper(this);
+    //FindHelper *pAliasFinder = new FindHelper(this);
+    //FindHelper *pSystemparFinder = new FindHelper(this);
```
peterNordin: Remove the comments

peterNordin:
Anyway an other thing that you should fix is the orange highlight, it sets the effect on the modelobject but that may crash, as was seen in #1789, maybe it should be set on the mpIcon, I assume that all modelobjects have icons

peterNordin:
Also, this is a really good improvement of the find widget, the old one was not very nice

robbr48:
Ok, I will fix the highlighter in a different PR, to keep things clean.

PR #1792, opened 2018-11-04
Add compiler handler

peterNordin:
Adds generator for compiler command that should work with GCC and Clang
Only used in library import compilation for now. Will be extended to FMU export later

Related to #1711

### DIFF
--- a/HopsanGenerator/include/GeneratorTypes.h
+++ b/HopsanGenerator/include/GeneratorTypes.h
@@ -25,6 +25,8 @@
 #ifndef GENERATORTYPES_H
 #define GENERATORTYPES_H
 
+#include "hopsangenerator_win32dll.h"
+
 #include <QString>
 #include <QStringList>
 #include <QList>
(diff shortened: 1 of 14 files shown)


HopsanGenerator/include/GeneratorTypes.h
```diff
@@ -25,34 +25,73 @@
...
 
-#include "GeneratorUtilities.h"
+namespace hopsan {
 
+namespace os_strings {
+    constexpr auto win = "win";
+    constexpr auto win32 = "win32";
+    constexpr auto win64 = "win64";
+    constexpr auto apple = "apple";
+    constexpr auto Linux = "linux";
```
robbr48: Why capital letter on Linux but not on the others?

peterNordin:
Also I will extend the code later, and use it for FMU export and such. I just realized that FMU import is not tested, it indirectly uses this code. We need a test for that as well.

HopsanGenerator/include/GeneratorTypes.h
peterNordin: I will fix this later, the code needs some reworking anyway.

peterNordin:
FMU import seems to be broken from Hopsan GUI (Linux), compile works but not loading

peterNordin:
Fixed now (and tested), merging

PR #1796, opened 2018-11-06
False parameter not used warning

MarkusBagge:
Issue: A parameter used as part of an expression triggers a false warning for "parameter not used".
Solution: 
1) Prepend and append a space " " to expression (haystack) and parameter (needle).
2) Replace operators in haystack by space to avoid finding parameters which are substrings of other parameters.
3) Search for needle in haystack.
Before; expression was searched for in parameter name (I think?) so parameter "X" wasn't matched in for instance "3+X".

### DIFF
--- a/HopsanCore/src/ComponentSystem.cpp
+++ b/HopsanCore/src/ComponentSystem.cpp
@@ -1871,12 +1871,31 @@ bool ComponentSystem::checkModelBeforeSimulation()
                 break;
             }
 
-            // If a system parameter is used in the component, then erase it from the list
-            std::vector<HString>::iterator itp = std::find(unusedSysParNames.begin(), unusedSysParNames.end(), pCompParameters->at(p)->getValue());
-            if(itp != unusedSysParNames.end())
+            // If a system parameter is used in the component, then erase it from the list.
+            // To match only complete parameter names, prepend and append a space to the expression and replace operators by a space.
+            // Then find parameter names in expression.
+            HString expression = (" " + pCompParameters->at(p)->getValue() + " ")
+                    .replace("+", " ")
+                    .replace("-", " ")
+                    .replace("*", " ")
+                    .replace("/", " ")
+                    .replace("^", " ")
+                    .replace("(", " ")
+                    .replace(")", " ")
+                    .replace("<", " ")
+                    .replace(">", " ")
+                    .replace("|", " ")
+                    .replace("&", " ");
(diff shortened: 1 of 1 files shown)


HopsanCore/src/ComponentSystem.cpp
```diff
@@ -1871,12 +1871,22 @@ bool ComponentSystem::checkModelBeforeSimulation()
...
+            // To match only complete parameter names, prepend and append a space to the expression and replace operators by a space.
+            // Then find parameter names in expression.
+            HString expression = (" " + pCompParameters->at(p)->getValue() + " ")
+                    .replace("+", " ")
+                    .replace("-", " ")
+                    .replace("*", " ")
+                    .replace("/", " ")
+                    .replace("^", " ")
+                    .replace("(", " ")
+                    .replace(")", " ");
```
peterNordin: Numhop (should be called embedded script) also supports < > | & operators,
see https://github.com/peterNordin/libNumHop/blob/master/README.md

HopsanCore/src/ComponentSystem.cpp
```diff
@@ -1871,12 +1871,22 @@ bool ComponentSystem::checkModelBeforeSimulation()
...
+            for(std::vector<HString>::iterator itp = unusedSysParNames.begin(); itp!=unusedSysParNames.end();)
             {
-                unusedSysParNames.erase(itp);
+                (expression.find(" "+(*itp)+" ")!=std::string::npos) ? unusedSysParNames.erase(itp) : itp++;
```
peterNordin: I belive erasing from the container (vector) while looping over it is unsafe. I think that iterators may no longer be valid after erase. Does erase return an iterator ? Or is that c++11 (cant use that in hopsancore).

MarkusBagge: True. Made a new commit. Hopefully it will automatically show up here in some way?

HopsanCore/src/ComponentSystem.cpp
peterNordin: Yes, I need to update the documentation. And I agree, libnumhop should be able to extract variable names. I will add an issue to implement  that.

HopsanCore/src/ComponentSystem.cpp
peterNordin: Yes that should work

MarkusBagge:
Done. Hopefully. Beginner in this advanced git-stuff but I think you now have a single commit. Let me know if I messed up. Thanks.

PR #1797, opened 2018-11-10
SignalPID2 controller output limits

MarkusBagge:
- Added descriptions to Umin/Umin which refers to limits at which to perform anti-windup.
- Added controller output limits. Set separatelty from anti-windup limits. Acts as having a saturation block downstream a PID controller.
- Added messages if min/max limits are mixed up.

This is just a proposal. Most important would be the documentation part.

### DIFF
--- a/componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
+++ b/componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
@@ -47,7 +47,7 @@ private:
     bool mUseDeInput;
 
     // Constants
-    double mK,mTi,mTd,mTt,mUmax,mUmin;
+    double mK,mTi,mTd,mTt,mUmax,mUmin,mUoutmax,mUoutmin;
 
     // NodeData pointers
     double *mpErr, *mpDerr, *mpOut;
@@ -69,8 +69,10 @@ public:
         addConstant("Ti", "Integral time", "s", 1, mTi);
         addConstant("Tt", "Anti-windup tracking constant", "s", 1, mTt);
         addConstant("Td", "Derivative time", "s", 0, mTd);
-        addConstant("Umin", "", "", -1e100, mUmin);
-        addConstant("Umax", "", "", 1e100, mUmax);
+        addConstant("Umin", "Lower output for anti-windup", "", -1e100, mUmin);
+        addConstant("Umax", "Upper output for anti-windup", "", 1e100, mUmax);
+        addConstant("Uoutmin", "Minimum output limit", "", -1e100, mUoutmin);
+        addConstant("Uoutmax", "Maximum output limit", "", 1e100, mUoutmax);
     }
 
     void initialize()
@@ -85,7 +87,16 @@ public:
         {
             addWarningMessage("Tt is lower then Ti this is not correct!");
         }
+        if (mUmin > mUmax)
+        {
+            addErrorMessage("Umin must be <= Umax!");
+        }
+        if (mUoutmin > mUoutmax)
+        {
+            addErrorMessage("Uoutmin must be <= Uoutmax!");
+        }
     }
+
     void simulateOneTimestep()
     {
         double dErr;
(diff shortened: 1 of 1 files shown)


componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
```diff
@@ -69,8 +69,10 @@ class SignalPID2 : public ComponentSignal
         addConstant("Ti", "Integral time", "s", 1, mTi);
         addConstant("Tt", "Anti-windup tracking constant", "s", 1, mTt);
         addConstant("Td", "Derivative time", "s", 0, mTd);
-        addConstant("Umin", "", "", -1e100, mUmin);
-        addConstant("Umax", "", "", 1e100, mUmax);
+        addConstant("Umin", "Minimum output for anit-windup", "", -1e100, mUmin);
```
peterNordin: Maybe "Lower anti-windup limit" and "Upper anti-windup limit", as the anti-windup think is an internal thing. Someone not accustom to anti-windup may misinterpret the work output in this case.

componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
```diff
@@ -85,7 +87,16 @@ class SignalPID2 : public ComponentSignal
...
             addWarningMessage("Tt is lower then Ti this is not correct!");
         }
+        if (mUmin > mUmax)
+        {
+            addErrorMessage("Umin must be <= Umax!");
+        }
+        if (mUoutmin > mUoutmax)
+        {
+            addErrorMessage("Uoutmin must be <= Uoutmax!");
+        }
```
peterNordin: Good idea to add these checks. Maybe we should also check that Umin/max lies between the bound of Uoutmin/max. Because I do not think that you ever want to have Umax to ba larger then Uoutmax, that would cause windup risk. What do you think@robbr48

componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
```diff
@@ -109,27 +120,27 @@ class SignalPID2 : public ComponentSignal
...
         }
         else if (v < mUmin)
         {
-            u = mUmin;
+            mI = mI + mTimestep/mTt * (mUmin-v);
         }
-        else
+
+        // Apply controller output limits.
+        if (v > mUoutmax)
```
peterNordin: (minor note) The start { should be on the same line as the if () or the one after else if () should be on the next line. Right now there is a mix of two different styles.

peterNordin:
Yes I will try to write the documentation (later).  It is possible to have a markdown file that is rendered to html. Currently only the "laminar orifice" has one.

componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
MarkusBagge: I didn't add it so that you could disable anti-windup if you, for some reason I cannot come up with , want. Maybe Warning or Info message rather than Error? We may also silently do e.g. `mUmax=min(mUmax, mUoutmax)` plus a message. Just let me know what you prefer.

componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
MarkusBagge: Thanks for the notice. Personally prefer to have 
```
if () {
} else if() {
} else {
}
```

over
```
if ()
{
}
else if()
{
}
else
{
}
```

...but I will go for the latter, like in https://flumes.iei.liu.se/hopsan/docs/latest/html/page_codingconventions.html. This is what you mean right?

componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
MarkusBagge: I referred to the "output" to distinguish it from having anything to do with maximum/minimum value of the integrator part. I.e. the parameter does not limit the size of the integrator part, it just limits the (output) conditions under which it's allowed to "work". Again, @peterNordin and @robbr48 just let me know what you prefer.

componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
robbr48: I always use the lower version in Hopsan and the upper version in my second projects, because that is the coding style we agreed on :)

MarkusBagge: Ok, it's no deal breaker so I'll stick to the Hopsan style until guidelines changes. But you've got my  vote for it :)

componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
peterNordin: Next to naming, deciding this sort of thing is the most difficult thing. First I was think that we should silently limit the value internally, but then that would be strange, if parameters you have set, or not, would be ignored. Lets just leave it in the current state (no further change needed), after-all it is a simulation of a controller and you may wast to do strange things.

componentLibraries/defaultLibrary/Signal/Control/SignalPID2.hpp
peterNordin: I still think that the description is confusing, but you are correct, only proper documentation can address this and I have no better idea myself. I might adjust the description slightly when I get around to writing that documentation.

PR #1798, opened 2018-11-10
Base unit Ohm 

MarkusBagge:
Some model uses unit ohm but there is no corresponding quantity for this. Now quantity Resistance is added. 
Unit "Ohm" used instead if "ohm" for readability in case of prefix; mOhm, kOhm etc.
Note that one may also expect ohm to be impedance or reluctance. Those (aliases?) are not added.

### DIFF
--- a/HopsanCore/src/Quantities.cpp
+++ b/HopsanCore/src/Quantities.cpp
@@ -61,6 +61,8 @@ hopsan::QuantityRegister::QuantityRegister()
 
     registerQuantity("Temperature", "K");
 
+    registerQuantity("Resistance", "ohm");
+
     // Register quantity aliases
     registerQuantityAlias("Position", "Length");
 }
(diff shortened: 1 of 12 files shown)


peterNordin:
Actually the .nb files and generated component code (.hpp) that was changed should maybe use the new "Resistance" quantity instead of "ohm".

MarkusBagge:
Makes sense. Nothing against using the "ugly" names as long as we are standardized/consistent. And if we are, is "visual name" needed? Isn't it about doing string replace and change it to a predefined LaTeX or utf8 string then? Else, the "visual name" also needs to be added to the custom units dialogue too. Maybe a separate issue for the addition of "visual name"?

MarkusBagge:
Did a forced push. Now using "Resistance" instead of "Ohm" now. Also using "megohm" etc instead of "MOhm".  Took back changes on .nb files because they were treated as binary and I perhaps I mess something up if changing them. But I guess they do need to be updated somehow to keep them in sync?

peterNordin:
Yes the "visual name" can be added later, and you are right .nb wiles should be changed by someone with access to Mathematica such as @petkr14

MarkusBagge:
No worries at all. I have updated to ohm in model .hmf files and Resistance in component .hpp files. New forced commit pushed.

PR #1810, opened 2018-12-06
Fixes to animation real-time plotting

### LINKED ISSUE #1805: Setting for scope behaviour in animation
Real-time plotting during animation is hard-coded to only store last 3 seconds. 

Make this a model setting. It should be possible to have no limit.

Removing limit completely.

### LINKED ISSUE #1806: Crash when double-clicking scope in real-time animation and logging data to disk
**Describe the bug**
Crash when double-clicking a scope component during real-time animation if cache data to disk is enabled.

**To Reproduce**
1. Add a scope to the model
2. Enable cache log data to disk
3. Start real-time animation
4. Double-click on the scope

**Expected behavior**
It should open a plot window and begin real-time plotting.

robbr48:
Crash when real-time plotting with logging to disk enabled occurred because no generation existed in LogDataHandler before model was simulated first time. 

- Require model to be simulated before starting real-time animation. Resolves #1806.
- Removed time limit for real-time plotting. Resolves #1805.

### DIFF
--- a/HopsanGUI/Widgets/AnimationWidget.cpp
+++ b/HopsanGUI/Widgets/AnimationWidget.cpp
@@ -183,6 +183,7 @@ AnimationWidget::AnimationWidget(QWidget *parent) :
     //mpContainer->collectPlotData();
     mpPlotData = mpContainer->getLogDataHandler();
     mpPlayButton->setDisabled(mpPlotData->isEmpty());
+    mpPlayRealTimeButton->setDisabled(mpPlotData->isEmpty());
     mpRewindButton->setDisabled(mpPlotData->isEmpty());
 
     if(!mpPlotData->isEmpty() && !mpPlotData->getTimeVectorVariable(-1).isNull())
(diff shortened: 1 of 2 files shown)


HopsanGUI/Widgets/AnimationWidget.cpp
```diff
@@ -183,6 +183,7 @@ AnimationWidget::AnimationWidget(QWidget *parent) :
     //mpContainer->collectPlotData();
     mpPlotData = mpContainer->getLogDataHandler();
     mpPlayButton->setDisabled(mpPlotData->isEmpty());
+    mpPlayRealTimeButton->setDisabled(mpPlotData->isEmpty());
```
peterNordin: Hmm, it feels like this is not the actual solution to the problem, because why does "real-time" animation require that at least one logdata generation exiats? It should not need that I assume. But probably some code assumes that logdata allways exists.

robbr48: Real-time animation itself does not require any logdata. The problem is real-time plotting, i.e. when you double-click on a scope during real-time animation. This creates temporary variables and inserts them into the current generation. If there is no current generation, it will crash.

robbr48: But I agree that this is not the actual solution. It would be better if you could create a new empty  generation if it does not exist.

robbr48: I will merge this PR now to prevent crashes, and then make a separate PR about the actual log generation problem.

PR #1816, opened 2018-12-13
HCOM: Change diagram size command

### LINKED ISSUE #1223: Add chds HCOM commands
---

Original Redmine Issue: 1223, https://flumes.iei.liu.se/redmine/issues/1223
Original Date: 2014/05/16

---

chdl = chage diagram limits
chds = change diagram size

robbr48:
Resolves #1223.

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -479,6 +479,15 @@ void HcomHandler::createCommands()
     adpvrCmd.group = "Plot Commands";
     mCmdList << adpvrCmd;
 
+    HcomCommand chdsCmd;
+    chdsCmd.cmd = "chds";
+    chdsCmd.description.append("Change diagram size (and position)");
+    chdsCmd.help.append(" Usage: chds [width] [height]\n");
+    chdsCmd.help.append(" Usage: chds [x] [y] [width] [height]\n");
+    chdsCmd.fnc = &HcomHandler::executeChangeDiagramSizeCommand;
+    chdsCmd.group = "Plot Commands";
+    mCmdList << chdsCmd;
+
     HcomCommand chdlCmd;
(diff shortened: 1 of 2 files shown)


HopsanGUI/HcomHandler.cpp
```diff
@@ -1490,6 +1499,36 @@ void HcomHandler::executeAddPlotRightAxisCommand(const QString cmd)
     changePlotVariables(cmd, 1, true);
 }
 
+void HcomHandler::executeChangeDiagramSizeCommand(const QString cmd)
+{
+    QStringList args = splitCommandArguments(cmd);
+
+    if (args.size() != 2 && args.size() != 4) {
+        HCOMERR("Wrong number of arguments in chds, should be 4");
```
peterNordin: two or four ?

HopsanGUI/HcomHandler.cpp
```diff
@@ -1490,6 +1499,36 @@ void HcomHandler::executeAddPlotRightAxisCommand(const QString cmd)
...
+        return;
+    }
+
+    if (mpCurrentPlotWindow) {
+        bool ok;
```
peterNordin: Ok is not checked so invalid values like 'dog' will be converted to zero and accepted.
Actually I think that the input args should be evaluated, then it would be possible to control size programatically in scripts. I have added such code before in other commands... or maybe it was in functions... not sure.

peterNordin: Also what happens if size or position is negative.

robbr48: Good points, I have added SymHop evaluation and sign checks. There also seems to be a built-in check in Qt that prevents moving the window outside the screen.

We should probably add evaluation in all HCOM commands that takes numerical arguments.

PR #1817, opened 2018-12-14
Make save/restore state work for recompiling/reloading libraries

robbr48:
The major change here is to use a shared pointer for the LogDataHandler in ModelWidget. In this way, it can be kept alive when the model is closed and then re-inserted into the model after it is loaded again.

### DIFF
--- a/HopsanGUI/GUIObjects/AnimatedComponent.cpp
+++ b/HopsanGUI/GUIObjects/AnimatedComponent.cpp
@@ -393,7 +393,7 @@ void AnimatedComponent::updateAnimation()
 
     if(mpModelObject->getTypeName() == HOPSANGUISCOPECOMPONENTTYPENAME)
     {
-        LogDataHandler2 *pHandler = mpModelObject->getParentContainerObject()->getLogDataHandler();
+        LogDataHandler2* pHandler = mpModelObject->getParentContainerObject()->getLogDataHandler().data();
 
         QList<SharedVectorVariableT> vectors;
 
@@ -912,7 +912,7 @@ void AnimatedIcon::mouseDoubleClickEvent(QGraphicsSceneMouseEvent *event)
     //Open plot window if double-clicking on scope component
     if(mpAnimatedComponent->mpModelObject->getTypeName() == HOPSANGUISCOPECOMPONENTTYPENAME)
     {
-        LogDataHandler2 *pHandler = mpAnimatedComponent->mpModelObject->getParentContainerObject()->getLogDataHandler();
+        LogDataHandler2* pHandler = mpAnimatedComponent->mpModelObject->getParentContainerObject()->getLogDataHandler().data();
         QString name = mpAnimatedComponent->mpModelObject->getName();
         PlotWindow *pPlotWindow = mpAnimatedComponent->mpPlotWindow;
 
--- a/HopsanGUI/ModelHandler.cpp
+++ b/HopsanGUI/ModelHandler.cpp
@@ -41,6 +41,7 @@
 #include "Configuration.h"
 #include "DesktopHandler.h"
 #include "GraphicsView.h"
+#include "GraphicsViewPort.h"
 #include "GUIObjects/GUISystem.h"
 #include "MessageHandler.h"
 #include "MainWindow.h"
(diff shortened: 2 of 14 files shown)


HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -175,7 +175,8 @@ ModelWidget::~ModelWidget()
     delete mpAnimationWidget;
     createOrDestroyToplevelSystem(false);
     mpSimulationThreadHandler->deleteLater();
-    mpLogDataHandler->deleteLater();
+    mpLogDataHandler->setParent(nullptr);
+    mpLogDataHandler.clear();
```
peterNordin: This might be risky, clear will delete the object immediately rather then "later" (if its the last one) but this can be solved when the object is created (see other comment)

HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -121,7 +121,7 @@ ModelWidget::ModelWidget(ModelHandler *pModelHandler, CentralTabWidget *pParentT
     mpGraphicsView  = new GraphicsView(this);
     //mpGraphicsView->setScene(mpToplevelSystem->getContainedScenePtr());
 
-    mpLogDataHandler = new LogDataHandler2(this);
+    mpLogDataHandler = QSharedPointer<LogDataHandler2>(new LogDataHandler2(this));
```
peterNordin: To enure that the object is deleted later, supply deleter (I think that we want that). See https://stackoverflow.com/a/12624043

HopsanGUI/GUIObjects/AnimatedComponent.cpp
```diff
@@ -393,7 +393,7 @@ void AnimatedComponent::updateAnimation()
 
     if(mpModelObject->getTypeName() == HOPSANGUISCOPECOMPONENTTYPENAME)
     {
-        LogDataHandler2 *pHandler = mpModelObject->getParentContainerObject()->getLogDataHandler();
+        LogDataHandler2* pHandler = mpModelObject->getParentContainerObject()->getLogDataHandler().data();
```
peterNordin: Not sure if this must be fixed now, as it is a lot of work. But in all these places you add .data() to convert back to a raw pointer. It will work as before, but it would be better to rewrite the code to make use of the shared pointer (or maybe a weak pointer it Qt has such a thing). This feels like cheating.... well maybe somthing for some later time......

HopsanGUI/ModelHandler.cpp
```diff
@@ -212,14 +213,14 @@ ContainerObject *ModelHandler::getCurrentViewContainerObject()
...
-LogDataHandler2 *ModelHandler::getCurrentLogDataHandler()
+QSharedPointer<LogDataHandler2> ModelHandler::getCurrentLogDataHandler()
 {
     ModelWidget *pMW = getCurrentModel();
     if (pMW)
     {
         return pMW->getLogDataHandler();
     }
-    return 0;
+    return QSharedPointer<LogDataHandler2>(nullptr);
```
peterNordin: Not that it makes any difference, but in this case returning nullptr directly should work since the type is known. But an alternative is to use the universal initializer (I think its called that) {}  ... (no need to change)

HopsanGUI/ModelHandler.cpp
```diff
@@ -635,9 +636,9 @@ void ModelHandler::refreshMainWindowConnections()
     if (pCurrentModel)
     {
         gpMainWindow->mpToggleRemoteCoreSimAction->setChecked(pCurrentModel->getUseRemoteSimulationCore());
-        gpMainWindow->mpDataExplorer->setLogdataHandler(pCurrentModel->getLogDataHandler());
-        gpMainWindow->mpImportDataFileAction->setEnabled(pCurrentModel->getLogDataHandler()!=nullptr);
-        gpPlotWidget->setLogDataHandler(pCurrentModel->getLogDataHandler());
+        gpMainWindow->mpDataExplorer->setLogdataHandler(pCurrentModel->getLogDataHandler().data());
+        gpMainWindow->mpImportDataFileAction->setEnabled(pCurrentModel->getLogDataHandler().data() != nullptr);
```
peterNordin: I think qsharedpointer has a isNull() function or similar

HopsanGUI/ModelHandler.cpp
robbr48: "return 0" worked fine on my computer, but failed both on Travis and AppVeyor. I guess it's a Qt4 problem.

HopsanGUI/Widgets/ModelWidget.cpp
robbr48: Good point, I will fix it.

PR #1818, opened 2018-12-18
Merge HoLC text editor into HopsanGUI

robbr48:
This will be used for modifying or creating component libraries from within HopsanGUI.
- [X] Support all languages/formats (HCOM, XML, C++, Modelica, Python)
- [X] Highlighting
- [X] Line numbering
- [X] Find function
- [X] Block commenting
- [x] Auto-complete (HCOM/C++)

### DIFF
--- a/HopsanGUI/Utilities/HighlightingUtilities.cpp
+++ b/HopsanGUI/Utilities/HighlightingUtilities.cpp
@@ -336,8 +336,14 @@ CppHighlighter::CppHighlighter(QTextDocument *parent)
 {
     HighlightingRule rule;
 
-    keywordFormat.setForeground(Qt::darkYellow);
-    keywordFormat.setFontWeight(QFont::Normal);
+    //mFunctionFormat.setFontWeight(75);
+    mFunctionFormat.setForeground(Qt::black);
+    rule.pattern = QRegExp("\\b[A-Za-z0-9_]+(?=\\()");
+    rule.format = mFunctionFormat;
+    mHighlightingRules.append(rule);
+
+    mKeywordFormat.setForeground(Qt::darkYellow);
+    mKeywordFormat.setFontWeight(QFont::Normal);
     QStringList keywordPatterns;
     keywordPatterns << "\\bchar\\b" << "\\bclass\\b" << "\\bconst\\b"
                     << "\\bdouble\\b" << "\\benum\\b" << "\\bexplicit\\b"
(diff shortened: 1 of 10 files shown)


HopsanGUI/Utilities/HighlightingUtilities.cpp
```diff
@@ -348,60 +354,74 @@ CppHighlighter::CppHighlighter(QTextDocument *parent)
                     << "\\bslots\\b" << "\\bstatic\\b" << "\\bstruct\\b"
                     << "\\btemplate\\b" << "\\btypedef\\b" << "\\btypename\\b"
                     << "\\bunion\\b" << "\\bunsigned\\b" << "\\bvirtual\\b"
-                    << "\\bvoid\\b" << "\\bvolatile\\b" << "\\busing\\b";
-    foreach (const QString &pattern, keywordPatterns) {
+                    << "\\bvoid\\b" << "\\bvolatile\\b" << "\\busing\\b"
```
peterNordin: Should this list really contain Qt keywords, foreach and slots (not that it really matters). Also constexpr should be added. (C++11 so cant be used in hopsan components... well it can if a user make own components and want c++11)

HopsanGUI/Utilities/HighlightingUtilities.cpp
```diff
@@ -348,60 +354,74 @@ CppHighlighter::CppHighlighter(QTextDocument *parent)
...
-        highlightingRules.append(rule);
+        rule.format = mKeywordFormat;
+        mHighlightingRules.append(rule);
+    }
+
+    mHopsanKeywordFormat.setForeground(Qt::darkMagenta);
+    mHopsanKeywordFormat.setFontWeight(QFont::Normal);
+    QStringList hopsanKeywordPatterns;
+    hopsanKeywordPatterns << "\\bPort\\b" << "\\bFirstOrderTransferFunctionVariable\\b" << "\\bSecondOrderTransferFunctionVariable\\b"
+                          << "\\bFirstOrderTransferFunction\\b" << "\\bSecondOrderTransferFunction\\b"
```
peterNordin: This list should be longer, can say right now what is missing. Maybe we should use external text file with keywords to make it easier to add new ones.

HopsanGUI/Utilities/HighlightingUtilities.cpp
```diff
@@ -348,60 +354,74 @@ CppHighlighter::CppHighlighter(QTextDocument *parent)
...
     }
 
-    preProcessorFormat.setForeground(Qt::darkBlue);
-    preProcessorFormat.setFontWeight(QFont::Normal);
+    mPreProcessorFormat.setForeground(Qt::darkBlue);
+    mPreProcessorFormat.setFontWeight(QFont::Normal);
     keywordPatterns.clear();
     keywordPatterns << "^#?\\binclude\\b" << "^#?\\bifdef\\b" << "^#?\\bifndef\\b"
                     << "^#?\\belseif\\b" << "^#?\\belse\\b" << "^#?\\bendif\\b";
-    foreach (const QString &pattern, keywordPatterns) {
```
peterNordin: One could also write 
if defined SOMETHING
if defined(SOMETHING)
same with else
Could be multiple whitespaces in there aswell so regexp maching is needed

HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -0,0 +1,682 @@
...
+{
+    mFileInfo = scriptFileInfo;
+
+    mpEditor = new TextEditor(highlighter, this);
+    QFont font("Monospace");
+    font.setStyleHint(QFont::TypeWriter);
+    mpEditor->setFont(font);
+
+    mpHcomHighlighter = new HcomHighlighter(nullptr);
+    mpXmlHighlighter = new XmlHighlighter(nullptr);
```
peterNordin: Are these ever deleted?
Should use std::unique_ptr instead of raw pointer to solve that automatically

HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -0,0 +1,682 @@
...
+    setExtraSelections(extraSelections);
+}
+
+
+
+//! @brief Update function for line number area
+void TextEditor::updateLineNumberArea(const QRect &rect, int dy)
+{
+    if (dy)
+        mpLineNumberArea->scroll(0, dy);
```
peterNordin: Missing braces on these lines, also maybe clarify dy!=0, also add documentation line for dy

HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -0,0 +1,682 @@
...
+        {
+            if(line.simplified().startsWith("class "))
+                bracketCounter = 0;
+
+            bracketCounter += line.count("{");
+            bracketCounter -= line.count("}");
+            if(bracketCounter != 1)
+                continue;
+
+            if(line.contains("()")) //Ignore functions
```
peterNordin: What if function has arguments

HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -0,0 +1,682 @@
...
+            variables[v].remove("*");
+            variables[v].remove(";");
+            for(int d=0; d<dataTypes.size(); ++d)
+            {
+                variables[v].remove(dataTypes[d]+" ");
+            }
+            variables[v].remove(" ");
+            while(variables[v].contains("["))
+            {
+                variables[v].remove("["+variables[v].section("[",1,1).section("]",0,0)+"]");
```
peterNordin: What does this black magic do?
maybe create a named lambda function that do this thing to make the code easier to understand

HopsanGUI/Widgets/TextEditorWidget.h
```diff
@@ -35,50 +35,98 @@
...
 private:
-    HcomEditor *mpEditor;
-    QFileInfo mScriptFileInfo;
+    void setHighlighter(HighlighterTypeEnum highlighter);
+
+    TextEditor *mpEditor;
+    QFileInfo mFileInfo;
     bool mIsSaved = true;
     QString mSavedText;
+    HcomHighlighter *mpHcomHighlighter;
```
peterNordin: These should probably be std::unique_ptr see other comment

HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -0,0 +1,682 @@
...
+
+
+//! @brief Updates list of auto complete words
+void TextEditor::updateAutoCompleteList()
+{
+    if(mLanguage == HighlighterTypeEnum::Hcom) {
+        mpCompleter->setModel(new QStringListModel(gpTerminalWidget->mpHandler->getAutoCompleteWords(), mpCompleter));
+    }
+    else if(mLanguage == HighlighterTypeEnum::Cpp) {
+        QStringList lines = toPlainText().split("\n");
```
peterNordin: Ok maybe not so common, but one may write multiple lines on thre same line
maybe should also split on that somehow, 

double var; int othervar;

But difficult since for (a; b; c) also contains mutliple ; on same line (but inside parenthesis ()

HopsanGUI/Widgets/TextEditorWidget.cpp
robbr48: In that case I believe it is easier to just delete them in the destructor of the parent class.

HopsanGUI/Utilities/HighlightingUtilities.cpp
robbr48: True, but I will ignore this for now. This only affects the highlighting anyway, so the code will work regardless.

HopsanGUI/Widgets/TextEditorWidget.cpp
robbr48: It removes brackets from array variables, i.e. converts "myArray[5]" to "myArray". I added a comment to clarify.

HopsanGUI/Widgets/TextEditorWidget.cpp
robbr48: Not sure actually, but it does not seem to be a problem. Better leave this as it is I guess. This line may not even be necessary.

HopsanGUI/Widgets/TextEditorWidget.cpp
robbr48: True, but I think we can survive without that feature at the moment :)

HopsanGUI/Utilities/HighlightingUtilities.cpp
robbr48: I removed the Qt keywords and added constexpr.

HopsanGUI/Widgets/TextEditorWidget.h
robbr48: I added deleteLater() in constructor, as mentioned above.

HopsanGUI/Widgets/TextEditorWidget.cpp
robbr48: Done!

HopsanGUI/Utilities/HighlightingUtilities.cpp
robbr48: I added all component utilities. There are probably more things we can add later.

robbr48:
@peterNordin are my changes ok?

PR #1819, opened 2018-12-20
Issue 1808 add sys parameters from expression

### LINKED ISSUE #1808: System parameteres in expressions are not added when pasting components
**Describe the bug**
When pasting a component with a parameter mapped to a system parameter, the system parameter is added to the system if it does not already exist. However, if the system parameter is used in a NumHop expression (e.g. "-x"), it is not added.

**To Reproduce**
*If the bug can be reproduce, please give the steps taken to reproduce the behavior:*
1. Add component A
2. Add system paramter X
3. Set a parameter in A to -X
4. Copy component A
5. Create a new model
6. Paste component A
7. It is not possible to simulate the model, since "X" is not available

**Expected behavior**
X should be added to the target model.

peterNordin:
Use numhop expression to extract variable names

**Note!** Update of submodule numhop needed, also renamed libNumHop and IndexingCSVParser to lower case. May need "git submodule update --init" or "git submodule sync"

Resolves #1808

### DIFF
--- a/HopsanGUI/GUIObjects/GUIContainerObject.cpp
+++ b/HopsanGUI/GUIObjects/GUIContainerObject.cpp
@@ -1508,11 +1508,22 @@ void ContainerObject::copySelected(CopyStack *xmlStack)
                 if(thisSystemsParameterNames.contains(var)) {
                     CoreParameterData parData;
                     getParameter(var, parData);
-                    QDomElement parElement = appendDomElement(*copyRoot, "parameter");
-                    parElement.setAttribute("name", parData.mName);
-                    parElement.setAttribute("value", parData.mValue);
-                    parElement.setAttribute("type", parData.mType);
-                    //! @todo copy all data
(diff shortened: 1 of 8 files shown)


HopsanGUI/GUIObjects/GUIContainerObject.cpp
```diff
@@ -1486,30 +1486,45 @@ void ContainerObject::copySelected(CopyStack *xmlStack)
     appendCoordinateTag(*copyRoot, center.x(), center.y());
 
     // Copy components
+    const auto thisSystemsParameterNames = getParameterNames();
```
robbr48: I'm not sure if I like this variable name. Can't it just be called "systemParameterNames"?

peterNordin: When I tried to understand my old dusty code to fix this the name confused me. Since this code deals with figuring out what system parameters are used. This new name is more clear and may reduce confusion the next time someone tries to understand that code. But yes the name is silly

PR #1826, opened 2019-01-18
Open component source code HopsanGUI

robbr48:
- Open base component sourcecode for subtype components
- Open source code from context menu
- Open source code from library widget
- Open source code from component properties dialog
- Fixed crash in restore state when text tabs are open
- Fixed wrong tab name after restore state when text tabs are open
- Prevent opening same text file in multiple tabs

### DIFF
--- a/HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
+++ b/HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
@@ -70,6 +70,7 @@
 #include "LibraryHandler.h"
 #include "Widgets/LibraryWidget.h"
 #include "MessageHandler.h"
+#include "ModelHandler.h"
 
 #ifdef USEDISCOUNT
 extern "C" {
@@ -259,6 +260,17 @@ void ComponentPropertiesDialog3::applyPressed()
     setVariableValues();
 }
 
+void ComponentPropertiesDialog3::openSourceCode()
+{
+    auto appearance = gpLibraryHandler->getModelObjectAppearancePtr(mpModelObject->getTypeName());
+    QString basePath = appearance->getBasePath();
+    if(!basePath.isEmpty()) {
+        basePath.append("/");
+    }
+    QString fileName = appearance->getSourceCodeFile();
+    gpModelHandler->loadTextFile(basePath+fileName);
+}
+
 
 void ComponentPropertiesDialog3::editPortPos()
 {
@@ -425,6 +437,13 @@ QDialogButtonBox *ComponentPropertiesDialog3::createOKButtonBox()
     QPushButton *pOkButton = new QPushButton(tr("&Ok"), this);
     QPushButton *pApplyButton = new QPushButton(tr("&Apply"), this);
     QDialogButtonBox *pButtonBox = new QDialogButtonBox(Qt::Horizontal, this);
+    QString filePath = mpModelObject->getAppearanceData()->getSourceCodeFile();
+    if(!filePath.isEmpty())
+    {
+        QPushButton *pSourceCodeButton = new QPushButton(tr("&Source Code"), this);
+        pButtonBox->addButton(pSourceCodeButton, QDialogButtonBox::ActionRole);
+        connect(pSourceCodeButton, SIGNAL(clicked()), this, SLOT(openSourceCode()));
+    }
     pButtonBox->addButton(pApplyButton, QDialogButtonBox::ActionRole);
     pButtonBox->addButton(pOkButton, QDialogButtonBox::ActionRole);
     pButtonBox->addButton(pCancelButton, QDialogButtonBox::ActionRole);
@@ -434,6 +453,7 @@ QDialogButtonBox *ComponentPropertiesDialog3::createOKButtonBox()
     pApplyButton->setEnabled(mAllowEditing);
     pOkButton->setEnabled(mAllowEditing);
     pOkButton->setDefault(true);
+
     return pButtonBox;
 }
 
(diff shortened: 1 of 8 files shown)


HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
```diff
@@ -259,6 +260,14 @@ void ComponentPropertiesDialog3::applyPressed()
     setVariableValues();
 }
 
+void ComponentPropertiesDialog3::openSourceCode()
+{
+    auto& appearance = gpLibraryHandler->getModelObjectAppearancePtr(mpModelObject->getTypeName());
```
peterNordin: This looks strange, the function returns a shared pointer. Here a reference to the shared pointer object is created. But then the temporary sharedpointer is destroyed (I assume). Then what does the reference refere to.
I think that the & should be removed.

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
```diff
@@ -259,6 +260,14 @@ void ComponentPropertiesDialog3::applyPressed()
...
+    QString basePath = appearance->getBasePath();
+    QString fileName = appearance->getSourceCodeFile();
+    gpModelHandler->loadTextFile(basePath+fileName);
```
peterNordin: I dont think that we can assume that basepath includes a trailing /, in my opinion it should not. So a / must br added, but only if basePath is not empty... we need some  clever helpfunction for this. But I assumt that the currnt  code works for some reason, maybe basepath has a trailing /

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
```diff
@@ -425,6 +434,13 @@ QDialogButtonBox *ComponentPropertiesDialog3::createOKButtonBox()
     QPushButton *pOkButton = new QPushButton(tr("&Ok"), this);
     QPushButton *pApplyButton = new QPushButton(tr("&Apply"), this);
     QDialogButtonBox *pButtonBox = new QDialogButtonBox(Qt::Horizontal, this);
+    QString filePath = mpModelObject->getAppearanceData()->getSourceCodeFile();
```
peterNordin: In this case basePath is not used, maybe it should be. Or maybe getSourceCodeFile shoyld return the a solute path

HopsanGUI/GUIObjects/GUIModelObject.cpp
```diff
@@ -1344,6 +1346,13 @@ QAction *ModelObject::buildBaseContextMenu(QMenu &rMenu, QGraphicsSceneContextMe
     {
         openPropertiesDialog();
     }
+    else if(selectedAction == sourceCodeAction)
+    {
+        auto& appearance = gpLibraryHandler->getModelObjectAppearancePtr(mModelObjectAppearance.getTypeName());
```
peterNordin: Reference again, see other comment
also  below, using basePath without /, not sure if it is actually a problem

HopsanGUI/GUIObjects/GUIModelObjectAppearance.cpp
```diff
@@ -820,6 +820,12 @@ QString ModelObjectAppearance::getBasePath() const
     return mBasePath;
 }
 
+//! @brief Get the base path that all icon paths are relative
+QFileInfo ModelObjectAppearance::getXMLFile() const
```
peterNordin: Wrong comment

HopsanGUI/GUIObjects/GUIModelObjectAppearance.cpp
```diff
@@ -1204,6 +1210,12 @@ void ModelObjectAppearance::setBasePath(const QString path)
     setRelativePathFromAbsolute(); //Reset relative path to new basepath
 }
 
+//! @brief Access method to manually set the BaseIconPath
+void ModelObjectAppearance::setXMLFile(const QFileInfo file)
+{
+    mXMLFile = QFileInfo(file);
```
peterNordin: Wrong comment, and unnecessary? Qfileinfo copy constructor call

HopsanGUI/ModelHandler.cpp
```diff
@@ -371,6 +371,14 @@ TextEditorWidget *ModelHandler::loadTextFile(QString fileName)
     }
     QFileInfo fileInfo(file);
 
+    //Abort if file is already open
+    for(const auto& editor : mTextEditors) {
+        if(fileInfo == editor->getFileInfo()) {
+            gpMessageHandler->addErrorMessage("File is already open: "+fileInfo.absoluteFilePath());
+            return 0;
```
peterNordin: nullptr

HopsanGUI/Widgets/LibraryWidget.cpp
```diff
@@ -864,6 +870,17 @@ void LibraryWidget::handleItemClick(QTreeWidgetItem *item, int column)
             {
                 QDesktopServices::openUrl(QUrl("file:///" + gpLibraryHandler->getModelObjectAppearancePtr(mItemToTypeNameMap.find(pFirstSubComponentItem).value())->getBasePath()));
             }
+            else if(pReply == pEditXMLAction) {
+                QFileInfo xmlFile = gpLibraryHandler->getModelObjectAppearancePtr(mItemToTypeNameMap.find(pFirstSubComponentItem).value())->getXMLFile();
+                gpModelHandler->loadTextFile(xmlFile.absoluteFilePath());
+            }
+            else if(pReply == pEditCodeAction) {
+                QString typeName = gpLibraryHandler->getModelObjectAppearancePtr(mItemToTypeNameMap.find(pFirstSubComponentItem).value())->getTypeName();
+                auto& appearance = gpLibraryHandler->getModelObjectAppearancePtr(typeName);
```
peterNordin: Reference and assuming basePath has trailing slash (see other comments)

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
robbr48: True. I will remove the &.

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
robbr48: Perhaps all functions returning a path should be responsible of adding a trailing slash if it is missing? I will just add an if statement to solve it for now.

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
robbr48: This variable is only used to check if there is a source file specified, to determine if the button should be visible or not. A bit stupid I guess, but the alternative would be to add a "hasSourceCodeFile()" function, which is not much better.

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
peterNordin: No, this is how it was previously solved, but then sometimes you do not want a trailing slash, I have had lots of problems with this sort of thing. The correct way is to never include a trailing slash, and then use a helpfunction similar to pythons path.join(part1,part2) that is clever and adds / as needed. But we can add that later

PR #1828, opened 2019-02-26
Apply & simulate button in componenet properties

robbr48:
"Apply & simulate" button in component properties dialog. Will apply all settings and simulate the model without closing the dialog.


peterNordin:
Is this even necessary? You can click apply then the simulate button without closing the dialog, or use the keyboard shortcut for simulation. At least that worked before, have not tried lately.

robbr48:
Well we could live without it, but it  would be really helpful.

- Keyboard shortcut does not work when dialog has focus
- Students don’t know the keyboard shortcut anyway
- Clicking the simulate button works, but gets annoying when for example tuning a control parameter and running many simulations

I guess we could also make the shortcut work in dialogs though.

PR #1832, opened 2019-03-12
Improve point plotting in Optimization

robbr48:
- Clear plot window before every simulation (but don't recreate the actual window)
- Best color is red, worst is green, candidates black
- Only update candidates when actual points have not changed

### DIFF
--- a/HopsanGUI/OptimizationHandler.cpp
+++ b/HopsanGUI/OptimizationHandler.cpp
@@ -89,6 +89,13 @@ void OptimizationHandler::startOptimization(ModelWidget *pModel, QString &modelP
 {
     if(mpWorker)
     {
+        if(mPlotPoints) {
+            PlotWindow *pParPlotWindow = gpPlotHandler->getPlotWindow("parplot");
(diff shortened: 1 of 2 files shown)


HopsanGUI/OptimizationHandler.cpp
```diff
@@ -1440,13 +1453,13 @@ void OptimizationMessageHandler::objectiveChanged(size_t idx)
 
 void OptimizationMessageHandler::candidatesChanged()
 {
-    mpHandler->plotPoints();
+    mpHandler->plotPoints(true);
```
peterNordin: In cases like this I would prefer that the parameter (argument) is of a enum class type, so that you can write a resonable value instead of "true". In this case PointPlotOptions::OnlyCandidates or similar.

robbr48: Fixed!

peterNordin:
Did the lastforce-push add content from another PR ? In tis case the component properties Apply and simulate?

PR #1833, opened 2019-03-12
Launch batch/shell scripts from HCOM

robbr48:
Launch .bat and .sh files using ``exec`` function in HCOM.

Syntax: 
``exec script.bat arg1 arg2 ...``

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -2390,8 +2390,10 @@ void HcomHandler::executeRunScriptCommand(const QString cmd)
             }
         }
         HCOMINFO("Launching "+file.fileName()+"...");
-        QDir::setCurrent(QFileInfo(path).absolutePath());
+        QDir prevDir = QDir::current();
+        QDir::setCurrent(mPwd);
         system(cmd.toStdString().c_str());
+        QDir::setCurrent(prevDir.absolutePath());
         HCOMINFO(file.fileName()+" finished.");
         return;
     }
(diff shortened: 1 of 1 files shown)


HopsanGUI/HcomHandler.cpp
```diff
@@ -2370,6 +2370,31 @@ void HcomHandler::executeRunScriptCommand(const QString cmd)
...
     path = dir+path.right(path.size()-path.lastIndexOf("/"));
     QFile file(path);
+#ifdef _WIN32
+    if(file.exists() && path.endsWith(".bat"))
+    {
+        QString cmd = "CMD.exe /C "+path;
+#else //Mac or Linux
+    if(file.exists() && path.endsWith(".sh"))
+    {
+        QString cmd = "sh "+path;
```
peterNordin: I would like the possibility to choose what shell to use, I write all my scripts in bash so sh might not for in all cases.
Not sure how, maybe an optional flag that can be passed to the exec command.

Also scripts may not be named .sh, and what if someone would like to run a python script, or powershell, or something else. Obviously this is a bigger task,maybe for later.

You may also want to run bash scripts in Windows. So assuming maybe we should not assume CMD.exe if WIN32.
Maybe we could have a flag where you can choose  "script interpreter" but default to CMD.exe on WIN32 if it is not given. For Linux sh could be the default, or maybe check a ENV variable "SHELL" i think (not sure) which is the selected default shell. One could also check for common file endings, so that .py scripts would launch with python.
Just some ideas

robbr48: I agree. There should be flags for specifying interpreter (-bat, -sh, -bash etc). If no flag is specified, it should guess based on file ending and platform.

PR #1834, opened 2019-03-12
WIP: Possibility to evaluate external functions in optimization scripts

robbr48:
Evaluate external functions in optimization scripts, for example from .bat scripts.

Remaining work: Parallel evaluation of all points somehow.


peterNordin (commented):
Looks good so far, but I assume some work remains

robbr48:
Yes, especially parallel evaluation and maybe example scripts. I was actually able to optimize OpenModelica models from Hopsan already, but only with sequential algorithms.

robbr48:
I close this for now due to inactivity.

Issue #1829, opened 2019-03-11
Support import of column-wise CSV with headings


robbr48:
**Is your improvement request related to a problem? Please describe.**
The vast majority of CSV files from other simulation tools are column-wise and has variable names in first row. This format is not supported by Hopsan CSV import.

**Describe the solution you'd like**
Try to figure out the CSV format of the file before calling CSV parser.

robbr48:
To clarify: Column-wise CSVs with variable names can be loaded without error, but variable names are not parsed correctly. Variables are called "CSV1", "CSV2" etc.

robbr48:
Correction: Column-wise CSVs with variable names are not loaded, without any error message.

peterNordin:
Could you give an example of the header from the file in this case.

It is problematic since there is no standard for this, the file header could be one or multiple rows, use comment character or not, be comma separated or not.

But I guess we should add some more optional parameters, example:
1. A list of lines to ignore on the form:  1,2,4-6
2. The line containing column-names
3. Are the names character separated, and if so what character (could assume , or ;) or just fall back to space separation if such chars are not available
4. Also whether to ignore a comment char on the name line (maybe do this automatically)

Example file:
# first line (some meta info)
# var1, var2, var3  (variable names)
# --------------------------
1, 2, 3
2, 3, 4
.
.
.

robbr48:
I am trying to import a CSV file from OpenModelica with the following contentes (only first 7 lines):

"time","x1","x2","x3","x4","x5","y"
0,2,1,1,1,1,901
0.002,2,1,1,1,1,901
0.004,2,1,1,1,1,901
0.006,2,1,1,1,1,901
0.008,2,1,1,1,1,901
0.01,2,1,1,1,1,901

robbr48:
I have an idea for how to handle CSV files with unknown contents. Here is some pseudocode:

bool columnwise;
int linestoskip;
if(first element in file is numeric) {
  linestoskip = 0;
  columnwise = true; //Most common, so assume it
}
else if(first element is first row is non-numeric and last element of first row is non-numeric) {
  columnwise = true;
  linestoskip = index of first numeric element in first column;
} 
else {
  columnwise = false;
  linestoskip = index of first numeric element in first row;
}

if(!columnwise && linestoskip == 3) {
  assume Hopsan-style rowwise CSV
}

robbr48:
Then I also think we should add some flags to the "repl" command, like this:

[...]

PR #1835, opened 2019-03-12
Improved csv support

### LINKED ISSUE #1830: Importing plot data fails if no model is open
**Describe the bug**
Importing plot data with no model open causes crash

**To Reproduce**
*If the bug can be reproduce, please give the steps taken to reproduce the behavior:*
1. Start Hopsan
2. Do not open a model
3. Call "repl" command from HCOM
4. The program will crash

**Expected behavior**
Either an error message, or as import orphan variables.

robbr48:
Related to #1829.
Resolves #1830.

- [X] Check for column-wise/row-wise files
- [X] Count number of lines to skip
- [X] Attempt to detect Hopsan-style CSV
- [x] Implement load function for row-wise generic CSVs

### DIFF
--- a/HopsanCore/include/ComponentUtilities/CSVParser.h
+++ b/HopsanCore/include/ComponentUtilities/CSVParser.h
@@ -50,7 +50,7 @@ namespace hopsan {
 class HOPSANCORE_DLLAPI CSVParserNG
 {
 public:
-    CSVParserNG(const char separator_char = ',');
+    CSVParserNG(const char separator_char = ',', size_t linesToSkip=0);
     ~CSVParserNG();
 
     bool openFile(const HString &rFilepath);
--- a/HopsanGUI/LogDataHandler2.cpp
+++ b/HopsanGUI/LogDataHandler2.cpp
@@ -756,11 +756,23 @@ void LogDataHandler2::importFromCSV_AutoFormat(QString importFilePath)
     if (file.open(QFile::ReadOnly))
     {
         QTextStream ts(&file);
-        QStringList firstRow = ts.readLine().split(',');
+        QString firstLine = ts.readLine();
+        if(firstLine.isEmpty()) {
+            gpMessageHandler->addErrorMessage("CSV file is empty: "+importFilePath);
+            return;
+        }
+        QStringList firstRow = firstLine.split(',');
+
         QStringList firstColumn;
         firstColumn.push_back(firstRow.first());
         while(!ts.atEnd()) {
-            firstColumn.push_back(ts.readLine().split(',').first());
+            QStringList row = ts.readLine().split(',');
+            if(row.isEmpty()) {
+                firstColumn.push_back(QString());
+            }
+            else {
+                firstColumn.push_back(row.first());
+            }
         }
         file.close();
         if (!firstRow.isEmpty())
--- a/HopsanGUI/LogDataHandler2.cpp
+++ b/HopsanGUI/LogDataHandler2.cpp
@@ -804,7 +804,7 @@ void LogDataHandler2::importFromCSV_AutoFormat(QString importFilePath)
             }
             else if (!columnWise && !hopsanCSV)
             {
-                //! @todo Implement
+                importFromPlainRowCsv(importFilePath, ',', linesToSkip);
             }
             else {
                 importHopsanRowCSV(importFilePath);
@@ -918,11 +918,10 @@ void LogDataHandler2::importHopsanRowCSV(QString importFilePath)
 }
 
 
-void LogDataHandler2::importFromPlainColumnCsv(QString importFilePath, const QChar separator, const int linesToSkip, const int timecolumn)
+void LogDataHandler2::importFromPlainColumnCsv(QString importFilePath, const QChar separator, const int rowsToSkip, const int timecolumn)
 {
     if(importFilePath.isEmpty())
     {
-
         importFilePath = QFileDialog::getOpenFileName(0,tr("Choose .csv File"),
                                                        gpConfig->getStringSetting(CFG_PLOTDATADIR),
                                                        tr("Comma-separated values files (*.csv)"));
@@ -939,7 +938,7 @@ void LogDataHandler2::importFromPlainColumnCsv(QString importFilePath, const QCh
     QTextStream ts(&file);
     QStringList names;
     file.open(QFile::ReadOnly);
-    if(linesToSkip > 0) {
+    if(rowsToSkip > 0) {
         names = ts.readLine().split(separator);
     }
     for(auto& name : names) {
@@ -947,7 +946,7 @@ void LogDataHandler2::importFromPlainColumnCsv(QString importFilePath, const QCh
     }
     file.close();
 
-    CoreCSVParserAccess csvparser(importFilePath,separator,linesToSkip);
+    CoreCSVParserAccess csvparser(importFilePath,separator,rowsToSkip);
 
     if(!csvparser.isOk())
     {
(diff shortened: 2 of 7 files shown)


HopsanCore/include/ComponentUtilities/CSVParser.h
```diff
@@ -50,7 +50,7 @@ namespace hopsan {
 class HOPSANCORE_DLLAPI CSVParserNG
 {
 public:
-    CSVParserNG(const char separator_char = ',');
+    CSVParserNG(const char separator_char = ',', size_t linesToSkip=0);
```
peterNordin: Maybe a parameter should be added to the CsvParser components (lookup table components) to allow the user to select the number of lines to skip.
There is also support for "comment char" typically # to autoskip any header lines (lines at the top only) starting with that char, maybe add support for that as well....

HopsanGUI/LogDataHandler2.cpp
```diff
@@ -899,7 +924,18 @@ void LogDataHandler2::importFromPlainColumnCsv(QString importFilePath, const QCh
     QFileInfo fileInfo(file);
     gpConfig->setStringSetting(CFG_PLOTDATADIR, fileInfo.absolutePath());
 
-    CoreCSVParserAccess csvparser(importFilePath,separator);
+    QTextStream ts(&file);
+    QStringList names;
+    file.open(QFile::ReadOnly);
+    if(linesToSkip > 0) {
```
peterNordin: Assumes names on first line, but OK for now I guess, at least better then before

HopsanGUI/LogDataHandler2.cpp
```diff
@@ -927,7 +963,12 @@ void LogDataHandler2::importFromPlainColumnCsv(QString importFilePath, const QCh
         for (int i=1; i<data.size(); ++i)
         {
             SharedVariableDescriptionT pVarDesc = SharedVariableDescriptionT(new VariableDescription);
-            pVarDesc->mDataName = "CSV"+QString::number(i);
+            if(!names.isEmpty()) {
+                pVarDesc->mDataName = names[i];
```
peterNordin: Assumes number of names same as number of columns, wiĺl crash if (out of bounds)  if this is not the case, (bad csv file)

if (names.size() > i) would likely fix this problem

HopsanCore/include/ComponentUtilities/CSVParser.h
robbr48: I agree, but that belongs to a separate PR. Lookup tables component for 2D and 3D do not even allow choosing separator character yet.

HopsanGUI/LogDataHandler2.cpp
```diff
@@ -756,20 +756,45 @@ void LogDataHandler2::importFromCSV_AutoFormat(QString importFilePath)
     if (file.open(QFile::ReadOnly))
     {
         QTextStream ts(&file);
-        QStringList fields = ts.readLine().split(',');
-        if (!fields.isEmpty())
+        QStringList firstRow = ts.readLine().split(',');
+        QStringList firstColumn;
+        firstColumn.push_back(firstRow.first());
```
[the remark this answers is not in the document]
robbr48: That's true. I assumed the file was not empty. I will add a check.

HopsanGUI/LogDataHandler2.cpp
robbr48: Yes, I guess 90+% of all CSV files will have names on first line, and we have to assume something. Only problem would be if same name is used multiple times.

HopsanGUI/LogDataHandler2.cpp
robbr48: True, I will add a fix.

HopsanGUI/LogDataHandler2.cpp
```diff
@@ -756,20 +756,45 @@ void LogDataHandler2::importFromCSV_AutoFormat(QString importFilePath)
...
+        while(!ts.atEnd()) {
+            firstColumn.push_back(ts.readLine().split(',').first());
```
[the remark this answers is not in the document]
robbr48: Actually, calling split(',') for an empty string will return a QStringList with one (empty) element, so it should not be a problem. I will add a check anyway though, just to be on the safe side.

HopsanCore/include/ComponentUtilities/CSVParser.h
peterNordin: hmm I made those I think, why did I not add that, cant remember.... Yes we do it in a different PR
I can look it over once I have my new updated indexingcsvparser ready

HopsanGUI/LogDataHandler2.cpp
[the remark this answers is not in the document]
peterNordin: aha good to know

HopsanGUI/LogDataHandler2.cpp
peterNordin: yes, but hopsan could "increment" the name in that case, but the question is how much error checking is really necessary, at some point we must assume valid input files. But this could be a nice thing to check, if someone is creating these files by hand it is easy to copy paste something and a check lite this could safe that person some time

HopsanGUI/LogDataHandler2.cpp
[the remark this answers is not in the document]
peterNordin: Maybe the error message should be adjusted, unlikely but if someone has a file with the first line blank, the message wont make sense. The best thing would be to read firstLine until its not empty.... but maybe it can wait since I am improving the csv parser library and intend to use it in the GUI also. I dont know I can handled this special case there either.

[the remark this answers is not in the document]
robbr48: I will change the error message to something like "First line is empty". Of course we could skip the first lines if they are empty, but that's not really a necessary feature. Like you write below, we must assume that the files are valid somehow. We can simply decide that Hopsan will not support CSV files that starts with a blank line.

HopsanGUI/LogDataHandler2.cpp
robbr48: That's true. I will add a check for unique names.

HopsanGUI/LogDataHandler2.cpp
```diff
@@ -996,6 +995,101 @@ void LogDataHandler2::importFromPlainColumnCsv(QString importFilePath, const QCh
...
+    {
+        QStringList allNames, allAlias, allUnits;
+        QList< QVector<double> > allDatas;
+        int nDataValueElements = -1;
+        QTextStream contentStream(&file);
+        while (!contentStream.atEnd())
+        {
+            //! @todo this is not the most clever and speedy implementation
+            QStringList lineFields = contentStream.readLine().split(separator);
+            if (lineFields.size() > linesToSkip)
```
peterNordin: This looks odd, linesFields is columns I belive, and it is compared with a variable linesToSkip, Should it be columnsToSkip?

HopsanGUI/LogDataHandler2.cpp
```diff
@@ -996,6 +995,101 @@ void LogDataHandler2::importFromPlainColumnCsv(QString importFilePath, const QCh
...
+        }
+
+        if (parseOk)
+        {
+            ++mCurrentGenerationNumber;
+
+            // Figure out time
+            SharedVectorVariableT pTimeVec;
+            //! @todo what if multiple subsystems with different time
+            pTimeVec = insertTimeVectorVariable(allDatas[timeRow], fileInfo.absoluteFilePath());
```
peterNordin: What if there is no time row in the data, (timeRow = -1) might be used to indicate this. Then a time vector should not be created

peterNordin: I noticed that this may also be a problem in the existing column reader code, not sure if it needs fixing right now, probably not

HopsanGUI/LogDataHandler2.cpp
robbr48: I renamed linesToSkip to rowsToSkip/columnsToSkip in both improt functions to make it more clear.

HopsanGUI/LogDataHandler2.cpp
robbr48: I fixed it for the row-wise function only. We can add the other fix later if needed.

peterNordin:
Remember to change the PR name (remove WIP) before merging, also I think this one resolves 1830 ? In that case add that to the description before merging

PR #1836, opened 2019-03-15
Optimization table output

robbr48:
Use tables instead of widgets in layouts for optimization dialog
- TableView for parameter output
- TableWidget for progress bars

### DIFF
--- a/HopsanGUI/Dialogs/OptimizationDialog.cpp
+++ b/HopsanGUI/Dialogs/OptimizationDialog.cpp
@@ -40,6 +40,10 @@
 #include <QTimer>
 #include <QScrollArea>
 #include <QToolBar>
+#include <QTableView>
+#include <QStandardItemModel>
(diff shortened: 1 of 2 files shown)


HopsanGUI/Dialogs/OptimizationDialog.cpp
```diff
@@ -495,102 +449,83 @@ void OptimizationDialog::updateCoreProgressBars()
...
-        }
-        break;
-    default:
-        break;
+        mpCoreProgressBarsTableWidget->setItem(r,0,new QTableWidgetItem("model"+QString::number(r)));
+        mpCoreProgressBarsTableWidget->setCellWidget(r,1,mCoreProgressBarPtrs.last());
     }
 
+    // Total progress bar
+    delete mpTotalProgressBar;
```
peterNordin: deletelater maybe ? not sure if it matters in this case

PR #1848, opened 2019-05-28
Export model to black-box executable

robbr48:
Export a model to a pre-compiled stand-alone executable.

Remaining work:
- [x] Test on Windows
- [x] Support external libraries
- [x] Help flag
- [x] Support parameter files
- [x] Support configuration files
- [x] Display parameters flag


robbr48:
Some code is duplicated from CLI utilities, although with minor changes. I don't think this can be avoided though. I don't want to simply hijack files from HopsanCLI and copy them to the export directory.

HopsanGUI/GUIObjects/GUISystem.cpp
```diff
@@ -2079,7 +2079,87 @@ void SystemContainer::exportToSimulink()
...
+    }
+    else
+    {
+        garch = HopsanGeneratorGUI::TargetArchitectureT::x86;
+    }
+    auto pCoreSystem = mpCoreSystemAccess->getCoreSystemPtr();
+    QStringList externalLibraries;
+    //! @todo an idea here is to always treat the default library as external, and export it as such (and never build it in by default), that would reduce special handling of the default library
+    //! @todo This code prevents nesting an external fmu inside an export, not sure if we need to support this
+    spGenerator->setAutoCloseWidgetsOnSuccess(true);
```
peterNordin: no need to set it to true here if it is false, below. Or what is the intended behavior ?

HopsanGUI/MainWindow.cpp
```diff
@@ -804,6 +804,14 @@ void MainWindow::createActions()
     mHelpPopupTextMap.insert(mpExportToSimulinkAction, "Export model to Simulink S-function.");
     connect(mpExportToSimulinkAction, SIGNAL(hovered()), this, SLOT(showToolBarHelpPopup()));
 
+    mpExportToExe_32Action = new QAction(QIcon(QString(ICONPATH) + "Hopsan-ExportExe.png"), tr("Export to Executable Model (32-bit)"), this);
```
peterNordin: I would prefer if we start using the icons in .svg format

HopsanGUI/LibraryHandler.cpp
```diff
@@ -535,6 +536,14 @@ bool LibraryHandler::loadLibrary(SharedComponentLibraryPtrT pLibrary, LibraryTyp
                         sourceElement = sourceElement.nextSiblingElement(QString(XML_LIBRARY_SOURCE));
                     }
 
+                    // Read extra source files
+                    QDomElement extraSourceElement = xmlRoot.firstChildElement(QString(XML_LIBRARY_EXTRA_SOURCE));
```
peterNordin: Is extra sources really needed, could just append files to sources

HopsanGenerator/include/generators/HopsanExeGenerator.h
```diff
@@ -0,0 +1,47 @@
...
+ For details about the 'Hopsan Group' or information about Authors and
+ Contributors see the HOPSANGROUP and AUTHORS files that are located in
+ the Hopsan source code root directory.
+
+-----------------------------------------------------------------------------*/
+
+#ifndef HOPSANEXEGENERATOR_H
+#define HOPSANEXEGENERATOR_H
+
+//#include <QMap>
```
peterNordin: junk comment

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -0,0 +1,375 @@
...
+    //------------------------------------------------------------------//
+
+    QDir resourceDir(buildPath+"/resources/");
+    std::list<hopsan::HString> assets = pSystem->getModelAssets();
+    QMap<QString, QString> assetsMap;
+    if (!assets.empty()) {
+        printMessage("Exporting model assets");
+    }
+    for (const auto& asset : assets) {
+        QFileInfo assetInfo(asset.c_str());
```
peterNordin: Is this block of code a copy from the fmu export code or is it specific to the exe? If copy, then a common utility function should be made.

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -0,0 +1,375 @@
...
+        line.chop(1);
+        for (auto it = replaceMap.begin(); it != replaceMap.end(); ++it) {
+            line.replace(it.key(), it.value());
+        }
+        line.replace(R"(")", R"(\")");
+        modelLines.append(line);
+    }
+    modelLines.last().append("\\n");
+    modelFile.close();
+
```
peterNordin: I have a feeling that this code is also reused, but if so we could fix that later (use a common utility function) if possible.

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -0,0 +1,375 @@
...
+    QFile compileCppBatchFile;
+    compileCppBatchFile.setFileName(buildPath + "/compileCpp.sh");
+    if(!compileCppBatchFile.open(QIODevice::WriteOnly | QIODevice::Text))
+    {
+        printErrorMessage("Failed to open compileCpp.sh for writing.");
+        return false;
+    }
+    //Write the compilation script file
+    QTextStream compileCppBatchStream(&compileCppBatchFile);
+    compileCppBatchStream << mCompilerSelection.path+"g++ -pipe -std=c++11 -fPIC -c -DHOPSAN_INTERNALDEFAULTCOMPONENTS -DHOPSAN_INTERNAL_EXTRACOMPONENTS " << "exe_main.cpp exe_utilities.cpp " << mExtraSourceFiles.join(" ");
```
peterNordin: Since everything is compiled into one exe, -fPIC should not be needed, i think

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -0,0 +1,375 @@
...
+    QFile linkBatchFile;
+    linkBatchFile.setFileName(buildPath + "/link.sh");
+    if(!linkBatchFile.open(QIODevice::WriteOnly | QIODevice::Text))
+    {
+        printErrorMessage("Failed to open link.sh for writing.");
+        return false;
+    }
+    //Write the compilation script file
+    QTextStream linkBatchStream(&linkBatchFile);
+    linkBatchStream << mCompilerSelection.path+"g++ -pthread -fPIC -w -static-libgcc -Wl,--rpath,'$ORIGIN/.'";
```
peterNordin: If everything is linked statically then setting rpath here should not be needed, but I dont know if it should be possible to load external "dll libraries" or not

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
```diff
@@ -1056,7 +1056,7 @@ bool HopsanGeneratorBase::copyDefaultComponentCodeToDir(const QString &path) con
     return true;
 }
 
-bool HopsanGeneratorBase::copyExternalComponentCodeToDir(const QString &destinationPath, const QStringList &externalLibraries) const
+bool HopsanGeneratorBase::copyExternalComponentCodeToDir(const QString &destinationPath, const QStringList &externalLibraries, QStringList *extraSourceFiles) const
```
peterNordin: Any particular reason for using a pointer to QStringList ?

HopsanGenerator/templates/exe_utilities.cpp
```diff
@@ -0,0 +1,488 @@
...
+
+ The full license is available in the file LICENSE.
+ For details about the 'Hopsan Group' or information about Authors and
+ Contributors see the HOPSANGROUP and AUTHORS files that are located in
+ the Hopsan source code root directory.
+
+-----------------------------------------------------------------------------*/
+
+//!
+//! @file   ModelUtilities.cpp
```
peterNordin: file name does not match actual name

HopsanGenerator/templates/exe_utilities.cpp
```diff
@@ -0,0 +1,488 @@
...
+        std::cout << "Unknown error!\n";
+        return false;
+    }
+
+    return true;
+}
+
+void printHelpText(ComponentSystem *pSystem)
+{
+    std::cout << "This is a stand-alone executable Hopsan model.\n";
```
peterNordin: Why not use TCLAP like in the CLI application?

HopsanGUI/LibraryHandler.cpp
robbr48: It is needed because we don't want to compile the "master source file" for the library when including the components in exported code. An alternative solution could be to have "mainsource" and "source" instead of "source" and "extrasource", but it's just different names.

HopsanGUI/GUIObjects/GUISystem.cpp
robbr48: I'm not sure really. This code was copied from the exportToFMU() function above. I can remove it from both functions.

HopsanGUI/MainWindow.cpp
robbr48: I agree, but I think we should then change all icons in a separate PR.

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
robbr48: Good point, it is copied so I will make a utility function.

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
robbr48: That is not the intention. I will try with removing rpath. The purpose of this export is to obtain only one single exe file.

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
robbr48: Well the only reason was that FMI and Simulink export did not yet support extra source files. I can add support for them as well.

HopsanGenerator/templates/exe_utilities.cpp
robbr48: I want to minimize external dependencies. It should only support very basic arguments anyway.

HopsanGUI/GUIObjects/GUISystem.cpp
peterNordin: Hmm then I wrote that 😃 not sure myself really

HopsanGUI/MainWindow.cpp
peterNordin: Yes

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
peterNordin: I meant, why use a pointer instead of a reference like it is in most other places

robbr48: Because a reference argument can't have a default value, i.e. nullptr. But never mind, I added support for this in he other generators as well now.

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
robbr48: No, I will fix it now. The `generateModelFile()` member function is actually duplicated in both FMI and exe generator. I will move it to the base class.

HopsanGUI/LibraryHandler.cpp
peterNordin: OK I see what you mean. Does this mean that one has to add extra source files to libraries in order for them to be exportable to .exe?

robbr48: No, we generate a new file called `extra-components.cpp`, where we include component .hpp files and register them in HopsanCore.

peterNordin: Yes but the reason I ask is because I wrote some of this code maybe a year ago or more and I vagely remember solving this problem somehow without adding extra sources to the xml. But it was so long ago now that I have forgotten what I did. I assume that you are partially using the code I wrote then. I will have a look at the code again and see if I can figure out what I am actually asking about here.

HopsanCore/include/HopsanCoreGitVersion.h
```diff
@@ -1,7 +1,7 @@
-// This file has been automatically generated by ./writeGitVersionHeader.sh
+// This file has been automatically generated by /home/robbr48/git/hopsan/HopsanCore/../writeGitVersionHeader.sh
```
peterNordin: Do not commit this file, I know it inconvenient but I don't have a better solution right now

HopsanGenerator/src/GeneratorTypes.cpp
```diff
@@ -240,6 +240,9 @@ bool ComponentLibrary::loadFromXML(QString filepath)
         else if (elementName == "source") {
             mSourceFiles.append(reader.readElementText());
         }
+        else if (elementName == "extrasource") {
```
peterNordin: Here extrasource is loaded from XML, but it does not seem to be saved in the save to XML function
Also the HoLC code may need to be update to set this field properly (or just empty)

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
```diff
@@ -1056,7 +1056,7 @@ bool HopsanGeneratorBase::copyDefaultComponentCodeToDir(const QString &path) con
     return true;
 }
 
-bool HopsanGeneratorBase::copyExternalComponentCodeToDir(const QString &destinationPath, const QStringList &externalLibraries) const
+bool HopsanGeneratorBase::copyExternalComponentCodeToDir(const QString &destinationPath, const QStringList &externalLibraries, QStringList &extraSourceFiles) const
```
peterNordin: is unclear that the extraSourceFiles are meant to be an output variable here, (no documentation block) and I would prefer if it is called rExtraSourceFiles,  (r for reference and return value). (Some code styles would recommend sending by pointer her, but lets not do that since it is not done in other places in Hopsan.)

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
```diff
@@ -1127,6 +1127,11 @@ void hopsan::register_extra_components(hopsan::ComponentFactory* pComponentFacto
             }
 
             QString generatorErrorMessage;
+            QStringList newSourceFiles = lib.mExtraSourceFiles;
```
peterNordin: why make a copy and then just take size of, could loop over range directly, or check size on original ?

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -0,0 +1,292 @@
...
+#endif
+
+    if (!linkingOK) {
+        printErrorMessage("Failed to link exported code for executable model.");
+        return false;
+    }
+
+    QDir outputDir = QFileInfo(outputExecutableFile).absoluteDir();
+    outputDir.cdUp();
+#ifdef WIN32
```
peterNordin: this should be _WIN32, WIN32 without leading _ does not always exist

HopsanGUI/LibraryHandler.cpp
peterNordin: OK, lets see if I got this right.  The reason for the "extra sources" field is to enable building of libraries that consists of multiple cpp files. When the library is exported as a bundle (in an FMU or EXE) the "registration code" is auto-generated, and the "main source file" that is listed in sources is not needed, so it is ignored by the export code. The problem with the previous code was that it did not export any of the additional source files at all in this case. So such libraries would not work.  Your addition of "extra-sources" solves this by listing them under a separate tag in the XML?

robbr48: Yes, that is correct. We have probably not noticed it before, since most libraries have only one source file. 

A related problem is how to handle include paths. Right now they are specified by the `<cflags>` tag, usually with a relative path. Extracting paths from this string and converting them to absolute path if needed is for sure possible, but not a very nice solution. Perhaps we should instead add  `<includepath>` tags?

A third problem is how to handle other compile and link flags. We could just append them to the build flags for the executable, but how can we know that they don't conflict with existing flags or libraries?

HopsanGenerator/src/GeneratorTypes.cpp
robbr48: Done and done. I don't want to add too much new functionality to HoLC since we are going to merge it into HopsanGUI, but now it can at least handle extra source files.

HopsanGUI/LibraryHandler.cpp
peterNordin: Ok, I have had one such library in the past. But since we only imported it was enough to place all source files in sources. I think that since it is now possible to auto generate the registration code at import/export, maybe in the future there is no need for a "main source".
  The include paths were used for aditional headers that were shipped with the library, so relativ path forks find, but if you depend on headers in your system or other installation path, then that is ofcourse not portable between systems.
 Build flags for different compilers can be enabled I think, but some new code is needed. Right now you can set attributes in the XML element for OS and Architecture (I think). This should also support different compilers.

HopsanGenerator/src/GeneratorTypes.cpp
peterNordin: Good, then that field wont be lost when you open a library in holc as it will save the XML file overwriting the old one

HopsanGUI/LibraryHandler.cpp
robbr48: The problem with include paths is that they are relative to the location of the library, not to the build directory of the exported code. If it is specified using build flags we need to extract the code from `-I<path>` and then convert it to an absolute path. If it is instead specified using an <includepath> XML tag we don't need the extraction step, we can just convert it directly. What do you think?

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
robbr48: I added a comment and renamed the argument. Sholdn't all other reference arguments in HopsanGenerator also use the "r" prefix then?

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
robbr48: No idea :) I removed the unnecessary copy.

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
robbr48: Ops, my bad. It is fixed.

HopsanGUI/LibraryHandler.cpp
peterNordin: A yes, before we built the code directly inside the library directory. Thats why it was not a problem. Yes I think that you are right. We should separate includepaths and also linkpaths (or whatever they should be called)  from the cflags and lflags. In this PR or a different one? If you do it in this one, remember to write both save and load, also for holc so that this data is not lost if library loaded there.

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
peterNordin: Yes initially the style was  'type &rParameter' for all references, but I think that we should instead only use that style for references that are also output variables. Then the r can mean both 'return' and 'reference'. Inputs as references should be. 'const type &parameter'. An alternative would be to follow the google styleguide recommendation and allways use pointers for output parameters.

HopsanGUI/LibraryHandler.cpp
robbr48: I added XML tags for include paths, link paths and link libraries in CAF files. The link libraries maybe don't need their own tag, but it can make it easier for users to understand.

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
robbr48: Ok, sounds reasonable. We currently use pointers in some places and references in others. But this discussion does not belong in this PR, so I will keep these references for now.

HopsanGenerator/src/GeneratorUtilities.cpp
```diff
@@ -223,6 +229,9 @@ bool compileComponentLibrary(QString path, HopsanGeneratorBase *pGenerator, QStr
     ch.addLinkLibrary("hopsancore");
 #endif
     ch.addLinkLibrary("c++", {Compiler::Clang});
+    for(QString linkLibPath : cl.mLinkLibraries) {
+        ch.addIncludePath(linkLibPath);
```
peterNordin: Should not be addIncludePath, copy paste error I assume

HopsanGenerator/src/GeneratorUtilities.cpp
```diff
@@ -238,7 +247,7 @@ bool compileComponentLibrary(QString path, HopsanGeneratorBase *pGenerator, QStr
     pGenerator->printMessage("Work Directory: "+libRootDir);
     pGenerator->printMessage("Output file:    "+ch.outputFile());
     pGenerator->printMessage("Source files:   "+ch.sourceFiles().join(" "));
-    pGenerator->printMessage("Compiler flags: "+ch.compilerFlags(compilerSelection.compiler).join(" "));
+    pGenerator->printMessage("Compiler flags: "+ch.compilerFlags(compilerSelection.compiler).join(" ")+" ");
```
peterNordin: Why + ""

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -248,6 +253,12 @@ bool HopsanExeGenerator::compileAndLinkExe(const QString &buildPath, const QStri
         linkBatchStream << " " << objFile;
     }
     linkBatchStream << " -o "+outputExecutableFile+"\n";
+    for(const QString linkPaths : mLinkPaths) {
+        linkBatchStream << " -L" << linkPaths;
```
peterNordin: Need " " around libkpaths if they may contain spaces
Also call each item linkPath (singular)

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -248,6 +253,12 @@ bool HopsanExeGenerator::compileAndLinkExe(const QString &buildPath, const QStri
         linkBatchStream << " " << objFile;
     }
     linkBatchStream << " -o "+outputExecutableFile+"\n";
```
peterNordin: Might  need to quote the outputExecutableFile here  if path contains spaces

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -267,7 +278,14 @@ bool HopsanExeGenerator::compileAndLinkExe(const QString &buildPath, const QStri
     {
         linkBatchStream << " " << objFile;
     }
-    linkBatchStream << " -o "+outputExecutableFile+" -ldl\n";
+    for(const QString linkPaths : mLinkPaths) {
+        linkBatchStream << " -L" << linkPaths;
```
peterNordin: Quotes

HopsanGenerator/src/generators/HopsanExeGenerator.cpp
```diff
@@ -267,7 +278,14 @@ bool HopsanExeGenerator::compileAndLinkExe(const QString &buildPath, const QStri
...
+    }
+    for(const QString linkLibrary : mLinkLibraries) {
+        linkBatchStream << " -l" << linkLibrary;
+    }
+    linkBatchStream << " -ldl";
+    linkBatchStream << " -o "+outputExecutableFile << "\n";
```
peterNordin: Quotes

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -1859,6 +1863,12 @@ bool HopsanFMIGenerator::compileAndLinkFMU(const QString &fmuBuildPath, const QS
     {
         linkBatchStream << " " << objFile;
     }
+    for(const QString linkPaths : mLinkPaths) {
+        linkBatchStream << " -L" << linkPaths;
```
peterNordin: Quotes

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -1879,6 +1889,12 @@ bool HopsanFMIGenerator::compileAndLinkFMU(const QString &fmuBuildPath, const QS
...
```
peterNordin: Quotes

HopsanGenerator/src/generators/HopsanSimulinkGenerator.cpp
```diff
@@ -118,9 +118,11 @@ bool HopsanSimulinkGenerator::generateToSimulink(QString savePath, QString model
 #ifdef _WIN32
     compileScriptStream << " -DWIN32";      //!< @todo Not sure if this one is needed, the correct macro to check for is _WIN32
 #endif
-    Q_FOREACH(const QString &s, getHopsanCoreIncludePaths())
-    {
-        compileScriptStream << QString(" -I%1").arg(s);
+    for(const QString includePath : getHopsanCoreIncludePaths()) {
+        compileScriptStream << QString(" -I%1").arg(includePath);
```
peterNordin: Quotes

HopsanGenerator/src/generators/HopsanSimulinkGenerator.cpp
```diff
@@ -118,9 +118,11 @@ bool HopsanSimulinkGenerator::generateToSimulink(QString savePath, QString model
...
+    }
+    for(const QString includePath : includePaths) {
+        compileScriptStream << QString(" -I%").arg(includePath);
```
peterNordin: %1 and quotes

HopsanGenerator/src/generators/HopsanSimulinkGenerator.cpp
```diff
@@ -133,9 +135,15 @@ bool HopsanSimulinkGenerator::generateToSimulink(QString savePath, QString model
     }
     compileScriptStream  << QString(" %1.cpp ").arg(name);
 #ifndef _WIN32
-    compileScriptStream << "-ldl ";
+    compileScriptStream << "-ldl";
 #endif
-    compileScriptLStream << QString("-output %1").arg(name);
+    for(const QString linkPath : linkPaths) {
+        compileScriptStream << " -L" << linkPath;
```
peterNordin: Quotes

PR #1852, opened 2019-06-18
Issue 1851 compile with spaces in paths

### LINKED ISSUE #1851: Unable to recompile libraries when Hopsan installation path contains spaces
**Describe the bug**
HopsanGenerator does not add quotation marks around include and link path when recompiling libraries. If Hopsan is installed in "c:\program files" or similar, compilation fails.

**To Reproduce**
I have not tested myself, but it is reported by multiple users.

**Expected behavior**
Quotation marks should always be added around link and include paths.

Windows systems, Hopsan 2.11.0, official release.

**Additional information**
A workaround is to install Hopsan in a location without paths.

peterNordin:
Resolves #1851, at least for library import and fmu import/export

### DIFF
--- a/HopsanGenerator/src/GeneratorUtilities.cpp
+++ b/HopsanGenerator/src/GeneratorUtilities.cpp
@@ -810,7 +810,7 @@ void CompilerHandler::addBuildFlags(const QVector<BuildFlags> &flags)
 
 void CompilerHandler::addIncludePath(QString ipath, const Compilers compilers)
 {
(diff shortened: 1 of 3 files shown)


peterNordin:
I also need to test this by building a release and installing it on windows.

peterNordin:
I have tested with a 'program files' installation and it seems to work better now
But there are still issues if you name a library or FMU with a space in the actual file name.

PR #1853, opened 2019-06-27
Add or remove components to libraries from HopsanGUI

robbr48:
Changes/features:
- Library version 0.3: Only loads components with CAF files specified in library XML
- Libraries always show up in GUI even if components failed to load in HopsanCore
- Non-loaded components have red font and cannot be added to models
- Create new libraries
- Add new components using wizard dialog
- Add existing components (CAF files must be inside library folder)
- Remove component to libraries, with option to also delete files
- Removed all dual-view code (deprecated)
- Minor fixes to library widget behaviour


robbr48:
@peterNordin Do you have time to review this? Especially the new library version could be important to discuss.

peterNordin:
@robbr48 yes I will have a look over the weekend

HopsanGUI/LibraryHandler.cpp
```diff
@@ -728,46 +1385,42 @@ bool LibraryHandler::loadLibrary(SharedComponentLibraryPtrT pLibrary, LibraryTyp
...
+                        if(!existsInCore) {
                             gpMessageHandler->addWarningMessage("Failed to load component of type: "+pAppearanceData->getFullTypeName(), "failedtoloadcomp");
-                            continue;
                         }
                     }
 
                     // Success, add component to library
-                    if (success)
-                    {
+                    if (true) {
```
peterNordin: could remove this if statement since true

HopsanGUI/LibraryHandler.h
```diff
@@ -54,6 +58,7 @@ class ComponentLibrary
 {
 public:
     QString id;
+    QString version;
```
peterNordin: later we should try to reuse the library and component types from the generator library, but some linking changes are needed

HopsanGUI/LibraryHandler.h
```diff
@@ -84,6 +113,46 @@ class ComponentLibraryEntry
     SharedComponentLibraryPtrT pLibrary;        //!< What library the object belongs to
     HiddenVisibleEnumT visibility=Hidden;       //!< Visible or hidden in library widget
     QStringList displayPath;                    //!< The display path in the library widget
+    EnabledDisabledEnumT disabled=Enabled;      //!< Is component enabled or disabled?
```
peterNordin: This variable name is confusing

HopsanGUI/Widgets/LibraryWidget.cpp
```diff
@@ -650,231 +515,199 @@ void LibraryWidget::handleItemClick(QTreeWidgetItem *item, int column)
...
+            else {
+                path = gpLibraryHandler->getModelObjectAppearancePtr(mItemToTypeNameMap.find(pFirstSubComponentItem).value())->getBasePath();
             }
-            else if(pReply == pOpenFolderAction)
-            {
-                QDesktopServices::openUrl(QUrl("file:///" + gpLibraryHandler->getModelObjectAppearancePtr(mItemToTypeNameMap.find(pFirstSubComponentItem).value())->getBasePath()));
+            QDesktopServices::openUrl(QUrl("file:///" + path));
+        }
+        else if(pReply == pEditXMLAction) {
+            if(!isComponentItem(item)) {
```
peterNordin: I would like to have a isComponentLibraryItem help function here, for clarity

HopsanGUI/Widgets/LibraryWidget.cpp
```diff
@@ -650,231 +515,199 @@ void LibraryWidget::handleItemClick(QTreeWidgetItem *item, int column)
...
+            else {
                 QFileInfo xmlFile = gpLibraryHandler->getModelObjectAppearancePtr(mItemToTypeNameMap.find(pFirstSubComponentItem).value())->getXMLFile();
                 gpModelHandler->loadTextFile(xmlFile.absoluteFilePath());
             }
-            else if(pReply == pEditCodeAction) {
+        }
+        else if(pReply == pEditCodeAction) {
+            if(!isComponentItem(item)) {
+                //Edit library source files
+                SharedComponentLibraryPtrT pLibrary = mItemToLibraryMap[item];
```
peterNordin: not sure if this one should be editable (maybe read-only), since it should be auto generated

HopsanGenerator/include/hopsangenerator.h
```diff
@@ -28,6 +28,12 @@ extern "C" {
...
...DLLAPI bool callAddComponentToLibrary(const char* libraryXMLPath, const char* typeName, const char* displayName, const char* cqsType, const char * const constantNames[], const int numConstantNames, const char * const constantDisplayNames[], const int numConstantDisplayNames, const char * const constantUnits[], const int numConstantUnits, const char * const constantInits[], const int numConstantInits, const char * const inputNames[], const int numInputNames, const char * const inputDescriptions[], const int numInputDescriptions, const char * const inputUnits[], const int numInputUnits, const char * const inputInits[], const int numInputInits, const char * const outputNames[], const int numOutputNames, const char * const outputDescriptions[], const int numOutputDescriptions, const char * const outputUnits[], const int numOutputUnits, const char * const outputInits[], const int numOutputInits, const char * const portNames[], const int numPortNames, const char * const portDescriptions[], const int numPortDescriptions, const char * const portTypes[], const int numPortTypes, const int portsRequired[], const int numPortsRequired, messagehandler_t messageHandler=0, void* pMessageObject=0);
```
peterNordin: nice function :) actually we should invert this relationship so that the GUI uses the Qt api directly instead of going through the c-api. The C-api is only required when used from HopsanCLI. Then we could reduce the C-api to only include functions for those things needed by the CLI. I will do that at some point. (no change needed now)

HopsanGenerator/src/HopsanGeneratorLib.cpp
```diff
@@ -336,3 +337,248 @@ bool callExeExportGenerator(const char* outputPath, void* pHopsanSystem, const c
...
+    cafSpec.mSourceCode = QFileInfo(hppPath).fileName();
+    cafSpec.mRecompilable = true;
+    if(!pGenerator->generateCafFile(cafPath, cafSpec)) {
+        pGenerator->printErrorMessage("Failed to generate component appearance file.");
+        return false;
+    };
+
+    //Generate source file for new component
+    ComponentSpecification compSpec;
+    compSpec.typeName = typeName;
```
peterNordin: not sure if needed, but subtype can no be specified here, so only "main" components can be added. Later typename should be capable of containing full type with some separator. Ex: "HydraulicOrific:MySpecialOrifice", but for now subtype is needed. But maybe not int this case because you can not add such component via dialog anyway I assume.

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
```diff
@@ -930,6 +932,73 @@ bool HopsanGeneratorBase::generateCafFile(QString &rPath, ComponentAppearanceSpe
...
+
+    QTextStream out(&hppFile);
+    out << generateSourceCodefromComponentSpec(comp);
+
+    hppFile.close();
+    return true;
+}
+
+bool HopsanGeneratorBase::generateLibrarySourceFile(const ComponentLibrary &lib)
+{
```
peterNordin: I think that this function duplicates some existing functionality inside the ComponentLibrary utility class. Check the "generateRegistrationCode" function

hopsangeneratorgui/include/hopsangeneratorgui/hopsangeneratorgui.h
```diff
@@ -49,6 +49,16 @@ class HopsanGeneratorGUI : public QWidget
 
     bool checkComponentLibrary(const QString& libraryXMLPath);
 
+    bool addComponentToLibrary(const QString &libraryXmlPath, const QString &cafPath);
+
+    bool addComponentToLibrary(const QString &libraryXmlPath, const QString &typeName, const QString &displayName, const QString &cqsType,
```
peterNordin: We should use a class with all of these as members instead. But right now we have one version in the generator and another in the GUI. So we can wait until they are unified.

hopsangeneratorgui/src/hopsangeneratorgui.cpp
```diff
@@ -525,6 +530,91 @@ bool HopsanGeneratorGUI::checkComponentLibrary(const QString& libraryXMLPath)
...
+                                               const QStringList &portNames, const QStringList &portDescriptions, const QStringList &portTypes, const QList<bool> &portsRequired)
+{
+    auto lw = mPrivates->createNewWidget();
+    loadGeneratorLibrary();
+
+    constexpr auto functionName = "callAddComponentToLibrary";
+    const auto libpath = libraryXmlPath.toStdString();
+    MessageForwarder forwarder(lw->widget());
+    const auto xmlpath = libraryXmlPath.toStdString();
+    const auto typenamestr = typeName.toStdString();
```
peterNordin: subtype is missing, but I am not sure if it is needed in this case anyway, when you have written a new component. Also is should be merged into typename.... later

peterNordin:
I did some tests:

When loading the ExtensionLibrary I get the following warning (it conflicts with it self)
A component with type name 'Subsystem|Initial_Step' was already registered by library 'HopsanExtensionLibrary'. Ignoring version in library 'HopsanExtensionLibrary'.

Choosing the following for a library opens the dir/file for the first subcomponent instead of the library:
Open Containing Folder
Edit component appearance
Edit source code

I am unable to find how to create a new library or to insert or remove components, those context menu options never appear for me.
I tried when loading the ExtensionLibrary

It is possible to chose  on sub-directory items:
Edit component appearance
Edit source code

The red fading in the color of the terminal / messages does not work for me when a componetn library fails to compile

When a ComponentLibvrary fails to compile it is not added to the list, I think that according to the new code, the should be added anyway
I tried introducing an error into the ExampleComponetnLibrary

robbr48:
@peterNordin 

> When loading the ExtensionLibrary I get the following warning (it conflicts with it self)
> A component with type name 'Subsystem|Initial_Step' was already registered by library 'HopsanExtensionLibrary'. Ignoring version in library 'HopsanExtensionLibrary'.

I will investigate. As you noticed above, I did not yet consider subtypes. This is probably related.

> Choosing the following for a library opens the dir/file for the first subcomponent instead of the library:
> Open Containing Folder
> Edit component appearance
> Edit source code

Are you sure? For me the correct XML and source files are opened. The containing folder is wrong for the top-level library folder though. 
 
> I am unable to find how to create a new library or to insert or remove components, those context menu options never appear for me.
> I tried when loading the ExtensionLibrary
>
> It is possible to chose on sub-directory items:
> Edit component appearance
> Edit source code

Strange. When I load the ExtensionLibrary I do get the "Add New Component" and "Add Existing Component" entries if a right-click on a library folder (the ones with green icons).

 
> The red fading in the color of the terminal / messages does not work for me when a componetn library fails to compile

Is this only a cosmetic problem? For me it seems to work.
 
> When a ComponentLibvrary fails to compile it is not added to the list, I think that according to the new code, the should be added anyway
> I tried introducing an error into the ExampleComponetnLibrary

Again, this works for me. I tried introducing errors in both main source file and a component source file. The library is loaded, and all components are red.

HopsanGUI/LibraryHandler.cpp
```diff
@@ -351,6 +374,603 @@ const QVector<SharedComponentLibraryPtrT> LibraryHandler::getLibraries(const QSt
...
+    for(int r=1; r<mpOutputVariablesTable->rowCount(); ++r) {
+        mpOutputVariablesTable->item(r,1)->setBackground(defaultBrush);
+        mpOutputVariablesTable->item(r,4)->setBackground(defaultBrush);
+    }
+    for(int r=1; r<mpPortsTable->rowCount(); ++r) {
+        mpPortsTable->item(r,1)->setBackground(defaultBrush);
+    }
+
+    //Validate type name
+    ComponentSpecification spec = getSpecification();
```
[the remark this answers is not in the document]
robbr48: Yes, there is an `isNameValid()` function, but it does not check that the first character is a letter. We can still use it though, and check the first character separately.

HopsanGUI/LibraryHandler.h
robbr48: I agree. There are more specification classes that are duplicated in HopsanGUI, HopsanCore and HopsanGenerator.

HopsanGUI/LibraryHandler.h
robbr48: It did make sense when it was a Boolean, but I changed to enum to make it less confusing :man_facepalming:

HopsanGenerator/include/hopsangenerator.h
robbr48: Sounds good to me!

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
robbr48: Yes it does, but they have only two lines in common, and the utility function adds annoying comments about which library the component belongs to. These comments makes sense if we include a component in e.g. an FMU, but not in the source file for the actual library.

hopsangeneratorgui/src/hopsangeneratorgui.cpp
robbr48: You can easily work around this by first creating the component and then modifying it's CAF file to add the subtype. We can perhaps add this to the generator later.

robbr48:
@peterNordin The problem with the extenationLibrary is actually a problem with the library itself. It contains two CAF files with the same typename/subtypename. So the error is correct.

peterNordin:
Hmm OK I will look in that. Also I will try a clean rebuild and see if I can get stuff working

peterNordin:
@robbr48 OK I did a clean rebuild and now all the options appear and fading color works. 

However, I ran into some new problems. When using the "example component library" and adding a new component under the sub-dir hydraulics, the generated library would not compile. The new component hpp include was added without the relative hydarulics/file.hpp path. 

Also the generated library includes debug output (I assume) to std:err that library is being registered. Also no indentation on the actual registration lines (not such a big deal, but anyway...)

Also I got the add new component dialog to crash the program, just adding some random constants, not sure what I did. I remember only filling in the name description and unit (should use quantity dropdown boxes) then clicking something ....

peterNordin:
I did not fill in typename and such for the component, just add new constant parameter with name and description, and click "done"

peterNordin:
The _savestate model shows up in the recent models list

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
peterNordin: The purpose of the comments was to make it easier for anyone who wants to edit the library manually after it has been created. But sure the number of comment can likely be reduced.
We can merge those two later, But I prefer to have one generator instead of two for the same thing.

HopsanGUI/LibraryHandler.cpp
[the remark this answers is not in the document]
peterNordin: Maybe create a isTypenNameValid utility, by the way, why must the first char be a letter?
But I agre it should.

robbr48:
> The _savestate model shows up in the recent models list

Fixed.

> However, I ran into some new problems. When using the "example component library" and adding a new component under the sub-dir hydraulics, the generated library would not compile. The new component hpp include was added without the relative hydarulics/file.hpp path.

Fixed.

> Also the generated library includes debug output (I assume) to std:err that library is being registered. Also no indentation on the actual registration lines (not such a big deal, but anyway...)

I removed the output, fixed the indentation, and also corrected library name (was hardcoded).

> Also I got the add new component dialog to crash the program, just adding some random constants, not sure what I did. I remember only filling in the name description and unit (should use quantity dropdown boxes) then clicking something ....
> I did not fill in typename and such for the component, just add new constant parameter with name and description, and click "done"

Fixed. It did not check for empty type name before checking first character.

HopsanGUI/Widgets/LibraryWidget.cpp
robbr48: How do you mean? Just rename the `isComponentItem`function?

HopsanGUI/Widgets/LibraryWidget.cpp
robbr48: I will make it const then for clarity.

HopsanGUI/LibraryHandler.cpp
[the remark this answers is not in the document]
robbr48: The component name is used as class name, and according to C++ standard, class names must start with either a letter or an underscore. I decided to only allow letters in this cases, since underscore makes no sense at the beginning of a type name.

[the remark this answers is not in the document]
peterNordin: OK makes sence

HopsanGUI/Widgets/LibraryWidget.cpp
peterNordin: No, in this case it checks if the item is not a component then it must be a library. It would be easier to read if the code said 'if isLibrary'.
But I realised now that this also works on sub directories. So maybe not an.issue after all. But its a bit strange to be able to edit XML for a sub-directory, but not such a big deal

HopsanGUI/Widgets/LibraryWidget.cpp
peterNordin: I meant, in HoLC I think it showed a meaasage in the editor that you should not edit the file since it will be overwriten everytime you change anything

robbr48: That's true. We don't have the read-only feature in HopsanGUI text editor yet. You may also want to edit this file, for example for temporarily commenting out a component that makes compilation fail. 

I guess we should not overwrite the entire file when we add/remove components, but instead append/remove the specific lines of code. This requires some clever code parsing though.

PR #1864, opened 2019-07-17
Unhardcode HoLC launcher icon for DEB pkg

palob:
Since the icon already is installed to a location compliant with [freedesktop.org standards](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html) you don't need to specify the full icon path or the file type extension in the launcher. 
The icon will be found anyway.

Cf. this [example `.desktop` launcher](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#example) presented in the freedesktop.org documentation or [your own launcher for the Hopsan GUI](https://github.com/Hopsan/hopsan/blob/d6545a0e9c99e7ed8d32cb36bdfbc08576abb89d/buildDebPackage/HopsanGUI.desktop)


palob:
Thank you.

PR #1866, opened 2019-08-22
General improvements to source code editing

robbr48:
Mostly minor changes to improve usability.

### DIFF
--- a/HopsanGUI/Widgets/LibraryWidget.cpp
+++ b/HopsanGUI/Widgets/LibraryWidget.cpp
@@ -99,6 +99,7 @@ LibraryWidget::LibraryWidget(QWidget *parent)
     connect(gpLibraryHandler, SIGNAL(contentsChanged()), this, SLOT(update()));
     connect(mpTree,     SIGNAL(itemPressed(QTreeWidgetItem*,int)),  this,                   SLOT(handleItemClick(QTreeWidgetItem*,int)));
     connect(mpFilterEdit,   SIGNAL(textChanged(QString)), this, SLOT(update()));
+    connect(mpTree, SIGNAL(itemDoubleClicked(QTreeWidgetItem*,int)), this, SLOT(handleItemDoubleClick(QTreeWidgetItem*,int)));
 
     QGridLayout *pLayout = new QGridLayout(this);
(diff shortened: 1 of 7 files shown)


HopsanGUI/Widgets/LibraryWidget.cpp
```diff
@@ -754,6 +755,24 @@ void LibraryWidget::handleItemClick(QTreeWidgetItem *item, int column)
...
 
+void LibraryWidget::handleItemDoubleClick(QTreeWidgetItem *item, int column)
+{
+    if(item != nullptr &&
+        mItemToTypeNameMap.contains(item) &&
+        gpLibraryHandler->getEntry(mItemToTypeNameMap.find(item).value()).displayPath.startsWith(componentlibrary::roots::externalLibraries))
+    {
+        //Edit component source file
+        QString typeName = gpLibraryHandler->getModelObjectAppearancePtr(mItemToTypeNameMap.find(item).value())->getTypeName();
+        auto appearance = gpLibraryHandler->getModelObjectAppearancePtr(typeName);
```
peterNordin: hmm no need to get the appearance ptr twice 
Also rather long chains, that may crash to program is something goes wrong somewhere. Lets say getModelObjectAppearance would return nullptr for some reason. I should not but maybe should always check pointer in non-performance-critical code.

HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -62,6 +62,8 @@ TextEditorWidget::TextEditorWidget(QFileInfo scriptFileInfo, HighlighterTypeEnum
     QFont font("Monospace");
     font.setStyleHint(QFont::TypeWriter);
     mpEditor->setFont(font);
+    QFontMetrics metrics(font);
+    mpEditor->setTabStopWidth(4*metrics.width(' '));
```
peterNordin: Maybe tab width should be configurable (no need to fix now) also I have started to prefer two spaces

robbr48: We could add that later if people request it. This is only for loading existing files anyway. The default Qt behavior was to show tabs as ~10 characters, which looked ridiculous.

PR #1867, opened 2019-08-29
Repair drag-copy feature

robbr48:
Drag-copying with right mouse button has not been working since #1859.

I solved it by delaying the context menu for `Component` class in the same was as was already done for the `ModelObject` class. In this way the mouse press event is not "stolen".


robbr48:
If I remember correctly, context menus behave differently in Windows and Linux. In Windows it is triggered on mouse release, but on Linux on mouse press. So this hack was probably added to make the menu appear on release also in Linux, when we know whether or not a drag event was triggered.

PR #1869, opened 2019-09-09
General improvements to component editing in HopsanGUI

### LINKED ISSUE #1868: Improvements to component editing in HopsanGUI
General improvements to component

- [x] New component dialog should keep its settings
- [x] Indentation of generated code is weird
- [x] Some variable names in generated code are wrong
- [x] Simplify variable names in generated code
- [x] Make it possible to create a new library when no external library is open
- [x] Support tabbing in text editor with multiple rows selected
- [x] Shift-tab does not work in text editor
- [x] Make default values in new component dialog zero by default
- [x] Don't close new component dialog when failed to add component

robbr48:
Resolves #1868.


HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -399,17 +399,47 @@ void TextEditor::keyPressEvent(QKeyEvent* event)
...
         {
-            this->textCursor().deletePreviousChar();
-            line = this->textCursor().block().text().toLatin1();
-            pos = textCursor().positionInBlock();
-            if(pos%2 == 0)
-            {
-                break;
+            auto cursor = textCursor();
+            QString text = cursor.selection().toPlainText();
+            if(text.startsWith(" ")) {
```
peterNordin: I dont understand this, why is always two initial spaces removed, OK maybe I understand, its two spaces at the first line in the selection, since it does not start with a new line

HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -399,17 +399,47 @@ void TextEditor::keyPressEvent(QKeyEvent* event)
...
+                text.remove(0,1);
+            }
+            if(text.startsWith(" ")) {
+                text.remove(0,1);
             }
+            text.replace("\n ","\n");
+            text.replace("\n ","\n");
```
peterNordin: Why do this twice? it is unclear to me what this is for, something with indentation I assume

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
```diff
@@ -148,25 +148,34 @@ QString HopsanGeneratorBase::generateSourceCodefromComponentSpec(ComponentSpecif
     QString varDeclarations;
     for(int i=0; i<comp.parNames.size(); ++i)
     {
-        varDeclarations.append("        double "+comp.parNames[i]+";\n");
+        if(!varDeclarations.isEmpty()) {
+            varDeclarations.append("        ");
```
peterNordin: hard-coding indentation is bizarre, but we can fix this in "the future", good for now.
I have written some automation for this in some place, but  I dont remember where, and probably not applicable to this code

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
```diff
@@ -193,7 +210,7 @@ QString HopsanGeneratorBase::generateSourceCodefromComponentSpec(ComponentSpecif
         QString id = QString::number(portId);
         if(comp.portNodeTypes[i] == "NodeSignal")
         {
-            allVarNames << comp.portNames[i]+"__y";
+            allVarNames << "*mp"+comp.portNames[i]+"";
```
peterNordin: This variable now has slightly confusing name (allVarNames), it is no longer variable names, but actually part of the declaration, since * is part of the "name"

HopsanGUI/Widgets/TextEditorWidget.cpp
robbr48: We want to remove two spaces at the beginning, but if there is only one spaces we only want to remove one. Therefore the easiest solution is to iteratively remove one space at a time. If we want to change tab width later we could replace this by a loop.

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
robbr48: Sure, we can add a setting for indentation size later. It should then affect both the generated code and tabbing in the text editor.

PR #1871, opened 2019-09-16
Copy all documentation graphics files to HTML directory

robbr48:
This is needed for building documentation to the new `documentation` repository.

All graphics files for the documentation will now be copied into the `/html`directory. In this way we avoid relative paths (`../graphics/image.png`), which does not work in the new repository.

### DIFF
--- a/doc/Doxyfile
+++ b/doc/Doxyfile
@@ -1149,7 +1149,7 @@ HTML_EXTRA_STYLESHEET  = hopsan.css
 # files will be copied as-is; there are no commands or markers available.
 # This tag requires that the tag GENERATE_HTML is set to YES.
 
-HTML_EXTRA_FILES       =
+HTML_EXTRA_FILES       = graphics/mainwindow.png graphics/graphics/library.png graphics/graphics/tfanalysis_icon.png graphics/fft_icon.png graphics/sensitivity_icon.png graphics/sensitivity_dialog.png graphics/sensitivity_plot.png graphics/optimization_icon.png graphics/ports.png graphics/node.png graphics/systemparameters_icon.png graphics/configuration_icon.png graphics/simulate_icon.png graphics/plot_icon.png graphics/favoritevariable_icon.png graphics/toggleplotvariables_icon.png graphics/plotresettimevector_icon.png graphics/zoomplot_icon.png graphics/panplot_icon.png graphics/saveplot_icon.png graphics/loadplot_icon.png graphics/exportplot_icon.png graphics/exportplotgfx_icon.png graphics/toggleplotcurvecontrolpanel_icon.png graphics/export_simulink_icon.png graphics/export_fmu_icon.png graphics/import_fmu_icon.png graphics/energylosses_icon.png graphics/energylosses_model1.png graphics/energylosses_barplot1.png graphics/energylosses_model2.png graphics/energylosses_barplot2.png graphics/tlm.png graphics/cq.png graphics/animation_icon.png
 
 # The HTML_COLORSTYLE_HUE tag controls the color of the HTML output. Doxygen
 # will adjust the colors in the style sheet and background images according to
--- a/doc/Doxyfile
+++ b/doc/Doxyfile
@@ -1,4 +1,4 @@
-# Doxyfile 1.8.11
+﻿# Doxyfile 1.8.11
 
 # This file describes the settings to be used by the documentation system
 # doxygen (www.doxygen.org) for a project.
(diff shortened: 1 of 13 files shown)


doc/Doxyfile
```diff
@@ -1149,7 +1149,7 @@ HTML_EXTRA_STYLESHEET  = hopsan.css
...
 # This tag requires that the tag GENERATE_HTML is set to YES.
 
-HTML_EXTRA_FILES       =
+HTML_EXTRA_FILES       = graphics/mainwindow.png graphics/graphics/library.png graphics/graphics/tfanalysis_icon.png graphics/fft_icon.png graphics/sensitivity_icon.png graphics/sensitivity_dialog.png graphics/sensitivity_plot.png graphics/optimization_icon.png graphics/ports.png graphics/node.png graphics/systemparameters_icon.png graphics/configuration_icon.png graphics/simulate_icon.png graphics/plot_icon.png graphics/favoritevariable_icon.png graphics/toggleplotvariables_icon.png graphics/plotresettimevector_icon.png graphics/zoomplot_icon.png graphics/panplot_icon.png graphics/saveplot_icon.png graphics/loadplot_icon.png graphics/exportplot_icon.png graphics/exportplotgfx_icon.png graphics/toggleplotcurvecontrolpanel_icon.png graphics/export_simulink_icon.png graphics/export_fmu_icon.png graphics/import_fmu_icon.png graphics/energylosses_icon.png graphics/energylosses_model1.png graphics/energylosses_barplot1.png graphics/energylosses_model2.png graphics/energylosses_barplot2.png graphics/tlm.png graphics/cq.png graphics/animation_icon.png
```
peterNordin: It would be better if this list could be auto generated by the buildDocumentation scripts, at least for bash it should be easy. For batch not so easy, but maybe we can require that a working bash terminal is available on Windows, and call the bash script to do that.

Something like, (should be 7 spaces before =): 
`files=$(find graphics -type f)`
`sed "s/HTML_EXTRA_FILES       =/HTML_EXTRA_FILES       = ${files}/" Doxyfile >test`

But I could not get this to work with git bash on windows for some reason

doc/userAnimation.dox
```diff
@@ -3,7 +3,7 @@
 
 Hopsan models can be animated, either in real-time or as replays. Animation is accessed through the Animate icon in the simulation toolbar, or from the menus.
 
-\htmlimagerightcaption{../graphics/animation_icon.png, Animate}
+\htmlimagerightcaption{animation_icon.png, Animate}
```
peterNordin: Is it not possible to use relative path  graphics/animation_icon.png  (without the leading ../) then maybe we could just copy/move the graphics directory in the build step?

robbr48: I guess that would work too. But Doxygen already supports copying files using HTML_EXTRA_FILES command, so that's why I decided to use it.

doc/Doxyfile
robbr48: I tried too, but it does not work for me either. One problem is that ${files} contain slashes, so we need to use some other delimiting character.  But I still cannot make it work.

robbr48: Ok I fixed it on Windows. The `find` command separates the output files with newlines instead of spaces. You need an intermediate step where newlines are replaced by spaces:
```
export files=$(find graphics -type f)
files2="${files//$'\n'/ }"
sed "s@HTML_EXTRA_FILES       =@HTML_EXTRA_FILES       = $files2@" Doxyfile >test
```
I have not tested this on Linux.

peterNordin: Hmm I am confused, this list does not include for instance "Edit Toolbar.png" but still that file gets copied to the html output directory. Non of the images with space in the path are in the list but they still show up. (Space was also a problem for the find command example I gave)

peterNordin: OK I see the difference now in the doxy file

\image html "File Toolbar.png"
vs
\htmllinkimage{mainwindow.png, 500}

Using \image, the file is copied automatically, even if not in the list

robbr48: Exactly. If we could live with having all pictures in full size the copying would not be needed. But that looked weird, and, also we could not have images with text to the right of them which is used for toolbar icons.

peterNordin: The following command work in Kubuntu, should work on windows (git bash) as well, you were right there were new-lines in the find results.
`files=$(find graphics -type f) && files=$(echo $files)`
`sed "s|HTML_EXTRA_FILES\s*= |HTML_EXTRA_FILES = ${files}|" -i Doxyfile`
It doe not solve image names with space in them, but those seem to work anyway, and maybe the should be renamed to not have spaces.

peterNordin: Seem to work in windows git bash also

peterNordin: But the script runs in batch so... maybe it doesn't matter, we rarely add new images anyway and all the documentation needs a refresh anyway.
Or maybe copying the graphics directory and using relative paths will work. Do as you wish.

robbr48: I added a Shell script in Linux and a PowerShell script for Windows. This requires Windows users to have PowerShell, but it is available for all currently supported versions of Windows (i.e. Windows 7 and above). I think this is better than to rely on bash in Windows.

peterNordin: OK, soon maybe bash will work out-of-the-box on windows also, with subsystem for linux, but I have not tried that yet

peterNordin: Actually I cant run this because my windows blocks calling of power-shell scripts due to security reasons

peterNordin: Changeing to PowerShell.exe -ExecutionPolicy Bypass -File ./copyGraphics.ps1
Seem to solve that problem

peterNordin: I made a PR to your branch

PR #1872, opened 2019-09-19
Generating and compiling Modelica components

robbr48:
Components can now have .mo files as source files. When compiling the library, the Modelica code is translated to C++ code and saved as a .hpp file with same name. This file is then used in the actual library compilation.

Right now there is no error handling for the Modelica code translation. This remains a future project.


HopsanGenerator/src/HopsanGeneratorLib.cpp
```diff
@@ -456,7 +483,7 @@ bool callAddComponentToLibrary(const char* libraryXmlPath, const char* targetPat
 
     //Add new component to library
     lib.mComponentXMLFiles.append(QDir(xmlPath.absolutePath()).relativeFilePath(QFileInfo(cafPath).absoluteFilePath()));
-    lib.mComponentCodeFiles.append(QFileInfo(hppPath).fileName());
+    lib.mComponentCodeFiles.append(QFileInfo(sourcePath).fileName().replace(".mo", ".hpp"));
```
peterNordin: What if a c++ source file has  .mo in its name,   like:   my-super.mo-sourcefile.hpp
Replace if ends with .mo ?

HopsanGenerator/src/generators/HopsanGeneratorBase.cpp
```diff
@@ -1116,6 +1119,90 @@ bool HopsanGeneratorBase::generateLibrarySourceFile(const ComponentLibrary &lib)
...
+            outStream << indent << portName+".p = "+portName+".c + "+portName+".Zc*"+portName+".Qdot;\n";
+        }
+    }
+
+    outStream << indent << "\n";
+    outStream << "algorithm\n";
+    outStream << indent << "\n";
+    outStream << "end " << comp.typeName << ";\n";
+
+//              equation
```
peterNordin: Comments ?

PR #1873, opened 2019-10-01
Update indexing csv parser dependency

peterNordin:
Update indexingcsvparser library

- Now support loading values from white-space padded columns
- Supports header extraction (not used in Hopsan yet)
- Can optionally strip leading and trailing white-spaces when extracting string data (not used in Hopsan yet)


peterNordin:
Yes that is the only new thing that can be used in hopsan right now. The library it-self has other built in tests that check other things.

PR #1877, opened 2019-10-12
Issue 1839 lookuptable textinput

### LINKED ISSUE #1839: Lookup table component with string CSV parameters
Sometimes you just want to make a quick lookup table, without the hassle of external files.

Lookup table component(s?) with string parameters accepting comma-separated values. These should be parsed at initialize and stored in arrays.

peterNordin:
At textblock parameter type and GUI edit capability to lookup table components
Resolves #1839

### DIFF
--- a/HopsanCore/include/Component.h
+++ b/HopsanCore/include/Component.h
@@ -135,6 +135,7 @@ public:
     void addConstant(const HString &rName, const HString &rDescription, const HString &rUnit, const HString &defaultValue, HString &rData);
     void addConstant(const HString &rName, const HString &rDescription, const HString &rUnit, const bool defaultValue, bool &rData);
     void addConstant(const HString &rName, const HString &rDescription, const HString &rQuantity, const HString &rUnit, const double defaultValue, double &rData);
+    void addConstant(const HString &rName, const HString &rDescription, HTextBlock &rData);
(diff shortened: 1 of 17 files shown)


peterNordin:
Each component has its own format, lookup 1d, 2d, 3d and plo. The format is already available in the component description. Maybe the description should be shown alongside the edit dialog, or a button that opens it can be added. 

The lines must en in a newline charcter, but those can not be entered in the main dilog, which is not suitable for this type of input anyway.
The main dialog seems to show the new-line characters as spaces, but I don't know what may happen if you try to edit the text inline, it might even break the text. Maybe the main text edit should be removed for this type of parameters. Or at least set to read-only. It is convenient to be able to hover over it to see the entire text without having to open the dilog.

robbr48:
I think removing the line edit is the best solution. If it confuses me, it will for sure confuse other people as well :)

peterNordin:
I made it read-only, so you can still hover it and see a tooltip with the data. Or select all and copy to external file or similar. Linebreaks seem to be preserved.

peterNordin:
Also a fun thing I discovered, if you want to create a system parameter with CSV data, no special editor is available, but you can past multiple lines from a text editor into the text edit field and the line breaks will be preserved (but look like a space). Will need to fix a dialog for system parameters also, but maybe later.

PR #1882, opened 2019-10-20
Fix not being able to import FMU, due to new 0.3 Library XML format

peterNordin:
Exported FMUs cant be imported after the change to 0.3 library xml format. The reason is that the CAF files are not included in the import generated library xml. This may also be a problem with other import generators.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ bin/
 HopsanNG_bd/
 output64/
 output/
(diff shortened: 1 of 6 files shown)


peterNordin:
@robbr48 I need your input on this PR, please comment on my changes and especially the todo's I added, also I am not sure if my solution in the Modelica import code is valid.

PR #1888, opened 2020-01-07
Remove limitation on animation speed setting

### LINKED ISSUE #1346: Animation: Real-time animation should be able to run with unlimited speed
---

Original Redmine Issue: 1346, https://flumes.iei.liu.se/redmine/issues/1346
Original Date: 2014/10/15
Original Assignee: Robert Braun

---

None

robbr48:
Animation speed was for some reason limited to 20. Now it is unlimited.

Resolves #1346.

### DIFF
--- a/HopsanGUI/Widgets/AnimationWidget.cpp
+++ b/HopsanGUI/Widgets/AnimationWidget.cpp
@@ -129,8 +129,7 @@ AnimationWidget::AnimationWidget(QWidget *parent) :
     mpTimeSlider = new QSlider(Qt::Horizontal);
 
     mpSpeedSpinBox = new QDoubleSpinBox(this);
-    mpSpeedSpinBox->setMinimum(-20.0);
-    mpSpeedSpinBox->setMaximum(20.0);
+    mpSpeedSpinBox->setRange(std::numeric_limits<double>::lowest(),std::numeric_limits<double>::max());
     mpSpeedSpinBox->setDecimals(5);
     mpSpeedSpinBox->setSingleStep(0.01);
     mpSpeedSpinBox->setValue(1.0);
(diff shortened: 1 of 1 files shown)


HopsanGUI/Widgets/AnimationWidget.cpp
```diff
@@ -129,8 +129,7 @@ AnimationWidget::AnimationWidget(QWidget *parent) :
     mpTimeSlider = new QSlider(Qt::Horizontal);
 
     mpSpeedSpinBox = new QDoubleSpinBox(this);
-    mpSpeedSpinBox->setMinimum(-20.0);
-    mpSpeedSpinBox->setMaximum(20.0);
+    mpSpeedSpinBox->setRange(std::numeric_limits<double>::min(),std::numeric_limits<double>::max());
```
peterNordin: This prevents negative values, is that intended?
std::numeric_limits<double>::min() is a positive very small value
Otherwise maybe the setRange call can be completely removed

But if the reason is to prevent 0 (and negative scaling) then OK

peterNordin: Aha, just looked at the issue, for realtime, I guess you do not want negative values.
What about playbak, is that an other spinbox? (did not check code) 
Because pause and reverse could be useful then

robbr48:
@petno I changed it to allow negative speed, and added a check so that real-time playback cannot be started if speed is negative.

HopsanGUI/Widgets/AnimationWidget.cpp
```diff
@@ -433,6 +432,10 @@ void AnimationWidget::play()
 //! @todo The plot data object in container should perhaps be cleared of these generations
 void AnimationWidget::playRT()
 {
+    if(mpSpeedSpinBox->value() < 0) {
```
peterNordin: I have not had time to test or look at the code in detail, but what happens if you change the speed during real-time simulation, to a negative value.
Will it stop, or just create lots of errors ?

Maybe limiting the lowest value to zero in RT simulation where the value goes into to drive the simulation.

robbr48: It will simply not simulate at all in RT with negative speed. If setting it to negative during RT animation, the animation will stop at the current position and the time will start rewinding without anything happening. If switching back to positive, the animation will continue once the animation time catches up with the last working animation time. I think we can live with this for now, as changing the limitations depending on animation type would be complicated.

PR #1890, opened 2020-01-13
Added some mathematical functions for vectors in HCOM

robbr48:
These functions already existed for scalars but did not work for vectors, which was causing confusion.

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -6042,7 +6042,7 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
         if(pData) {
             mAnsType = HcomHandler::DataVector;
             mAnsVector = pLogDataHandler->createOrphanVariable(QString(funcName+"%1").arg(pData->getSmartName()), pData->getVariableType());
-            mAnsVector->assignFrom(pData->getSharedTimeOrFrequencyVector(), invokeMathFunctionOnData(pData.data(),func));
+            mAnsVector->assignFrom(pData->getSharedTimeOrFrequencyVector(), pData.data()->invokeMathFunctionOnData(func));
             return;
         }
     }
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -6010,24 +6010,23 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
 
     //Simple mathematical vector functions with no arguments
     FuncMap_t funcMap;
-    funcMap.insert("sin", qMakePair(static_cast<ScalarMathFunction_t>(sin),&VectorVariable::sinOfData));
-    funcMap.insert("cos", qMakePair(static_cast<ScalarMathFunction_t>(cos),&VectorVariable::cosOfData));
-    funcMap.insert("tan", qMakePair(static_cast<ScalarMathFunction_t>(tan),&VectorVariable::tanOfData));
-    funcMap.insert("asin", qMakePair(static_cast<ScalarMathFunction_t>(asin),&VectorVariable::asinOfData));
-    funcMap.insert("acos", qMakePair(static_cast<ScalarMathFunction_t>(acos),&VectorVariable::acosOfData));
-    funcMap.insert("atan", qMakePair(static_cast<ScalarMathFunction_t>(atan),&VectorVariable::atanOfData));
-    funcMap.insert("log", qMakePair(static_cast<ScalarMathFunction_t>(log),&VectorVariable::logOfData));
-    funcMap.insert("exp", qMakePair(static_cast<ScalarMathFunction_t>(exp),&VectorVariable::expOfData));
-    funcMap.insert("sqrt", qMakePair(static_cast<ScalarMathFunction_t>(sqrt),&VectorVariable::sqrtOfData));
-    funcMap.insert("round", qMakePair(static_cast<ScalarMathFunction_t>(round),&VectorVariable::roundOfData));
-    funcMap.insert("floor", qMakePair(static_cast<ScalarMathFunction_t>(floor),&VectorVariable::floorOfData));
-    funcMap.insert("ceil", qMakePair(static_cast<ScalarMathFunction_t>(ceil),&VectorVariable::ceilOfData));
-    funcMap.insert("abs", qMakePair(static_cast<ScalarMathFunction_t>(fabs),&VectorVariable::absOfData));
+    funcMap.insert("sin", static_cast<ScalarMathFunction_t>(sin));
+    funcMap.insert("cos", static_cast<ScalarMathFunction_t>(cos));
+    funcMap.insert("tan", static_cast<ScalarMathFunction_t>(tan));
+    funcMap.insert("asin", static_cast<ScalarMathFunction_t>(asin));
+    funcMap.insert("acos", static_cast<ScalarMathFunction_t>(acos));
+    funcMap.insert("atan", static_cast<ScalarMathFunction_t>(atan));
+    funcMap.insert("log", static_cast<ScalarMathFunction_t>(log));
+    funcMap.insert("exp", static_cast<ScalarMathFunction_t>(exp));
+    funcMap.insert("sqrt", static_cast<ScalarMathFunction_t>(sqrt));
+    funcMap.insert("round", static_cast<ScalarMathFunction_t>(round));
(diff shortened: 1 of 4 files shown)


HopsanGUI/LogVariable.cpp
```diff
@@ -1774,6 +1774,114 @@ QVector<double> VectorVariable::absOfData() const
...
+    QVector<double> *pData = mpCachedDataVector->beginFullVectorOperation();
+    QVector<double> sqrtdata(pData->size());
+    for (int i=0; i<pData->size(); ++i)
+    {
+        sqrtdata[i] = qSqrt((*pData)[i]);
+    }
+    mpCachedDataVector->endFullVectorOperation(pData);
+    return sqrtdata;
+}
+
```
peterNordin: Since all of these blocks are the same, a help function can be used, and the unique part (the math function) could be sent in as a lambda or as a template parameter in the help function. If you have the time, otherwise OK. But I guess that the existing functions are also like this.

robbr48: Good point, I replaced all the single-argument function implementations with a map of function pointers.

peterNordin: I was referring to the ones in LogVariable, but sure the HCOM functions were also nearly identical

You could do something like:


`template <typename Function>`
` QVector<double> cleverFunctionName(CachedDataVectorType& pCachedDataVector, Function func )`
`    QVector<double> *pData = pCachedDataVector->beginFullVectorOperation();`
`    QVector<double> returndata(pData->size());`
`    for (int i=0; i<pData->size(); ++i)`
`    {`
`          returndata[i] = func((*pData)[i]);`
`    }`
`    pCachedDataVector->endFullVectorOperation(pData);`
`    return returndata;`

Then call with:
`return cleverFunctionName(pCachedDataVector, qCos)`

Hopefully I got this right, just writing out of memory

HopsanGUI/HcomHandler.cpp
```diff
@@ -6006,124 +6008,51 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
                 return;
             }
         }
 
         if(pData)
         {
             mAnsType = HcomHandler::DataVector;
-            mAnsVector = pLogDataHandler->createOrphanVariable(QString("Ceil%1").arg(pData->getSmartName()), pData->getVariableType());
-            mAnsVector->assignFrom(pData->getSharedTimeOrFrequencyVector(), pData->ceilOfData());
+            LogDataHandler2 *pLogDataHandler = mpModel->getViewContainerObject()->getLogDataHandler().data();
```
peterNordin: Is this line needed ? Was it not called on line 6008

HopsanGUI/LogVariable.cpp
robbr48: Actually my commit deals with both LogVariable functions and scalar math functions, since each HCOM function call requires both. They first try to find a vector with specified name, and if it fails they look for a scalar. Hence, each function call needs two function pointer, one for the vector operation and one scalar. 

Also, I decided to use cmath functions intead of Qt equivalents, since the Qt functions have different return types while cmath functions all return double (for example, Qt's "floor" function returns an integer).

HopsanGUI/HcomHandler.cpp
```diff
@@ -6006,124 +6008,51 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
+    funcMap.insert("acos", qMakePair(static_cast<ScalarMathFunction_t>(acos),&VectorVariable::acosOfData));
+    funcMap.insert("atan", qMakePair(static_cast<ScalarMathFunction_t>(atan),&VectorVariable::atanOfData));
+    funcMap.insert("log", qMakePair(static_cast<ScalarMathFunction_t>(log),&VectorVariable::logOfData));
+    funcMap.insert("exp", qMakePair(static_cast<ScalarMathFunction_t>(exp),&VectorVariable::expOfData));
+    funcMap.insert("sqrt", qMakePair(static_cast<ScalarMathFunction_t>(sqrt),&VectorVariable::sqrtOfData));
+    funcMap.insert("round", qMakePair(static_cast<ScalarMathFunction_t>(round),&VectorVariable::roundOfData));
+    funcMap.insert("floor", qMakePair(static_cast<ScalarMathFunction_t>(floor),&VectorVariable::floorOfData));
+    funcMap.insert("ceil", qMakePair(static_cast<ScalarMathFunction_t>(ceil),&VectorVariable::ceilOfData));
+    funcMap.insert("abs", qMakePair(static_cast<ScalarMathFunction_t>(fabs),&VectorVariable::absOfData));
+    if(desiredType != Scalar && funcMap.contains(getFunctionName(expr))) {
```
peterNordin: Previously isHcomFunctionCall("round", expr) and similar was called. That function does additional counting of parenthesis which getFunctionName does not seem to do, unless it solves that problem in some other way.

I dont remember when exactly but at some point I was forced to add counting to avoid treating a sub-part of an expression as a function. It might have been in case of a function taking multiple arguments, where the arguments were them selves expressions taking expressions as arguments. 
Or it could have been to avoid calling a function when arguments where function calls but the wrong number of parenthesis were present

Anyway, I would like that check back somehow, maybe just add it to getFunctionName, so that it returns nothing the parenthesis do not match. I think that should give the same behavior as before.

We really need test for this. I think that I have made a test model / script for these kind of things but I can not find it now. Not sure if those cases were included there though. ("Later" work)

peterNordin: I found this comment in isHcomFunctionCall
`// When nOpen == 0 we should have closed the function, and we should be at end of expr, else this is an expression and not a clean function call`
It likely has to do with the interpretation becoming confused when a function call contains sub-expressions with other functions in them. In some way...

HopsanGUI/LogVariable.cpp
robbr48: I also implemented your suggestion, so that all "...OfData()" methods in LogVariable class are replaced with a single template function.

HopsanGUI/HcomHandler.cpp
robbr48: No, you are right. I will remove it.

HopsanGUI/HcomHandler.cpp
robbr48: I solved it the easy way by adding a call to `isHcomFunctionCall` inside `getFunctionName`. It returns an empty string if not a function.

HopsanGUI/HcomHandler.h
```diff
@@ -102,6 +103,9 @@ class HcomHandler : public QObject
 
     void evaluateExpression(QString expr, VariableType desiredType=Undefined);
     double evaluateScalarExpression(QString expr, bool &rIsOK);
+    typedef double (*ScalarMathFunction_t)(double);
+    typedef QMap<QString, ScalarMathFunction_t> FuncMap_t;
+    void invokeSingleArgumentFunction(QString funcName, QString argStr, ScalarMathFunction_t f1);
```
peterNordin: This functions is not defined, only declared, and do not seem to be used, obviously otherwise it would not compile. Assuming left-over, that should be removed.

HopsanGUI/LogVariable.h
```diff
@@ -379,5 +376,17 @@ SharedVectorVariableT switchVariableGeneration(SharedVectorVariableT pVar, int g
 double pokeVariable(SharedVectorVariableT a, const int index, const double value);
 double peekVariable(SharedVectorVariableT a, const int b);
 
+template<typename Function>
+QVector<double> invokeMathFunctionOnData(VectorVariable *pVector, Function func)
```
peterNordin: I am not sure about this idea, but I think this one could have been a template member function in the VectorVariable class, then there would not have been a need to expose the getCachedDataVector(), but not so important.

HopsanGUI/LogVariable.cpp
peterNordin: Yes cmath is better in this case :)

HopsanGUI/HcomHandler.h
peterNordin: Approved, after removing or correcting this minor issu

robbr48: My bad, I removed the declaration.

HopsanGUI/LogVariable.h
robbr48: Good point. I moved `invokeMathFunctionOnData` to a member function and removed `getCachedDataVector`.

PR #1894, opened 2020-01-15
More advanced frequency analysis functions in HCOM

### LINKED ISSUE #1891: More frequency spectrums
The fft() HCOM function should be extended to support RMS, density and energy spectrums.

Perhaps the function should also be renamed to something more understandable.

robbr48:
- Energy density spectrum
- Power density spectrum
- Root mean square of vector

Resolves #1891.

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -364,6 +364,7 @@ HcomHandler::HcomHandler(TerminalConsole *pConsole) : QObject(pConsole)
     registerInternalFunction("ddt", "Differentiates vector with respect to time (or to custom vector)","Usage: ddt(vector)\nUsage: ddt(vector, timevector)");
     registerInternalFunction("int", "Integrates vector with respect to time (or to custom vector)", "Usage: int(vector)\nUsage: int(vector, timevector)");
     registerInternalFunction("fft", "Generates frequency spectrum plot from vector","Usage: fft(vector)\nUsage: fft(vector, power[true/false])\nUsage: fft(vector, timevector)\nUsage: fft(vector, timevector, power[true/false])");
+    registerInternalFunction("rms", "Computes the root mean square of given vector","Usage: rms(vector)");
     registerInternalFunction("gt", "Index-wise greater than check between vectors and/or scalars (equivalent to \">\" operator)","Usage: gt(varName, threshold)\nUsage: gt(var1, var2)");
     registerInternalFunction("lt", "Index-wise less than check between vectors and/or scalars  (equivalent to \"<\" operator)","Usage: lt(varName, threshold)\nUsage: lt(var1,var2)");
     registerInternalFunction("eq", "Index-wise fuzzy equal check between vectors and/or scalars  (equivalent to \"==\" operator)","Usage: eq(varName, threshold, eps)\nUsage: eq(var1, var2, eps)");
@@ -6587,6 +6588,31 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
             mAnsVector = pDataVar->toFrequencySpectrum(pTimeVar, type, windowingFunction, minTime, maxTime);
         }
         return;
+    }
+    else if(isHcomFunctionCall("rms", expr))
+    {
+        QString args = expr.mid(4, expr.size()-5);
+        QStringList splitArgs = SymHop::Expression::splitWithRespectToParentheses(args, ',');
+        if(splitArgs.size() != 1) {
+            HCOMERR("Wrong number of arguments provided for rms function.\n"+mLocalFunctionDescriptions.find("fft").value().second);
+            mAnsType = Undefined;
+            return;
+        }
+        mAnsType = Scalar;
+        const QString varName = args.section(",",0,0).trimmed();
+        evaluateExpression(varName, DataVector);
+        SharedVectorVariableT pVar = mAnsVector;
+        if (mAnsType == DataVector) {
+            mAnsType = Scalar;
+            mAnsScalar = pVar->rmsOfData();
+            return;
+        }
+        else
+        {
+            HCOMERR(QString("Variable: %1 was not found!").arg(varName));
+            mAnsType = Undefined;
+            return;
+        }
     }
     //else if(desiredType != Scalar && (expr.startsWith("greaterThan(") || expr.startsWith("gt(")) && expr.endsWith(")"))
     else if(desiredType != Scalar && (isHcomFunctionCall("greaterThan", expr) || isHcomFunctionCall("gt", expr)) )
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -363,7 +363,10 @@ HcomHandler::HcomHandler(TerminalConsole *pConsole) : QObject(pConsole)
     registerInternalFunction("lp1", "Applies low-pass filter of first degree to vector","Usage: lp1(vector, frequency)\nUsage: lp1(vector, timevector, frequency)");
     registerInternalFunction("ddt", "Differentiates vector with respect to time (or to custom vector)","Usage: ddt(vector)\nUsage: ddt(vector, timevector)");
     registerInternalFunction("int", "Integrates vector with respect to time (or to custom vector)", "Usage: int(vector)\nUsage: int(vector, timevector)");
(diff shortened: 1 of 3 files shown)


robbr48:
The HCOM `fft()` function is now split into three functions for energy, power and RMS spectral density:
```
psd(vector, timevector, windowing[optional], mintime[optional] maxtime[optional])
esd(vector, timevector, windowing[optional], mintime[optional] maxtime[optional])
rmsd(vector, timevector, windowing[optional], mintime[optional] maxtime[optional])
```
where
```
windowing = flattop/rectangular/hann
```

HopsanGUI/HcomHandler.cpp
```diff
@@ -363,7 +363,10 @@ HcomHandler::HcomHandler(TerminalConsole *pConsole) : QObject(pConsole)
     registerInternalFunction("lp1", "Applies low-pass filter of first degree to vector","Usage: lp1(vector, frequency)\nUsage: lp1(vector, timevector, frequency)");
     registerInternalFunction("ddt", "Differentiates vector with respect to time (or to custom vector)","Usage: ddt(vector)\nUsage: ddt(vector, timevector)");
     registerInternalFunction("int", "Integrates vector with respect to time (or to custom vector)", "Usage: int(vector)\nUsage: int(vector, timevector)");
-    registerInternalFunction("fft", "Generates frequency spectrum plot from vector","Usage: fft(vector)\nUsage: fft(vector, power[true/false])\nUsage: fft(vector, timevector)\nUsage: fft(vector, timevector, power[true/false])");
+    registerInternalFunction("esd", "Generates energy spectral density from vector","Usage: esd(vector, timevector [optional], windowing[rectangular/flattop/hann, optional], mintime[optional], maxtime[optional])\n");
```
peterNordin, line 366: For backwards compatibility maybe keep the original fft function, but add a deprecation warning message when calling it.
And make it call one of the new functions

HopsanGUI/HcomHandler.cpp
```diff
@@ -363,7 +363,10 @@ HcomHandler::HcomHandler(TerminalConsole *pConsole) : QObject(pConsole)
...
+    registerInternalFunction("psd", "Generates power spectral density from vector","Usage: psd(vector, timevector [optional], windowing[rectangular/flattop/hann, optional], mintime[optional], maxtime[optional])\n");
```
peterNordin, line 367: Is this a new way of noting optional parameters, before I think we did [name] to indicate optional. I have not tested this branch to see if this looks better though

HopsanGUI/HcomHandler.cpp
```diff
@@ -6439,154 +6442,126 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
-    else if(desiredType != Scalar && isHcomFunctionCall("fft", expr))
+    else if(desiredType != Scalar && (isHcomFunctionCall("esd", expr) || isHcomFunctionCall("psd", expr) || isHcomFunctionCall("rmsd", expr)))
     {
-        QString argStr = expr.mid(4, expr.size()-5);
-        QStringList args = SymHop::Expression::splitWithRespectToParentheses(argStr,',');
-        for(int a=0; a<args.size(); ++a) {
-            args[a] = args[a].trimmed();
+        QStringList splitArgs = extractFunctionCallExpressionArguments(expr);
+        if(splitArgs.size() < 1 || splitArgs.size() > 4) {
+            HCOMERR("Wrong number of arguments provided for esd function.\n"+mLocalFunctionDescriptions.find("esd").value().second);
```
peterNordin, line 6449: Message only mentions esd, but code runs also for psd and rmsd

HopsanGUI/HcomHandler.cpp
```diff
@@ -6439,154 +6442,126 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
-        {
-            evaluateExpression(timeVecArg, DataVector);
-            pTimeVar = mAnsVector;
-            if (mAnsType != DataVector)
-            {
-                HCOMERR(QString("Variable: %1 was not found!").arg(timeVecArg));
-                mAnsType = Undefined;
-                return;
+        if(splitArgs.size() > i) {
+            QString arg2 = splitArgs[1].trimmed();
```
peterNordin, line 6481: Index i or 1 ? in this case I think i==1 but looks odd

HopsanGUI/HcomHandler.cpp
```diff
@@ -6439,154 +6442,126 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
-        else if(windowingFuncArg.isEmpty()) {
-            mAnsVector = pDataVar->toFrequencySpectrum(pTimeVar, type);
+
+        mAnsType = DataVector;
+        mAnsVector = pVar->toFrequencySpectrum(pTimeVar, type, windowingFunction, minTime, maxTime);
+        return;
+    }
+    else if(isHcomFunctionCall("rms", expr))
+    {
+        QString args = expr.mid(4, expr.size()-5);
```
peterNordin, line 6543: Can we use extractFunctionCallExpressionArguments( instead of expr.mid ?
or does it not work with nested parentheses ?

HopsanGUI/HcomHandler.cpp
```diff
@@ -6439,154 +6442,126 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
+        QStringList splitArgs = SymHop::Expression::splitWithRespectToParentheses(args, ',');
+        if(splitArgs.size() != 1) {
+            HCOMERR("Wrong number of arguments provided for rms function.\n"+mLocalFunctionDescriptions.find("fft").value().second);
```
peterNordin, line 6546: Message say fft but function is rms

HopsanGUI/HcomHandler.cpp
```diff
@@ -6439,154 +6442,126 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
+            mAnsType = Undefined;
+            return;
         }
-        else if(minTimeArg.isEmpty()) {
-            mAnsVector = pDataVar->toFrequencySpectrum(pTimeVar, type, windowingFunction);
+        mAnsType = Scalar;
+        const QString varName = args.section(",",0,0).trimmed();
```
peterNordin, line 6551: rms seems to only have one argument, and did we not already split the args above ?

HopsanGUI/LogVariable.cpp
```diff
@@ -1216,6 +1216,20 @@ bool VectorVariable::positiveNonZeroMinMaxOfData(double &rMin, double &rMax, int
...
 
+double VectorVariable::rmsOfData() const
+{
+    QVector<double> *pData = mpCachedDataVector->beginFullVectorOperation();
+    double rms;
+    for (int i=0; i<pData->size(); ++i)
+    {
+        rms += (*pData)[i]*(*pData)[i];
+    }
+    rms /= pData->size();
```
peterNordin, line 1227: not sure if it can happen, but if size = 0 then this will cause unexpected results

HopsanGUI/HcomHandler.cpp
robbr48: Ok, I rewrote the fft functions it calls one of the other three, depending on the arguments.

HopsanGUI/HcomHandler.cpp
robbr48: No the old way is better. This is too verbose. I just didn't notice, since no other internal function have any optional arguments.

HopsanGUI/HcomHandler.cpp
robbr48: Fixed.

HopsanGUI/HcomHandler.cpp
robbr48: Fixed. It works with "1" too, but it is dangerous if someone decides to insert a new argument in the future.

HopsanGUI/HcomHandler.cpp
robbr48: Good point. It does handle nested parentheses internally, so that is no problem.

HopsanGUI/HcomHandler.cpp
robbr48: Fixed.

HopsanGUI/LogVariable.cpp
robbr48: I added a check which returns zero if data is empty. Also fixed uninitialized variable (`rms`).

peterNordin (approved):

PR #1895, opened 2020-01-16
Component for logging data to file

### LINKED ISSUE #1892: Component for logging signals to external file
It is not possible to use higher sampling frequencies for certain variables. This would be useful for e.g. frequency analysis.

A component with a multi-port signal input, which writes connected inputs to an external file. Parameters should include file name, sampling frequency and a comma-separated list of variable names.

Make sub-systems with larger number of samples work again, but this would be more difficult both to implement and use.

robbr48:
A new component with multi-port signal input which can log data to any of the following formats:
- Plain column-wise CSV
- Plain row-wise CSV
- Hopsan row-wise CSV
- Hopsan PLO version 3

Logging interval can be specified, as well as variable names. For Hopsan row-wise CSV, aliases and units can also be specified. 

Resolves #1892.

### DIFF
--- a/componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
+++ b/componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
@@ -1,0 +1,260 @@
+/*-----------------------------------------------------------------------------
+
+ Copyright 2017 Hopsan Group
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+
+ The full license is available in the file LICENSE.
+ For details about the 'Hopsan Group' or information about Authors and
+ Contributors see the HOPSANGROUP and AUTHORS files that are located in
+ the Hopsan source code root directory.
+
+-----------------------------------------------------------------------------*/
+
+//!
+//! @file   SignalLogger.hpp
+//! @author Robert Braun <robert.braun@liu.se>
+//! @date   2020-01-15
+//!
+//! @brief Contains a signal logger component
+//!
+//$Id$
+
+#ifndef SIGNALLOGGER_HPP_INCLUDED
+#define SIGNALLOGGER_HPP_INCLUDED
+
+#include "ComponentEssentials.h"
+#include "ComponentSystem.h"
+#include "ComponentUtilities.h"
+#include <fstream>
+#include <ctime>
+
+
+namespace hopsan {
+
+    //!
+    //! @brief
+    //! @ingroup SignalComponents
+    //!
+    class SignalLogger : public ComponentSignal
+    {
+
+    private:
+        Port *mpIn;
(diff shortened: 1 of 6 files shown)


componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
```diff
@@ -0,0 +1,258 @@
...
+                    mRows[mRows.size()-1].append(HString(ss.str().c_str()));
+                }
+            }
+        }
+
+        void finalize()
+        {
+            //Write number of rows for PLO file
+            if(mFileType == HopsanPLO) {
+                mRows[3] = "    "+HString(std::to_string(mNumVars+1).c_str())+ "    "+HString(std::to_string(mRows.size()-1).c_str());
```
peterNordin: Use "to_hstring" instead to_string is c++11

componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
```diff
@@ -0,0 +1,258 @@
...
+                //Extract target file name
+                HVector<HString> temp = mFilePath.split('/');
+                temp = temp[temp.size()-1].split('\\');
+                HString fileName = temp[temp.size()-1];
+
+                //Fetch parent system name (hmf filename is unknown)
+                HString modelName = getSystemParent()->getName();
+
+                //Generate date and time string
+                auto t = std::time(nullptr);
```
peterNordin: Lots of c++11 in this file, then we need to drop support for old MSVC compilers, and maybe really old GCC as well.
Would require dropping SIT export support also. But maybe not such bad thing ?

robbr48: I removed all C++11 stuff I think (auto, to_string and date/time stuff). I hope I did not miss anything. People are still using SIT export, so we should probably wait with allowing C++11 in (default) components.

componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
```diff
@@ -0,0 +1,260 @@
...
+                    mRows[i+1].append(aliasesVector[i]+",");
+                    mRows[i+1].append(unitsVector[i]);
+                }
+            }
+            if(mFileType == HopsanPLO) {
+                //Write custom PLO header rows
+
+                //Extract target file name
+                HVector<HString> temp = mFilePath.split('/');
+                temp = temp[temp.size()-1].split('\\');
```
peterNordin: Its unclear to me what happens here, does it split on both / and \, it looks like it extracts only the last character for the \\ split, maybe it already does, but it must support both / and \ file-paths

componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
peterNordin: OK

componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
robbr48: No, temp is a `HVector`, so it extracts the last `HString`. The purpose is to first chop the string at the last forward slash, and then chop the remainder at the last backslash. In this way it extracts the file name regardless of the kind of slashes. It could maybe be problematic for filenames with spaces on Linux, such as "/home/user/my\ model.hmf", but I don't think this will be a major issue.

peterNordin: Yes I see now what you mean

PR #1909, opened 2020-03-31
Support parameters in FMI export

robbr48:
- Specify parameters in `modelDescription.xml`
- Read parameters using `getReal`, `getInteger`, `getBoolean` and `getString`
- Write parameters using `setReal`, `setInteger`, `setBoolean` and `setString`

### DIFF
--- a/HopsanGenerator/src/GeneratorTypes.cpp
+++ b/HopsanGenerator/src/GeneratorTypes.cpp
@@ -27,6 +27,7 @@
 
 #include "HopsanEssentials.h"
 #include "Port.h"
+#include "Parameters.h"
 
 #include <QDir>
 #include <QFile>
@@ -728,3 +729,30 @@ QString BuildFlags::platformString() const
 //{
 //    return compilerString(mCompiler);
 //}
+
+void getParameters(QList<ParameterSpecification> &parameters, hopsan::ComponentSystem *pSystem)
+{
+    for(hopsan::ParameterEvaluator *par : (*pSystem->getParametersVectorPtr())) {
+        ParameterSpecification spec;
+        if(par->getType() == "double") {
+            spec.type = "Real";
+        }
+        else if(par->getType() == "integer") {
+            spec.type = "Integer";
+        }
+        else if(par->getType() == "bool") {
+            spec.type = "Boolean";
+        }
+        else if(par->getType() == "string") {
+            spec.type = "String";
+        }
+        else {
+            continue;
+        }
+        spec.name = par->getName().c_str();
+        spec.displayName = par->getName().c_str();
+        spec.init = par->getValue().c_str();
+        spec.description = par->getDescription().c_str();
+        parameters.append(spec);
+    }
+}
--- a/HopsanGenerator/src/GeneratorTypes.cpp
+++ b/HopsanGenerator/src/GeneratorTypes.cpp
@@ -746,6 +746,12 @@ void getParameters(QList<ParameterSpecification> &parameters, hopsan::ComponentS
         else if(par->getType() == "string") {
             spec.type = "String";
         }
+        else if(par->getType() == "textblock") {
+            spec.type = "String";
+        }
+        else if(par->getType() == "conditional") {
+            spec.type = "Integer";
+        }
         else {
             continue;
         }
--- a/HopsanGenerator/src/GeneratorTypes.cpp
+++ b/HopsanGenerator/src/GeneratorTypes.cpp
@@ -759,6 +759,7 @@ void getParameters(QList<ParameterSpecification> &parameters, hopsan::ComponentS
         spec.displayName = par->getName().c_str();
         spec.init = par->getValue().c_str();
         spec.description = par->getDescription().c_str();
+        spec.unit = par->getUnit().c_str();
         parameters.append(spec);
     }
 }
--- a/HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
+++ b/HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
@@ -113,6 +113,17 @@ typedef struct
     QString fmuVr;
     QString description;
     QString unit;
+    fmi1_base_type_enu_t dataType;
+    fmi1_import_variable_t *pFmiVariable;
+} hopsan_fmi1_import_variable_t;
+
+typedef struct
+{
+    QString name;
+    QString variableName;
+    QString fmuVr;
+    QString description;
+    QString unit;
     fmi2_base_type_enu_t dataType;
     fmi2_import_variable_t *pFmiVariable;
 }hopsan_fmi_import_variable_t;
@@ -279,7 +290,7 @@ bool HopsanFMIGenerator::generateFromFmu1(const QString &rFmuPath, const QString
     }
 
     //Declare lists for parameters, input variables and output variables
-    QStringList parNames, parVars, parRefs;
+    QList<hopsan_fmi1_import_variable_t> pars;
     QStringList inputNames, inputVars, inputRefs;
     QStringList outputNames, outputVars, outputRefs;
 
@@ -293,12 +304,18 @@ bool HopsanFMIGenerator::generateFromFmu1(const QString &rFmuPath, const QString
         fmi1_variability_enu_t variability = fmi1_import_get_variability(pVar);
         fmi1_base_type_enu_t type = fmi1_import_get_variable_base_type(pVar);
         fmi1_value_reference_t vr = fmi1_import_get_variable_vr(pVar);
+        QString description = fmi1_import_get_variable_description(pVar);
         name = toValidHopsanVarName(name);
         if(causality == fmi1_causality_enu_input && variability == fmi1_variability_enu_parameter)
         {
-            parNames.append(name);
-            parVars.append("m"+name);
-            parRefs.append(QString::number(vr));
+            hopsan_fmi1_import_variable_t par;
+            par.name = name;
+            par.variableName = "m"+name;
+            par.fmuVr = QString::number(vr);
+            par.dataType = type;
+            par.description = description;
+            par.pFmiVariable = pVar;
+            pars.append(par);
         }
         if(causality == fmi1_causality_enu_input && variability != fmi1_variability_enu_parameter)
         {
@@ -363,18 +380,24 @@ bool HopsanFMIGenerator::generateFromFmu1(const QString &rFmuPath, const QString
     QString classParent = "ComponentSignal";
 
     QString localVars;
-    if(!parVars.isEmpty())
+    if(!pars.isEmpty())
     {
-        localVars.append("double ");
-        foreach(const QString &varName, parVars)
+        for(const auto &par : pars)
         {
-            localVars.append(varName+", ");
+            if(par.dataType == fmi1_base_type_real) {
+                localVars.append("double ");
+            }
+            else if(par.dataType == fmi1_base_type_int) {
+                localVars.append("int ");
+            }
+            else if(par.dataType == fmi1_base_type_bool) {
+                localVars.append("bool ");
+            }
+            else if(par.dataType == fmi1_base_type_str) {
+                localVars.append("HString ");
+            }
+            localVars.append(par.variableName+";\n");
         }
-        if(localVars.endsWith(", "))
-        {
-            localVars.chop(2);
-        }
-        localVars.append(";\n");
     }
     if(!inputVars.isEmpty()  || !outputVars.isEmpty())
     {
@@ -395,16 +418,34 @@ bool HopsanFMIGenerator::generateFromFmu1(const QString &rFmuPath, const QString
     }
 
     QString addConstants;
-    for(int i=0; i<parNames.size(); ++i)
+    for(const auto &par : pars)
     {
-        if(i!=0)
-        {
-            addConstants.append("        ");
-        }
+        addConstants.append("        ");
 
-        fmi1_import_variable_t *pVar = fmi1_import_get_variable_by_name(fmu, parNames.at(i).toStdString().c_str());
-        double startValue = fmi1_import_get_real_variable_start(fmi1_import_get_variable_as_real(pVar));
-        addConstants.append("addConstant(\""+parNames.at(i)+"\", \"\", \"\", "+QString::number(startValue)+", "+parVars.at(i)+");\n");
+        QString startValueStr;
+        if(par.dataType == fmi1_base_type_real) {
+            double startValue = fmi1_import_get_real_variable_start(fmi1_import_get_variable_as_real(par.pFmiVariable));
+            startValueStr = QString::number(startValue);
+        }
+        else if(par.dataType == fmi1_base_type_int) {
+            int startValue = fmi1_import_get_integer_variable_start(fmi1_import_get_variable_as_integer(par.pFmiVariable));
+            startValueStr = QString::number(startValue);
+        }
+        else if(par.dataType == fmi1_base_type_bool) {
+            bool startValue = fmi1_import_get_boolean_variable_start(fmi1_import_get_variable_as_boolean(par.pFmiVariable));
+            if(startValue) {
+                startValueStr = "true";
+            }
+            else {
+                startValueStr == "false";
+            }
+        }
+        else if(par.dataType == fmi1_base_type_str) {
+            startValueStr = fmi1_import_get_string_variable_start(fmi1_import_get_variable_as_string(par.pFmiVariable));
+            startValueStr.append("\"");
+            startValueStr.prepend("\"");
+        }
+        addConstants.append("addConstant(\""+par.name+"\", \"\", \"\", "+startValueStr+", "+par.variableName+");\n");
     }
 
     QString addInputs;
@@ -429,11 +470,11 @@ bool HopsanFMIGenerator::generateFromFmu1(const QString &rFmuPath, const QString
 
     QString setPars;
     QString temp = extractTaggedSection(fmuComponentCode, "setpars");
-    for(int i=0; i<parNames.size(); ++i)
+    for(const auto &par : pars)
     {
         QString tempVar = temp;
-        tempVar.replace("<<<vr>>>", parRefs.at(i));
-        tempVar.replace("<<<var>>>", parVars.at(i));
+        tempVar.replace("<<<vr>>>", par.fmuVr);
+        tempVar.replace("<<<var>>>", par.fmuVr);
         setPars.append(tempVar+"\n");
     }
 
@@ -1647,7 +1688,7 @@ bool HopsanFMIGenerator::generateModelDescriptionXmlFile(ComponentSystem *pSyste
     QList<ParameterSpecification> parameterSpecs;
     getParameters(parameterSpecs, pSystem);
 
-    for(const auto &parSpec : parameterSpecs)
+    for(auto &parSpec : parameterSpecs)
     {
         QDomElement varElement = domDocument.createElement("ScalarVariable");
         varElement.setAttribute("name", parSpec.name);
--- a/HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
+++ b/HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
@@ -437,7 +437,7 @@ bool HopsanFMIGenerator::generateFromFmu1(const QString &rFmuPath, const QString
                 startValueStr = "true";
             }
             else {
-                startValueStr == "false";
+                startValueStr = "false";
             }
         }
         else if(par.dataType == fmi1_base_type_str) {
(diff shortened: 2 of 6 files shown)


peterNordin:
Did we not already have parameter support? Or am I confusing this with some other thing I worked on. Dont we have it for FMI2? Or was that only for import?

robbr48:
No, it was not implemented at all for export. We do support it for import though, and for S-function export.

peterNordin:
I remember now, it was for that other co-simulation project I did. I implemented to import capability on the Hopsan side, but only for FMI2. Not sure if it was later added for FMI1. I will compare to what I did and see if I have any additional ideas

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -199,14 +200,18 @@ void hopsan_set_real(int vr, double value)
         (*dataPtrs[vr]) = value;
     }
     else if(realParametersMap.count(vr)) {
-        spCoreComponentSystem->setParameterValue(realParametersMap[vr], hopsan::HString(std::to_string(value).c_str()));
+        std::ostringstream ss;
+        ss << value;
+        spCoreComponentSystem->setParameterValue(realParametersMap[vr], hopsan::HString(ss.str().c_str()));
```
peterNordin, line 205: There is a hopsan::to_hstring() function that mimics std::to_string, at least for the most common types

HopsanGenerator/src/GeneratorTypes.cpp
```diff
@@ -728,3 +729,30 @@ QString BuildFlags::platformString() const
...
+        else if(par->getType() == "integer") {
+            spec.type = "Integer";
+        }
+        else if(par->getType() == "bool") {
+            spec.type = "Boolean";
+        }
+        else if(par->getType() == "string") {
+            spec.type = "String";
+        }
+        else {
```
peterNordin, line 749: Should we support the other types as well,

Textblock and Conditional (the combobox type), Textblock could perhaps be treated as string and conditional as int

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -1310,6 +1310,31 @@ bool HopsanFMIGenerator::generateToFmu(QString savePath, ComponentSystem *pSyste
...
+        else if(parSpec.type == "Boolean") {
+            addParametersToMap.append(QString("boolParametersMap.insert(std::pair<int,std::string>(%1,\"%2\"));").arg(vr).arg(parSpec.name));
+        }
+        else if(parSpec.type == "String") {
+            addParametersToMap.append(QString("stringParametersMap.insert(std::pair<int,std::string>(%1,\"%2\"));").arg(vr).arg(parSpec.name));
+        }
+        else {
+            continue;   //Illegal data type, should never happen
+        }
+        QString temp = QString("dataPtrs[%1] = static_cast<%2*>(spCoreComponentSystem->getParameterDataPtr(\"%3\"));").arg(vr).arg(dataType).arg(parSpec.name);
```
peterNordin, line 1333: Has data type been set here ? Seems like it has not been assign

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -138,45 +149,88 @@ void hopsan_get_message(hopsan_message_callback_t message_callback, void* userSt
 
 double hopsan_get_real(int vr)
 {
-    return (*dataPtrs[vr]);
+    if(vr < sizeof(dataPtrs)/sizeof(double*)) {
+        return (*dataPtrs[vr]);
+    }
+    if(realParametersMap.count(vr)) {
```
peterNordin, line 155: Maybe it is better to add a static const variable with the max vr value, to avoid counting all of the time.
But maybe not such a  big deal since it only happens for parameters that are rarely accessed.

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -138,45 +149,88 @@ void hopsan_get_message(hopsan_message_callback_t message_callback, void* userSt
...
+        hopsan::HString value;
+        spCoreComponentSystem->getParameterValue(realParametersMap[vr],value);
+        bool ok;
+        return value.toDouble(&ok);
+    }
+    return 0;
```
peterNordin, line 161: Maybe use -1 for "error value" for doubles as well, (it is used for ints below) and 0 could be confused with a typical default value.

peterNordin:
I assume we can only test this with system parameters of type double  right now?

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -1310,6 +1310,31 @@ bool HopsanFMIGenerator::generateToFmu(QString savePath, ComponentSystem *pSyste
     fmuHopsanSourceCode.replace("<<<setdataptrs>>>", setDataPtrsString);
     fmuHopsanSourceCode.replace("<<<timestep>>>", QString::number(pSystem->getDesiredTimeStep()));
 
+    QList<ParameterSpecification> parSpecs;
+    getParameters(parSpecs, pSystem);
+    QString addParametersToMap;
+    for(const auto &parSpec : parSpecs) {
+        QString dataType;
+        if(parSpec.type == "Real") {
+            addParametersToMap.append(QString("realParametersMap.insert(std::pair<int,hopsan::HString>(%1,\"%2\"));").arg(vr).arg(parSpec.name.toStdString().c_str()));
```
peterNordin, line 1319: It would be nice with a  \n after these map insertions, to make the generated code more readbale, now all all ow them end up on the same very long line

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
peterNordin: Also I looked for this in the generated code, but it is not there at all ?So this entire line is completely discarded it seems like.

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -1618,6 +1643,29 @@ bool HopsanFMIGenerator::generateModelDescriptionXmlFile(ComponentSystem *pSyste
...
+    {
+        QDomElement varElement = domDocument.createElement("ScalarVariable");
+        varElement.setAttribute("name", parSpec.name);
+        varElement.setAttribute("valueReference", (unsigned int)vr);
+        varElement.setAttribute("causality", "parameter");
+        varElement.setAttribute("initial","exact");
+        varElement.setAttribute("variability", "fixed");
+        varElement.setAttribute("description", parSpec.description);
+        QDomElement dataElement = domDocument.createElement(parSpec.type);
+        dataElement.setAttribute("start", parSpec.init);   //! @todo Support start values
```
peterNordin, line 1659: I found the code I wrote, and I also appended the attribute "unit", I think that in this case it should be the "base unit" typically SI unit if a unit has been set for the parameter

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -25,9 +25,12 @@
 #include "fmu_hopsan.h"
 #include "HopsanCore.h"
 #include "HopsanEssentials.h"
+#include "HopsanTypes.h"
 #include <string>
 #include "model.hpp"
 #include <cassert>
+#include <iostream>
+#include <sstream>
```
peterNordin, line 33: I cant compile (Kubuntu 19.10), #include <map> is missing
IS iostream really needed ? (and sstream should not be needed if to_hstring() can be used (See other comment)

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: Good point, I forgot we had that one. Then we can get rid of <sstream> as well.

HopsanGenerator/src/GeneratorTypes.cpp
robbr48: Ok, I added them, but they are not tested yet. We can add some more advanced test cases later.

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
robbr48: It was not used, since parameters do not use data pointers. I removed the unused code.

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: It was even easier to just use a hard-coded value, since the code is auto generated anyway. It is actually a big deal, since this check is performed for every getReal/setReal call, also for input/output variables.

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: Fixed.

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
robbr48: Fixed this, and also some missing indentations.

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
robbr48: I added support for units, and confirmed that the units show up when importing the FMU back to Hopsan.

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: I fixed this as mentioned above.
Just for curiousity, why does this not work in Kubuntu? Is it a different compiler version?

HopsanGenerator/templates/fmu_hopsan.cpp
peterNordin: I meant the parametermap.count(), could add an other numparameters, but not so important.

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -138,45 +148,86 @@ void hopsan_get_message(hopsan_message_callback_t message_callback, void* userSt
...
 }
 
 const char* hopsan_get_string(int vr)
 {
-    //Write code here
+    if(stringParametersMap.count(vr)) {
+        hopsan::HString value;
+        spCoreComponentSystem->getParameterValue(stringParametersMap[vr],value);
+        bool ok;
+        return value.c_str();
```
peterNordin, line 191: I wonder how safe this is, since value is a temporary. The returned char * may be dangling. Maybe its OK if it is used emediately but I am not sure.

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -138,45 +148,86 @@ void hopsan_get_message(hopsan_message_callback_t message_callback, void* userSt
...
 }
 
 void hopsan_set_integer(int vr, int value)
 {
-    //Write code here
+    if(intParametersMap.count(vr)) {
+        std::ostringstream ss;
+        ss << value;
+        spCoreComponentSystem->setParameterValue(intParametersMap[vr], hopsan::HString(ss.str().c_str()));
+    }
```
peterNordin, line 212: Does not to_hstring work with int?

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -138,45 +148,86 @@ void hopsan_get_message(hopsan_message_callback_t message_callback, void* userSt
...
 }
 
 void hopsan_set_boolean(int vr, int value)
 {
-    //Write code here
+    if(realParametersMap.count(vr)) {
+        hopsan::HString hvalue = "true";
```
peterNordin, line 218: Wrong map used and checking hvalue istead of value in the next line

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -138,45 +148,86 @@ void hopsan_get_message(hopsan_message_callback_t message_callback, void* userSt
...
+        }
+        spCoreComponentSystem->setParameterValue(boolParametersMap[vr], hvalue);
+    }
 }
 
 void hopsan_set_string(int vr, const char* value)
 {
-    //Write code here
+    if(realParametersMap.count(vr)) {
+        spCoreComponentSystem->setParameterValue(boolParametersMap[vr], value);
```
peterNordin, line 229: Wrong map is used in both places in this function

HopsanGenerator/templates/fmu_hopsan.cpp
```diff
@@ -25,9 +25,11 @@
 #include "fmu_hopsan.h"
 #include "HopsanCore.h"
 #include "HopsanEssentials.h"
+#include "HopsanTypes.h"
 #include <string>
 #include "model.hpp"
 #include <cassert>
+#include "ComponentUtilities/num2string.hpp"
```
peterNordin, line 32: Map not included, I still can.build the exported fmu but this time there is some strange template error. Have not had time to.dive into what the problem is ( this time my laptop Ubuntu 18.04, GCC 7.5) could be some problem with the hstring type

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -1310,6 +1310,29 @@ bool HopsanFMIGenerator::generateToFmu(QString savePath, ComponentSystem *pSyste
...
+    QList<ParameterSpecification> parSpecs;
+    getParameters(parSpecs, pSystem);
+    QString addParametersToMap;
+    for(const auto &parSpec : parSpecs) {
+        QString dataType;
+        if(parSpec.type == "Real") {
+            addParametersToMap.append(QString("        realParametersMap.insert(std::pair<int,hopsan::HString>(%1,\"%2\"));\n").arg(vr).arg(parSpec.name.toStdString().c_str()));
+        }
+        else if(parSpec.type == "Integer") {
+            addParametersToMap.append(QString("        intParametersMap.insert(std::pair<int,std::string>(%1,\"%2\"));\n").arg(vr).arg(parSpec.name));
```
peterNordin, line 1322: I found the compilation error cause. Here and the other non real maps, uses std::string instead of hstring. But the maps are declared using hstring.  I tested to export the unit test model.

peterNordin: Could add one of each system parameter type to the fmi export test model

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: Now I see what you mean. But parameters are only set at the beginning of the simulation anyway, so it should not be performance critical. 

I was first going to use `contains()`, but I realized that it is not compatible with older compilers.

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
robbr48: Done (except conditional and text block, which are not supported as system parameters).

robbr48: And I also believe I fixed the compilation issues. I can't test it myself though.

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: Fixed.

HopsanGenerator/templates/fmu_hopsan.cpp
peterNordin: Yes likely due to different GCC version

HopsanGenerator/templates/fmu1CsImportComponentTemplate.hpp
```diff
@@ -180,6 +180,14 @@ class <<<className>>> : public <<<classParent>>>
             return;
         }
 
+        //Set parameters
```
peterNordin, line 183: This looks like an important bugfix, is it something that needs special mentioning in the release notes ?

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
peterNordin: Yes compilation issue fixed, regarding textblock and conditional, its just a GUI issue, you can actually have them if you edit the type in the hmf file manually. :) I forgot to fix that.

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -401,7 +401,10 @@ bool HopsanFMIGenerator::generateFromFmu1(const QString &rFmuPath, const QString
         {
             addConstants.append("        ");
         }
-        addConstants.append("addConstant(\""+parNames.at(i)+"\", \"\", \"\", 0, "+parVars.at(i)+");\n");
+
+        fmi1_import_variable_t *pVar = fmi1_import_get_variable_by_name(fmu, parNames.at(i).toStdString().c_str());
+        double startValue = fmi1_import_get_real_variable_start(fmi1_import_get_variable_as_real(pVar));
```
peterNordin, line 406: Import crash here, when import parameter is not of type double

HopsanGenerator/templates/fmu_hopsan.cpp
peterNordin: It should work with int, but the funny thing is that under the hood, it just calls stringstream just like this code, with the addition of setting a precision

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
robbr48: I made all parameter types supported for FMI1 import/export, and update the testmodel.

HopsanGenerator/templates/fmu1CsImportComponentTemplate.hpp
robbr48: We should mention that parameters are now supported for FMI import, both version 1 and 2. Or did you think of something else?

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: That should not be safe. I guess it works if you're lucky and the memory has not yet been overwritten, but we should fix it. We can't return string or hstring since the function must return a C compatible data type. Do you think a strcpy to a new const char buffer before returning will work?

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: It does work, so I replaced it.

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -429,11 +470,11 @@ bool HopsanFMIGenerator::generateFromFmu1(const QString &rFmuPath, const QString
...
     QString setPars;
     QString temp = extractTaggedSection(fmuComponentCode, "setpars");
-    for(int i=0; i<parNames.size(); ++i)
+    for(const auto &par : pars)
     {
         QString tempVar = temp;
-        tempVar.replace("<<<vr>>>", parRefs.at(i));
-        tempVar.replace("<<<var>>>", parVars.at(i));
+        tempVar.replace("<<<vr>>>", par.fmuVr);
+        tempVar.replace("<<<var>>>", par.fmuVr);
```
peterNordin, line 477: Should this really be par.fmuVr in both cases ?

HopsanGenerator/templates/fmu_hopsan.cpp
peterNordin: Yes this is the only way, to have an internal persistent variable to return from. This could be a HString that we assign and we just return HString.c_str(), but the problem is that the function will not become thread safe, but maybe there is no such guarantee anyway. 

So you can only call this function sequentially, and you must use the returned char* instantly and copy the value into a variable. I dont know if the FMI standard says anything about this.
Well actually you can use the returned pointer as long as our internal HString value is not being changed, but that will happen when the next call to the function is made.

But its the only solution I can see.

HopsanGenerator/templates/fmu_hopsan.cpp
peterNordin: I still think that < map >  should be explicitly included since std::map is used, so that this file becomes less dependent on what the other headers are including. 
Also the system includes should preferrably be moved last  (string and cassert), not that it matters in this case.

HopsanGenerator/templates/fmu1CsImportComponentTemplate.hpp
peterNordin: Yes that is what I meant, I will add it to the release notes

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: According to FMI specifications (2.0.1):
> The strings returned by fmi2GetString must be copied in the target environment because
the allocated memory for these strings might be deallocated by the next call to any of the
fmi2 interface functions or it might be an internal string buffer that is reused.

So if I understand correctly, the correct solution would be to make `value` a global variable to make sure it is persistent between each call to `hopsan_get_string`?

HopsanGenerator/templates/fmu_hopsan.cpp
robbr48: Ok, I added the include. We need to rearrange the order of includes in many files I think, but that can be done later.

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
robbr48: No, the second one should be the variable name. It is fixed now.

HopsanGenerator/templates/fmu_hopsan.cpp
peterNordin: Yes sounds correct

robbr48: Now it should work. I tested it with the unit test model, but it did of course work already before.

HopsanGenerator/templates/fmu1CsImportComponentTemplate.hpp
```diff
@@ -180,6 +180,14 @@ class <<<className>>> : public <<<classParent>>>
...
+        fmi1_value_reference_t vr;
+        double value;
+>>>setpars>>>        vr = <<<vr>>>;
```
peterNordin, line 186: This only work for real type parameters, text will cause compilation failure, and int/bool would just be cast to double

peterNordin: hmm apparently I could push directly to your branch :)

Issue #1906, opened 2020-03-30
Changeable units in bode plots in plot window


larsviktorlarsson:
The units on the x-axis should be able to change between Hz and rad/s
The unit on the Y-axis for the magnitude should be able to change between dB and Absolute value

[10:37:08] Info: HopsanCore 64-bit, Version: 2.13.0.20191126.1813
[10:37:08] Info: HopsanGUI 64-bit, Version: 2.13.0.20191126.1813

PR #1910, opened 2020-03-31
Create bode plots in rad/s, to match unit in plot window

### LINKED ISSUE #1905: Wrong frequency unit in Bode plots
When generating Bode plot in plot window data is in Hz but axis label says rad/s

robbr48:
Resolves #1905.

This will make bode plots always in rad/s. Does this make sense, or do we want it to default to Hz? What are your opinion @petkr14 @larsviktorlarsson?

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -5167,7 +5167,7 @@ void HcomHandler::executeBodeCommand(const QString cmd)
     int fMax = 500;
     if(args.size() > 2)
     {
-        fMax = args[2].toInt(); //!< @todo parse check needed
+        fMax = args[2].toInt()/(2*M_PI); //!< @todo parse check needed
     }
 
     PlotWindow *pWindow = gpPlotHandler->createNewPlotWindowOrGetCurrentOne("Bode plot");
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -1152,6 +1152,14 @@ void HcomHandler::createCommands()
     bodeCmd.group = "Plot Commands";
     mCmdList << bodeCmd;
 
+    HcomCommand nyquistCmd;
+    nyquistCmd.cmd = "nyquist";
+    nyquistCmd.description.append("Creates a Nyquist plot from specified curves");
+    nyquistCmd.help.append(" Usage: nyquist [invar] [outvar]");
+    nyquistCmd.fnc = &HcomHandler::executeNyquistCommand;
+    nyquistCmd.group = "Plot Commands";
+    mCmdList << nyquistCmd;
+
     HcomCommand optCmd;
     optCmd.cmd = "opt";
     optCmd.description.append("Initialize an optimization");
@@ -5174,6 +5182,33 @@ void HcomHandler::executeBodeCommand(const QString cmd)
     pWindow->closeAllTabs();
     pWindow->createBodePlot(pData1, pData2, fMax);
 }
+
+
+//! @brief Execute function for "nyquist" command
+void HcomHandler::executeNyquistCommand(const QString cmd)
+{
+    QStringList args = splitCommandArguments(cmd);
+    if(args.size() != 2)
+    {
+        HCOMERR("Wrong number of arguments.");
+        return;
+    }
+
+    QString var1 = args[0];
+    QString var2 = args[1];
+    SharedVectorVariableT pData1 = getLogVariable(var1);
+    SharedVectorVariableT pData2 = getLogVariable(var2);
+    if(!pData1 || !pData2)
+    {
+        HCOMERR("Data variable not found.");
+        return;
+    }
+
+    PlotWindow *pWindow = gpPlotHandler->createNewPlotWindowOrGetCurrentOne("Nyquist plot");
+    pWindow->closeAllTabs();
+    pWindow->createBodePlot(pData1, pData2, 0, false, true);
+}
+
 
 
 //! @brief Execute function for "opt" command
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -5172,7 +5172,7 @@ void HcomHandler::executeBodeCommand(const QString cmd)
         HCOMERR("Data variable not found.");
         return;
     }
-    int fMax = 500;
+    int fMax = 500*2*M_PI;
     if(args.size() > 2)
     {
         fMax = args[2].toInt()/(2*M_PI); //!< @todo parse check needed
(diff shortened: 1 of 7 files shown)


larsviktorlarsson:
> Resolves #1905.
> 
> This will make bode plots always in rad/s. Does this make sense, or do we want it to default to Hz? What are your opinion @petkr14 @larsviktorlarsson?

Ideally, I would like to be able to choose either (with right-click, as one can do in the standard plot window), see issue #1906 

But since the frequency range set in the dialogue for the frequency analysis tool is in Hz, the unit in the plot should be Hz as well. If we can change the range unit to rad/s as well, I vote for rad/s, since this is the SI unit and used as default for frequency in Hopsan (as far as I know?)

HopsanGUI/LogVariable.cpp
```diff
@@ -2134,6 +2134,7 @@ void createBodeVariables(const SharedVectorVariableT pInput, const SharedVectorV
 
     // Create the output variables for bode diagram
     SharedVectorVariableT pFrequencyVar = createFreeFrequencyVectorVariabel(freq);
+    pFrequencyVar->multData(M_PI*2);
```
peterNordin: Could have multiplied directly when creating freq, but maybe it does not matter. The createFreeFrequencyVariable function defaults to rad/s, so the assumtion us that the input data is in this format

peterNordin:
I rember having this discution a long time ago, regarding default unit for frquency. It also poped up about year ago in an other issue. There is a problem dealing with angular velocity and frequency Hz or rad/s since the units may get mixed up. But I do not know if that is a problem in this case.

Maybe the input dialog should change to rad/s or add the option to choose (and convert to rad/s behind the scenes). I think that it is best to return plot data in rad/s and fix whatever problem is preeventing conversion to Hz

peterNordin:
Also this code was written before we had a quantity system. And then frequency was whatever was suitable in different cases. Hz in this case. But the code was not updated when default was changed to rad/s

robbr48:
I changed so that rad/s is used also for the maximum limit in the dialog. Now everything is at least consistent. 

There is currently no support at all for changing units on the X-axis. This would be a useful feature in the future, not only for frequencies but also for changing time units to e.g. hours or milliseconds.  But right now I will leave it as it is.

While working on it, I also made it possible to create only Bode plots or only Nyquist plots, i.e. not having to create both all the time.

larsviktorlarsson:
> I changed so that rad/s is used also for the maximum limit in the dialog. Now everything is at least consistent.
> 
> There is currently no support at all for changing units on the X-axis. This would be a useful feature in the future, not only for frequencies but also for changing time units to e.g. hours or milliseconds. But right now I will leave it as it is.
> 
> While working on it, I also made it possible to create only Bode plots or only Nyquist plots, i.e. not having to create both all the time.

@robbr48 Awesome!



> I rember having this discution a long time ago, regarding default unit for frquency. It also poped up about year ago in an other issue. There is a problem dealing with angular velocity and frequency Hz or rad/s since the units may get mixed up. But I do not know if that is a problem in this case.
> 
> Maybe the input dialog should change to rad/s or add the option to choose (and convert to rad/s behind the scenes). I think that it is best to return plot data in rad/s and fix whatever problem is preeventing conversion to Hz

@peterNordin Hz vs rad/s is a nasty topic, and it will always depend on what is of interest. Hz we can usually relate to, and is most often given in datasheets, e.g. servo valves, but rad/s is more safe to use in calculations (it is most often also the resulting unit when we make calculations of e.g. resonance frequencies from the formula book). For sure I think the option of choosing between them is the way to go, but for the time being it's a really good improvement to make it consistent with the data. I guess this feature is not used very often, but now we plan to use it in a course, and the less misunderstandings the better..

HopsanGUI/HcomHandler.cpp
```diff
@@ -5167,7 +5175,7 @@ void HcomHandler::executeBodeCommand(const QString cmd)
     int fMax = 500;
     if(args.size() > 2)
     {
-        fMax = args[2].toInt(); //!< @todo parse check needed
+        fMax = args[2].toInt()/(2*M_PI); //!< @todo parse check needed
```
peterNordin, line 5178: Is the default value for fMax 500 reasonable now that its rad/s ?

HopsanGUI/PlotTab.cpp
```diff
@@ -1449,10 +1449,10 @@ void PlotTab::openCreateBodePlotDialog()
     {
         const double dataSize = pTimeVector->getDataSize()+1;
         const double stopTime = pTimeVector->last();
-        const double maxFreq = dataSize/stopTime/2;
+        const double maxFreq = dataSize/stopTime/2*M_PI*2;
```
peterNordin, line 1452: Here is multiplied by 2PI but at line 5178 in hcomhandler the value was divided by 2PI. I have not looked at the code in more detail only this diff, so maybe its supposed to be that way?
OK I think I understand, see different comment.

HopsanGUI/PlotWindow.cpp
```diff
@@ -962,72 +962,76 @@ void PlotWindow::createPlotWindowFromTab()
 
 
 //! @todo should not run code on non bodeplot tabs
-void PlotWindow::createBodePlot(SharedVectorVariableT var1, SharedVectorVariableT var2, int Fmax)
+void PlotWindow::createBodePlot(SharedVectorVariableT var1, SharedVectorVariableT var2, int Fmax, bool bode, bool nyquist)
```
peterNordin, line 965: OK, I think I understand. The Fmax parameter is in Hz. That is why there is multipliation and division with 2PI in various places. Could perhaps add a documentation block for this and the createNyquistPlot functions stating the intended unit.

HopsanGUI/PlotTab.cpp
```diff
@@ -1449,10 +1449,10 @@ void PlotTab::openCreateBodePlotDialog()
...
         QLabel *pMaxFrequencyLabel = new QLabel("Maximum frequency in bode plot:");
         QLabel *pMaxFrequencyValue = new QLabel();
-        QLabel *pMaxFrequencyUnit = new QLabel("Hz");
+        QLabel *pMaxFrequencyUnit = new QLabel("rad/s");
```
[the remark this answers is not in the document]
peterNordin: Definitely need a way to enter values manually
I tried this with the position servo example, cranked up the number of logsample (and lowered deltaT) and the default Fmax get really big, and pulling it down to a reasonable value becomes very difficult

petkr14:
Hej
Jag tycker inte att det är en stor fråga. Helst skulle man ju som Viktor säga vilja kunna växla men rad/s är ok. Ska man vara petig är dock standardenheten för frekvens Hertz medan rad/s är det för vinkelhastighet...jäkla 2 Pi....
/Petter

HopsanGUI/HcomHandler.cpp
robbr48: Solved by multiplying with 2pi.

HopsanGUI/PlotTab.cpp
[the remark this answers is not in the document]
robbr48: I added two connected spinboxes, one for Hz and one for rad/s. When user changes value in one of the boxes, the other box is automatically updated. The default value is the maximum frequency, as before.

PR #1911, opened 2020-04-01
Use angular velocity quantity instead of rad/s in components

### LINKED ISSUE #1907: Changeable units for angular velocity when used as input parameter in components
*I always get frustrated when I want to set the shaft speed of a pump (e.g. components "Q_type_Fixed_Displacement_Pump" and "Q_Type_Variable_Displacement_Pump") or an angular velocity source (e.g. components "Angle___Velocity_Source" and "Angular_velocity_source") and I have to put in a value in rad/s.*

*I would like to be able to set other units, such as rpm or rev/s or deg/s (i.e. same options as when you plot angular velocity).*

robbr48:
Resolves #1907.

Use "AngularVelocity"  and some other quantities instead of hard-coded default units.

I intentionally avoided changing generated components.

### DIFF
--- a/componentLibraries/defaultLibrary/Hydraulic/MachineParts/HydraulicValvePlate.hpp
+++ b/componentLibraries/defaultLibrary/Hydraulic/MachineParts/HydraulicValvePlate.hpp
@@ -83,7 +83,7 @@ namespace hopsan {
             addInputVariable("theta_1", "Angle 1", "deg", 6, &mpTh1);
             addInputVariable("theta_2", "Angle 2", "deg", 90, &mpTh2);
             addInputVariable("rho", "Oil Density", "kg/m^3", 890, &mpRho);
-            addInputVariable("movement", "Movement", "rad/s", 160, &mpMovement);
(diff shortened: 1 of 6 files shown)


peterNordin:
Yes "someone" should really update those generated ones.
Also doing changes like this is only safe if the hard-coded unit is the same as the base unit for the quantity. But in these cases it is, so no problem.

peterNordin:
When testing the bode plot PR, I found that the SignalSineWave function has the input variable f, set to Hz with description Frequency. But no Quantity. But this is typically such component where we cant just change the Quantity as then old models would result in the wrong behavior. 

Unless the unit Hz can be used to auto detect that a scaling is applied, not sure how that will work. Needs cleanup in the future when a foolproof auto-update method exists.

robbr48:
@peterNordin I think it should be possible to specify both quantity and unit, to use non-SI units as default. For the sine wave it could work with rad/s as default as well, but as you mention it would not be backwards compatible with existing models.

PR #1912, opened 2020-04-01
Open add component line edit by double-click

### LINKED ISSUE #1908: Add component directly on work pane should be double-click
*I always get frustrated when I am dragging components around and accidentally click on the work pane, because a text box appears in which I can search and add a component.*

*The search box should appear on double-click instead of single-click.*

robbr48:
Resolves #1908.

Change the quick add component feature so that it requires double-click instead of single-click. This makes it less likely to appear by accident.

### DIFF
--- a/HopsanGUI/GraphicsView.cpp
+++ b/HopsanGUI/GraphicsView.cpp
@@ -141,6 +141,46 @@ void GraphicsView::contextMenuEvent ( QContextMenuEvent * event )
     mIgnoreNextContextMenuEvent = false;
 }
 
+void GraphicsView::mouseDoubleClickEvent(QMouseEvent *event)
+{
+#if QT_VERSION >= 0x050000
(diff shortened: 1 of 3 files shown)


HopsanGUI/GraphicsView.cpp
```diff
@@ -141,6 +141,46 @@ void GraphicsView::contextMenuEvent ( QContextMenuEvent * event )
     mIgnoreNextContextMenuEvent = false;
 }
 
+void GraphicsView::mouseDoubleClickEvent(QMouseEvent *event)
```
peterNordin, line 144: This breaks entering subsystems by double-click

But it solves the problem of the box always popping up, I assume that some forwarding of the event, or calling parent double-click event, (I dont remember how this was done before), is needed.

peterNordin: Actually it breaks all double-clicking on things, like scopes or components for properties

robbr48:
Fixed by forwarding event to parent class.

PR #1914, opened 2020-04-01
Fix hcom self parameters

peterNordin:
Aside from making self.parName valid in HCOM it fixes som remaning issues with self, and adds built-in test capability to HopsanGUI

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -51,8 +51,7 @@ jobs:
 
     steps:
       - name: Install Qt
-        uses: ouuan/install-qt-action@v2.3.1
-        # jurplel/install-qt-action@v2
+        uses: jurplel/install-qt-action@v2
 
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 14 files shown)


peterNordin:
Yes I did not know what to do, since auto updating hcom scripts is difficult
So all I did was to enable the self.name variant, later on we can remove the other, that way you could say that "varname"  is a script variable but self.varname is the system parameter.

Maybe we should do that now, but then existing hcom script may not work.
Also I enabled the short version without .y var scalar parameters, so that it works the same way as in embedded scripts (numhop)

peterNordin:
Also a potential problem. If the --test option is used a GUI is not suposed to show up. But if the generator starts recompiling something that dialog will show up. Which would probably crash in non-gui environments such as CI. But I dont think that should happen.

robbr48:
@peterNordin I believe we should keep both ways for backward compatibility, at least for now.

peterNordin:
Apparently the built-in GUI tests do require a graphical x11 environment to start. 
So I removed them from CI for now, these tests must be run manually

Issue #1913, opened 2020-04-01
Tutorial/help for making frequency analysis in Hopsan


larsviktorlarsson:
**Is your improvement request related to a problem? Please describe.**
*When introducing frequency analysis in the uni courses, it will be difficult for the students to use the tool for the first time. When clicking on the question mark in the dialogue, the help opens but does not direct to the frequency analysis section in the user guide.*

**Describe the solution you'd like**
*The help button should direct to the frequency analysis section in the user guide. Also it would be nice if a tutorial for how to use the frequency analysis tool was available, so that it is easy to refer to. Maybe it could be included in the advanced tutorial https://hopsan.github.io/tutorials/tutorial_advanced_usage.pdf ?*

PR #1915, opened 2020-04-02
Correct user documentation for frequency analysis

robbr48:
Related to #1913.

- Use correct filename when opening frequency analysis documentation
- Mention possibility of creating only Bode or only Nyquist plots in documentation

We should also write a tutorial on frequency analysis, but this will have to wait for now.

### DIFF
--- a/doc/userFrequencyAnalysis.dox
+++ b/doc/userFrequencyAnalysis.dox
@@ -17,7 +17,7 @@ A dialog will appear, where input signal and output signal can be selected. It i
 - Use log samples that equal 2^n, e.g. 1024, 2048, 4096 and so on.
 - As input signal to the system, use a step with exponential delay or a pulse. Both input and output signal must begin and end at the same value.
 - Do not excite the system too much. Most systems are non-linear, and must be investigated around a working point.
-- A larger number of samples will improve results, especially if number of log samples are increased as well.
+- A larger number of log samples will improve the results.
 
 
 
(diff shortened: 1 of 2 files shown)


doc/userFrequencyAnalysis.dox
```diff
@@ -3,21 +3,21 @@
...
+A dialog will appear, where input signal and output signal can be selected. It is also possible to limit the frequency range. This is useful because numerical problems often occurs at high frequencies. You can also choose whether you want to create a Bode plot, a Nyquist plot or both. When clicking "Go", a new plot tab will appear for each generated plot.
 
 \subsection tfahints Hints
 - Use log samples that equal 2^n, e.g. 1024, 2048, 4096 and so on.
 - As input signal to the system, use a step with exponential delay or a pulse. Both input and output signal must begin and end at the same value.
 - Do not excite the system too much. Most systems are non-linear, and must be investigated around a working point.
-- Smaller time steps will improve results, especially if number of log samples are increased as well.
+- A larger number of samples will improve results, especially if number of log samples are increased as well.
```
peterNordin, line 20: This last sentence is a bit confusing, need to clarify the difference here between "samples" and "log samples" in this case I think. I assume that one is the simulation timestep and one refers to the number of logged samples from the simulation.

robbr48:
@peterNordin I removed the first part of the sentance. The important information is that many log samples will improve the results; number of simulated steps is irrelevant.

PR #1919, opened 2020-04-19
Issue 1906 change bode plot units

### LINKED ISSUE #1906: Changeable units in bode plots in plot window

peterNordin:
- Enable unit conversion for time and frequency (x-axis) in curves by PlotWindow drop down menu
- Change Bode plot underlying values to Gain (absolute) and Phase (rad)
- Apply default (dB and deg) unit conversion to plots

Resolves #1906

### DIFF
--- a/HopsanCore/src/Quantities.cpp
+++ b/HopsanCore/src/Quantities.cpp
@@ -63,6 +63,9 @@ hopsan::QuantityRegister::QuantityRegister()
 
     registerQuantity("Resistance", "ohm");
 
+    // This is a fake quantity for scaling between absolute and dB values
+    registerQuantity("Magnitude", "");
+
     // Register quantity aliases
     registerQuantityAlias("Position", "Length");
 }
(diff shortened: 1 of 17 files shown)


peterNordin:
> Seems to work fine. Plot markers are not updated when changing units, but that problem existed already before.

Was this for the amplitude and phase margin markers, or any ordinary marker?
Because I think that ordinary markers should update, when the unit changes.
However, they tend to move to some other place in the graph, but I will make a different issue for that

peterNordin:
This one requires #1910 to be merged first

robbr48:
> Was this for the amplitude and phase margin markers, or any ordinary marker?
> Because I think that ordinary markers should update, when the unit changes.
> However, they tend to move to some other place in the graph, but I will make a different issue for that

Yes, that's what I mean. The values are (probably) correct, but the location is not updated correctly.

PR #1920, opened 2020-04-21
Use makefiles for compiling exported FMUs

robbr48:
This enables parallel compilation, which speeds up compilation. It provides significant speed-up on Linux, and is at least notably faster on Windows. Unfortunately linking in MinGW is very slow and cannot be parallelized.

I had to use PowerShell scripts in Windows, since running `gcc.exe` from a batch script does not work for some reason. It should not be a problem though, since PowerShell is supported for Windows 7 and above.

### DIFF
--- a/HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
+++ b/HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
@@ -1730,9 +1730,13 @@ bool HopsanFMIGenerator::compileAndLinkFMU(const QString &fmuBuildPath, const QS
     QTextStream makefileStream(&makefile);
     makefileStream << "CXX = g++\n";
     makefileStream << "CC = gcc\n";
-    makefileStream << "CXXFLAGS = -fPIC -c -DHOPSAN_INTERNALDEFAULTCOMPONENTS -DHOPSAN_INTERNAL_EXTRACOMPONENTS -DHOPSANCORE_NOMULTITHREADING\n";
-    makefileStream << "CFLAGS = -fPIC -c\n";
-    makefileStream << "LFLAGS = -fPIC -w -shared -static-libgcc -Wl,--rpath,'$ORIGIN/.'\n";
+    QString fpicFlag;
+#ifndef _WIN32
+    fpicFlag= "-fPIC";
+#endif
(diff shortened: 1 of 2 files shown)


HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -1842,72 +1713,110 @@ bool HopsanFMIGenerator::compileAndLinkFMU(const QString &fmuBuildPath, const QS
...
 
-    if (!linkingOK) {
-        printErrorMessage("Failed to link exported FMU code.");
+    if (!compilationSuccessful) {
+        printErrorMessage("Failed to compile exported FMU C++ Hopsan code.");
         return false;
     }
 
+    QStringList objectFiles;
+    objectFiles << "fmu_hopsan.o";
```
peterNordin, line 1810: Are these used for anything ?

HopsanGenerator/src/generators/HopsanFMIGenerator.cpp
```diff
@@ -1842,72 +1713,110 @@ bool HopsanFMIGenerator::compileAndLinkFMU(const QString &fmuBuildPath, const QS
...
-    }
-    for(const QString linkLibrary : mLinkLibraries) {
-        linkBatchStream << " -l" << linkLibrary;
-    }
-    linkBatchStream << " -o \""+outputLibraryFile+"\"\n";
-    linkBatchFile.close();
+    QTextStream compileScriptStream(&compileScriptFile);
+#ifdef _WIN32
+    compileScriptStream << "$env:Path = \""+mCompilerSelection.path+";$env:Path\"\n";
+    compileScriptStream << "mingw32-make -j all\n";
```
peterNordin, line 1790: Maybe should add a limit to the number of threads, maybe 16 is reasonable nowdays.
Hopsan is limited in size, but if you are low on ram you may still run out, since it will build all .cpp files in parallel.
(Both for windows and linux)

robbr48: I didn't want to hard-core this limit, but perhaps it is needed for memory reasons. I read somewhere that the optimal value is roughtly 1.5 x number of processors to achieve good load balancing. But we can go with 16 jobs for now I guess.

PR #1921, opened 2020-04-22
Hann windowing before applying FFT

robbr48:
Always apply a [Hann window](https://en.wikipedia.org/wiki/Hann_function) before performing FFT. This includes both frequency spectrum and Bode plots. 

It is possible to add alternative window functions later, but this should be sufficient for almost all cases.

Any opinions @larsviktorlarsson?

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -1167,7 +1167,7 @@ void HcomHandler::createCommands()
     HcomCommand bodeCmd;
     bodeCmd.cmd = "bode";
     bodeCmd.description.append("Creates a bode plot from specified curves");
-    bodeCmd.help.append(" Usage: bode [invar] [outvar] [maxfreq]");
+    bodeCmd.help.append(" Usage: bode [invar] [outvar] [maxfreq] [windowfunction] [mintime] [maxtime]");
     bodeCmd.fnc = &HcomHandler::executeBodeCommand;
     bodeCmd.group = "Plot Commands";
     mCmdList << bodeCmd;
@@ -5177,7 +5177,7 @@ void HcomHandler::executeInheritTimestepCommand(const QString cmd)
 void HcomHandler::executeBodeCommand(const QString cmd)
 {
     QStringList args = splitCommandArguments(cmd);
-    if(args.size() < 2 || args.size() > 4)
+    if(args.size() < 2 || args.size() > 6)
     {
         HCOMERR("Wrong number of arguments.");
         return;
@@ -5198,9 +5198,44 @@ void HcomHandler::executeBodeCommand(const QString cmd)
         fMax = args[2].toInt()/(2*M_PI); //!< @todo parse check needed
     }
 
+    WindowingFunctionEnumT windowType = RectangularWindow;
+    if(args.size() > 3) {
+        if(args[3].toLower() == "hann") {
+            windowType = HannWindow;
+        }
+        else if(args[3].toLower() == "rectangular") {
+            windowType = RectangularWindow;
+        }
+        else {
+            HCOMERR("Unknown window function type: " + args[3]);
+            return;
+        }
+    }
+
+    if(args.size() == 5) {
+        HCOMERR("If minimum time is specified, maximum time is also required.");
+        return;
+    }
+
+    double minTime = -std::numeric_limits<double>::max();
+    double maxTime = std::numeric_limits<double>::max();
+    if(args.size() > 5) {
+        bool ok;
+        minTime = args[4].toDouble(&ok);
+        if(!ok) {
+            HCOMERR("Unknown minimum time value: " + args[4]);
+            return;
+        }
+        maxTime = args[5].toDouble(&ok);
+        if(!ok) {
+            HCOMERR("Unknown maximum time value: " + args[5]);
+            return;
+        }
+    }
+
     PlotWindow *pWindow = gpPlotHandler->createNewPlotWindowOrGetCurrentOne("Bode plot");
     pWindow->closeAllTabs();
-    pWindow->createBodePlot(pData1, pData2, fMax);
+    pWindow->createBodePlot(pData1, pData2, fMax, true, false, windowType, minTime, maxTime);
 }
 
 
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -6461,13 +6461,13 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
                 return;
             }
             bool ok;
-            minTime = minTimeArg.toDouble(&ok);
+            minTime = getNumber(minTimeArg, &ok);
             if(!ok) {
                 HCOMERR("Unknown time limit: "+minTimeArg);
                 mAnsType = Undefined;
                 return;
             }
-            maxTime = maxTimeArg.toDouble(&ok);
+            maxTime = getNumber(maxTimeArg, &ok);
             if(!ok) {
                 HCOMERR("Unknown time limit: "+maxTimeArg);
                 mAnsType = Undefined;
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -6476,7 +6476,7 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
         }
 
         if(powerArg.isEmpty()) {
-            mAnsVector = pDataVar->toFrequencySpectrum(pTimeVar, false);
+            mAnsVector = pDataVar->toFrequencySpectrum(pTimeVar, true);
         }
         else if(windowingFuncArg.isEmpty()) {
             mAnsVector = pDataVar->toFrequencySpectrum(pTimeVar, power);
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -6352,10 +6352,10 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
         }
         mAnsType = Undefined;
     }
-    //else if(desiredType != Scalar && expr.startsWith("fft(") && expr.endsWith(")"))
     else if(desiredType != Scalar && isHcomFunctionCall("fft", expr))
     {
-        QStringList args = expr.mid(4, expr.size()-5).split(",");
+        QString argStr = expr.mid(4, expr.size()-5);
+        QStringList args = SymHop::Expression::splitWithRespectToParentheses(argStr,',');
         for(int a=0; a<args.size(); ++a) {
             args[a] = args[a].trimmed();
         }
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -6320,107 +6320,139 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
     //else if(desiredType != Scalar && expr.startsWith("fft(") && expr.endsWith(")"))
     else if(desiredType != Scalar && isHcomFunctionCall("fft", expr))
     {
-        QString args = expr.mid(4, expr.size()-5);
-        QStringList splitArgs = SymHop::Expression::splitWithRespectToParentheses(args, ',');
-        if(splitArgs.size() == 1)
-        {
-            mAnsType = DataVector;
-            const QString varName = args.section(",",0,0).trimmed();
-            evaluateExpression(varName, DataVector);
-            SharedVectorVariableT pVar = mAnsVector;
-            if (mAnsType == DataVector)
-            {
-                mAnsType = DataVector;
-                mAnsVector = pVar->toFrequencySpectrum(pVar->getSharedTimeOrFrequencyVector(), false);
-                return;
-            }
-            else
-            {
-                HCOMERR(QString("Variable: %1 was not found!").arg(varName));
-                mAnsType = Undefined;
-                return;
-            }
+        QStringList args = expr.mid(4, expr.size()-5).split(",");
+        for(int a=0; a<args.size(); ++a) {
+            args[a] = args[a].trimmed();
         }
-        else if(splitArgs.size() == 2)
-        {
-            const QString varName = splitArgs[0].trimmed();
-            evaluateExpression(varName, DataVector);
-            SharedVectorVariableT pVar = mAnsVector;
-            if (mAnsType == DataVector)
-            {
-                bool power=false;
-                QString arg2 = splitArgs[1].trimmed();
-                SharedVectorVariableT pTimeVar;
-                if( (arg2=="true") || (arg2=="false") )
-                {
-                    power = (arg2 == "true");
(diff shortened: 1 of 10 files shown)


peterNordin:
Maybe it will be easier to test this once the other two PR related to Bode plots are merged

larsviktorlarsson:
> Always apply a [Hann window](https://en.wikipedia.org/wiki/Hann_function) before performing FFT. This includes both frequency spectrum and Bode plots.
> 
> It is possible to add alternative window functions later, but this should be sufficient for almost all cases.
> 
> Any opinions @larsviktorlarsson?

@robbr48 
 Hann window sounds good. I think it is important that the user can choose the size of the window (time limits). I think also that the window should be optional, i.e. the option of a rectangular window (ie no window function) should also be available, to enable the user to see the effect of the Hann window.

robbr48:
@larsviktorlarsson I added the options to choose window function type (rectangular or Hann), and also min and max time. This works both for frequency spectra and Bode/Nyquist plots and also from HCOM. I'm not sure of the reason for having min and max time limits in Bode plots, but perhaps there are use cases where it is needed.

larsviktorlarsson:
> @larsviktorlarsson I added the options to choose window function type (rectangular or Hann), and also min and max time. This works both for frequency spectra and Bode/Nyquist plots and also from HCOM. I'm not sure of the reason for having min and max time limits in Bode plots, but perhaps there are use cases where it is needed.

@robbr48 Sounds good. I think it is useful to have  min and max time on bode plots, because then the user can just pick a specific part of the data, and does not need to ensure that the simulated variable starts and stops at the same value (for instance in a long simulation with several events).

HopsanGUI/Utilities/GUIUtilities.cpp
```diff
@@ -473,6 +493,25 @@ void reduceVectorSize(QVector<double> &vector, int newSize)
...
+//! Both vectors must be of same size!
+//! @param x X-vector
+//! @param y Y-vector
+//! @param min Minimum x value
+//! @param max Maximum x value
+void limitVectorToRange(QVector<double> &x, QVector<double> &y, double min, double max)
+{
+    while(!x.isEmpty() && x.at(0) < min) {
+        x.remove(0);
+        y.remove(0);
```
peterNordin, line 506: Removing one by one looks inefficient, could use iterators and remove a range, but I am not sure if it matters.
Also min max could be called xmin xmax for clarity
Not super important

HopsanGUI/Utilities/GUIUtilities.cpp
```diff
@@ -374,6 +374,26 @@ QVector< complex<double> > realToComplex(const QVector<double> &rRealVector)
...
+//! @param data Vector with data to be windowed
+void windowFunction(QVector<double> &data, WindowingFunctionEnumT function)
+{
+    switch(function) {
+        case RectangularWindow:
+        {
+            break;  //Rectangular shall do nothing
+        }
+        case HannWindow: {
+            int N = data.size()-1;
```
peterNordin, line 388: should check that data is not empty

HopsanGUI/PlotTab.cpp
```diff
@@ -1468,7 +1533,47 @@ void PlotTab::openCreateBodePlotDialog()
...
         const double maxFreq = dataSize/stopTime/2;
-        QLabel *pMaxFrequencyLabel = new QLabel("Maximum frequency in bode plot:");
+
+        QGroupBox *pWindowingGroupBox = new QGroupBox("Windowing", pBodeDialog);
+        QGridLayout *pWindowingLayout = new QGridLayout(pWindowingGroupBox);
+
+        QLabel *pWindowingLabel = new QLabel("Windowing function:",pBodeDialog);
+        QComboBox *pWindowingComboBox = new QComboBox(pBodeDialog);
+        pWindowingComboBox->addItems(QStringList() << "Rectangular" << "Hann Function");
+        pWindowingComboBox->setCurrentIndex(1);
```
peterNordin, line 1543: Default should be rectangular like in hcom, and I realized its the same as do nothing

HopsanGUI/PlotTab.cpp
```diff
@@ -1366,12 +1380,51 @@ void PlotTab::openFrequencyAnalysisDialog(PlotCurve *pCurve)
...
+    mpWindowingMinTimeSpinBox->setDecimals(2-qFloor(log10(maxX-minX)));
+    mpWindowingMinTimeSpinBox->setSingleStep(pow(10,qFloor(log10(maxX-minX))-1));
+
+    QLabel *pMaxTimeLabel = new QLabel("Max time: ", pDialog);
+
+    mpWindowingMaxTimeSpinBox = new QDoubleSpinBox(this);
+    mpWindowingMaxTimeSpinBox->setMinimum(minX);
+    mpWindowingMaxTimeSpinBox->setMaximum(maxX);
+    mpWindowingMaxTimeSpinBox->setValue(maxX);
+    mpWindowingMaxTimeSpinBox->setDecimals(2-qFloor(log10(maxX-minX)));
```
peterNordin, line 1415: What would happen if maxX-minX == 0 ?
I am thinking about the log10() expression
Nut sure if that can happen

HopsanGUI/PlotTab.cpp
```diff
@@ -1366,12 +1380,51 @@ void PlotTab::openFrequencyAnalysisDialog(PlotCurve *pCurve)
...
+    QCheckBox *pPowerSpectrumCheckBox = new QCheckBox("Power spectrum", this);
     pPowerSpectrumCheckBox->setChecked(false);
 
+    QGroupBox *pWindowingGroupBox = new QGroupBox("Windowing", pDialog);
+    QGridLayout *pWindowingLayout = new QGridLayout(pWindowingGroupBox);
+
+    QLabel *pWindowingLabel = new QLabel("Windowing function:",pDialog);
+    QComboBox *pWindowingComboBox = new QComboBox(pDialog);
+    pWindowingComboBox->addItems(QStringList() << "Rectangular" << "Hann Function");
+    pWindowingComboBox->setCurrentIndex(1);
```
peterNordin, line 1395: Rectangular should be default ?

HopsanGUI/Utilities/GUIUtilities.cpp
robbr48: I improved it so it finds lowest and highest index of range and then uses a single call to remove for the entire length. Also renamed variables to xmin and xmax.

HopsanGUI/Utilities/GUIUtilities.cpp
robbr48: Actually it does not matter. If data is empty N will become negative, and then the for loop below will not be executed.

robbr48: I realized that there will indeed by a division by zero here if data size is 1, so that N becomes 0. I added a check to prevent this.

HopsanGUI/PlotTab.cpp
robbr48: It can happen if you use only one log sample in the simulation (unlikely), or if you make an X-Y plot with a constant variable as X (little less unlikely). I fixed it just be be on the safe side.

robbr48:
I added compensation factors for windowing, to maintain the energy content of the curve. There were also some weird error in the computation of the energy spectrum. It is now computed as the power spectral density multiplied by the total time, as it should.

HopsanGUI/Utilities/GUIUtilities.cpp
peterNordin: OK, I forgot that qt datatypes use int for size,

HopsanGUI/PlotTab.cpp
```diff
@@ -1397,6 +1397,11 @@ void PlotTab::openFrequencyAnalysisDialog(PlotCurve *pCurve)
     double minX = pCurve->getSharedTimeOrFrequencyVariable()->minOfData();
     double maxX = pCurve->getSharedTimeOrFrequencyVariable()->maxOfData();
 
+    if(minX == maxX) {
```
peterNordin, line 1400: Should not do == commparisson for doubles

HopsanGUI/PlotTab.cpp
```diff
@@ -1545,6 +1550,11 @@ void PlotTab::openCreateBodePlotDialog()
         double minX = pTimeVector->minOfData();
         double maxX = pTimeVector->maxOfData();
 
+        if(minX == maxX) {
```
peterNordin, line 1553: equal comparison for doubles

HopsanGUI/Utilities/GUIUtilities.cpp
```diff
@@ -377,18 +377,24 @@ QVector< complex<double> > realToComplex(const QVector<double> &rRealVector)
 //! @brief Apply windowing function
 //! Only Hann windows are supported for now
 //! @param data Vector with data to be windowed
-void windowFunction(QVector<double> &data, WindowingFunctionEnumT function)
+//! @param Ca Gain compensation factor
+//! @param Cb Frequency compensation factor
+void windowFunction(QVector<double> &data, WindowingFunctionEnumT function, double &Ca, double &Cb)
```
peterNordin, line 382: Not super important but if Ca and Cb are "return value parameters" they should be called rCa and rCb
Also doxygen @param should add [out]
@param [out] Ca

HopsanGUI/LogVariable.cpp
```diff
@@ -755,16 +756,16 @@ SharedVectorVariableT VectorVariable::toFrequencySpectrum(const SharedVectorVari
 
         // FFT is symmetric, so only use first half
         // Also skip f=0, but include n/2 (nyquist)
+        double fs = time.length()/maxt;     //Sampling frequency
```
peterNordin, line 759: .length() is not available in Qt4
Maybe we should drop Qt4 support for the next release, but lets keep it for this one

HopsanGUI/HcomHandler.cpp
```diff
@@ -6320,107 +6355,139 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
-            const QString varName = args.section(",",0,0).trimmed();
-            evaluateExpression(varName, DataVector);
-            SharedVectorVariableT pVar = mAnsVector;
-            if (mAnsType == DataVector)
-            {
-                mAnsType = DataVector;
-                mAnsVector = pVar->toFrequencySpectrum(pVar->getSharedTimeOrFrequencyVector(), false);
-                return;
+        QStringList args = expr.mid(4, expr.size()-5).split(",");
+        for(int a=0; a<args.size(); ++a) {
```
peterNordin, line 6359: Why not use "split args with respect to parentheses". What if the arguments are function taking other arguments,
I belive that is what that functions solved, or maybe that was an other one

HopsanGUI/HcomHandler.cpp
```diff
@@ -6320,107 +6355,139 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
                 return;
             }
         }
-        else
-        {
-            HCOMERR("Wrong number of arguments provided for fft function.\n"+mLocalFunctionDescriptions.find("fft").value().second);
-            mAnsType = Undefined;
-            return;
+
+        if(powerArg.isEmpty()) {
```
peterNordin, line 6478: windowing function and min max arguments can not be used if powerArg.isEmpty()
Not sure if window function is relevant. but min max time might be even in this case

robbr48: We could rearrange the number of arguments and put `power` last, but then you would have the oposite problem (i.e. need to specify windowing function if you want to use power option). I am also working on replacing the `fft()` function with two separate functions, one for energy and one for power spectrum.

HopsanGUI/PlotTab.cpp
robbr48: I'm not sure if it matters when the purpose of the check is to prevent division by zero, but I created an `xdiff` variable instead and compared it to zero.

HopsanGUI/PlotTab.cpp
robbr48: Same fix as above.

HopsanGUI/HcomHandler.cpp
robbr48: Good point. It is required for using input vector from another vector function, e.g. `fft(lp1(vector))`. I also added proper evaluation of min and max time using the `getNumber()` function.

robbr48:
While working on it, I also made power spectrum the default setting, both in HCOM and plot windows. I have the feeling it is more commonly used.

HopsanGUI/PlotTab.cpp
```diff
@@ -1397,28 +1397,29 @@ void PlotTab::openFrequencyAnalysisDialog(PlotCurve *pCurve)
     double minX = pCurve->getSharedTimeOrFrequencyVariable()->minOfData();
     double maxX = pCurve->getSharedTimeOrFrequencyVariable()->maxOfData();
 
-    if(minX == maxX) {
+    QLabel *pMinTimeLabel = new QLabel("Min time: ", pDialog);
+
+    double xdiff = maxX-minX;
+    if(0 == xdiff) {
```
peterNordin, line 1403: Hmmm, is this not the same thing as equals? maybe not exactly due to subtraction? but maybe equals comparison is the right thing in this case anyway.

HopsanGUI/PlotTab.cpp
peterNordin: yes seems OK in this case, I did not consider the purpose of the check

HopsanGUI/HcomHandler.cpp
peterNordin: I had forgotten about getNumber, It is likely needed in other places as well, (to do later)

HopsanGUI/HcomHandler.cpp
peterNordin: I did not look at all the code, only glanced it over so maybe I am missing something, but why cant the "windowingFunction, minTime, maxTime" arguments be included in the first case when power=true ?
And in the second case when windowfunction is not set, why cant we include min and max time ?  (just set windowing function to default value?

peterNordin:
> While working on it, I also made power spectrum the default setting, both in HCOM and plot windows. I have the feeling it is more commonly used.

I have no idea, I have forgotten all of this stuff, but your are probably right

HopsanGUI/HcomHandler.cpp
robbr48: There is only one function signature. If you want to use the last argument, you must provide all preceding arguments as well. It is possible to have multiple signatures, e.g. one with and one without the power argument, but it is quite complicated to distinguish between them. What if someone for example created a variable named "Hann", or a windowing function named "true"?

HopsanGUI/PlotTab.cpp
robbr48: Comparing double to zero should always be safe as far as I know. It will at least prevent division by zero, which is the purpose here.

HopsanGUI/HcomHandler.cpp
peterNordin: I was thinking that the missing defaults could be created here, but I realize now that this is in a hcom call, so then you must have given the windowing argument if min and max is given

PR #1929, opened 2020-04-23
Keep axis lock settings when browsing across empty generations

### LINKED ISSUE #1926: Lock axis setting is forgotten when browsing to an empty generation
**Describe the bug**
When browsing generations in plot window, the lock axis settings are forgotten and zoom reseted when passing an empty generation.

**To Reproduce**
1. Simulate a model three times
2. Remove generation 2
3. Plot something
4. Change zoom 
5. Lock one axis (or all axes)
6. Browse to generation 1
7. Lock axis boxes are unchecked and zoom restored to default

**Expected behavior**
Lock axis setting should be remembered and zoom kept

robbr48:
Resolves #1926.

Do not reset zoom and lock axis settings when removing last curve. Instead, reset it when adding first curve to an empty plot area.


peterNordin (approved):
Seems to work, except for the case where you navigate to a deleted generation and plot something new, then the lock will disappear. Better would be to still have the axis lock, but that gets very complicated.

PR #1938, opened 2020-05-05
Flat top windowing function

robbr48:
[Flat top windowing](https://en.wikipedia.org/wiki/Window_function#Flat_top_window) function. Implemented with the following parameters:
a0 = 1.0
a1 = 1.933
a2 = 1.286 
a3 = 0.388
a4 = 0.0322

### DIFF
--- a/HopsanGUI/Utilities/GUIUtilities.cpp
+++ b/HopsanGUI/Utilities/GUIUtilities.cpp
@@ -395,6 +395,21 @@ void windowFunction(QVector<double> &data, WindowingFunctionEnumT function, doub
             }
             rCa = 0.5;
             rCb = 1.5;
+            break;
+        }
+        case FlatTopWindow: {
+            int N = data.size()-1;
+            rCa = 0;
+            for(int n=0; n<=N && N>0; ++n) {
+                double w = 1.0 - 1.933*cos(2*M_PI*n/N) + 1.286*cos(4*M_PI*n/N) - 0.388*cos(6*M_PI*n/N) + 0.0322*cos(8*M_PI*n/N);
+                data[n] *= w;
+                rCa += w;
+                rCb += w*w;
+            }
+            rCa /= N;
+            rCb /= N;
+            rCb /= rCa*rCa;
+            break;
         }
     }
 
--- a/HopsanGUI/Utilities/GUIUtilities.cpp
+++ b/HopsanGUI/Utilities/GUIUtilities.cpp
@@ -400,8 +400,14 @@ void windowFunction(QVector<double> &data, WindowingFunctionEnumT function, doub
         case FlatTopWindow: {
             int N = data.size()-1;
             rCa = 0;
+            //Coefficients for flat top window according to ISO 18431-2
+            double a0 = 1.0;
+            double a1 = -1.933;
+            double a2 = 1.286;
+            double a3 = -0.388;
+            double a4 = 0.0322;
             for(int n=0; n<=N && N>0; ++n) {
-                double w = 1.0 - 1.933*cos(2*M_PI*n/N) + 1.286*cos(4*M_PI*n/N) - 0.388*cos(6*M_PI*n/N) + 0.0322*cos(8*M_PI*n/N);
+                double w = a0 + a1*cos(2*M_PI*n/N) + a2*cos(4*M_PI*n/N) + a3*cos(6*M_PI*n/N) + a4*cos(8*M_PI*n/N);
                 data[n] *= w;
                 rCa += w;
                 rCb += w*w;
--- a/HopsanGUI/Utilities/GUIUtilities.cpp
+++ b/HopsanGUI/Utilities/GUIUtilities.cpp
@@ -381,6 +381,12 @@ QVector< complex<double> > realToComplex(const QVector<double> &rRealVector)
 //! @param [out] Cb Frequency compensation factor
 void windowFunction(QVector<double> &data, WindowingFunctionEnumT function, double &rCa, double &rCb)
 {
+    if(data.size() <= 1) {
+        rCa = 1;
+        rCb = 1;
+        gpMessageHandler->addErrorMessage("Windowing functions require data of size greater than 1.");
+        return;
+    }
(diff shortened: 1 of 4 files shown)


HopsanGUI/Utilities/GUIUtilities.cpp
```diff
@@ -395,6 +395,21 @@ void windowFunction(QVector<double> &data, WindowingFunctionEnumT function, doub
...
+        case FlatTopWindow: {
+            int N = data.size()-1;
+            rCa = 0;
+            for(int n=0; n<=N && N>0; ++n) {
+                double w = 1.0 - 1.933*cos(2*M_PI*n/N) + 1.286*cos(4*M_PI*n/N) - 0.388*cos(6*M_PI*n/N) + 0.0322*cos(8*M_PI*n/N);
+                data[n] *= w;
+                rCa += w;
+                rCb += w*w;
+            }
+            rCa /= N;
```
peterNordin, line 409: Maybe need protection for division by zero in rCa rCb and rC?, the last one should it be b or c?

HopsanGenerator/include/generators/HopsanModelicaGenerator.h
```diff
@@ -36,8 +36,8 @@ class HopsanModelicaGenerator : public HopsanGeneratorBase
...
+    void generateComponentObject(ComponentSpecification &comp, QString &typeName, QString &displayName, QString &cqsType, /*QStringList &initAlgorithms,*/ QStringList &preAlgorithms, QStringList &equations, QStringList &finalAlgorithms, QList<PortSpecification> &portList, QList<ParameterSpecification> &parametersList, QList<VariableSpecification> &variables, QTextStream &logStream, bool &success);
+    void generateComponentObjectNumericalIntegration(ComponentSpecification &comp, QString &typeName, QString &displayName, QString &cqsType, QStringList &initAlgorithms, QStringList &preAlgorithms, QStringList &equations, QStringList &finalAlgorithms, QList<PortSpecification> &portList, QList<ParameterSpecification> &parametersList, QList<VariableSpecification> &variables, QTextStream &logStream, bool &success);
```
peterNordin, line 40: These bool &success are only added in the declaration but not in the definition. Leftovers ?

HopsanGenerator/src/generators/HopsanModelicaGenerator.cpp
```diff
@@ -200,15 +200,15 @@ void HopsanModelicaGenerator::parseModelicaModel(QString code, QString &typeName
...
+                QString tempLine = line;
                 tempLine.remove(" ");
                 int idx = tempLine.indexOf("hopsanCqsType=");
                 cqsType = tempLine.at(idx+15);
             }
             else if(words.at(0) == "parameter")         //"parameter" keyword
             {
                 QString name = words.at(2).section("(",0,0);
-                QString unit = lines.at(l).section("unit=",1,1).section("\"",1,1);
+                QString unit = line.section("unit=",1,1).section("\"",1,1);
```
peterNordin, line 211: I do not know modelica, but what about requirements on = does it have to be attached directly to the name (unit) or would it be valid modelica code with a space, like in 'unit ='

peterNordin:
> [Flat top windowing](https://en.wikipedia.org/wiki/Window_function#Flat_top_window) function. Implemented with the following parameters:
> a0 = 1.0
> a1 = 1.933
> a2 = 1.286
> a3 = 0.388
> a4 = 0.0322

What make these parameters special? Are they always suitable ?

HopsanGUI/Utilities/GUIUtilities.cpp
robbr48: I added a check for empty data sets at the top of the function.

I believe the computations are correct. The formulas are:
<a href="https://www.codecogs.com/eqnedit.php?latex=C_a&space;=&space;\textup{mean}(\omega)&space;\\&space;C_b&space;=&space;\frac{\textup{mean}(\omega^2)}{\textup{mean}(\omega)^2}" target="_blank"></a>

HopsanGenerator/src/generators/HopsanModelicaGenerator.cpp
robbr48: Sorry, this was not supposed to be in this PR. I accidentally included a commit from another branch. But to answer the question, I don't know. All examples I have seen have the equality symbol attached to the name, but I guess spaces are allowed too. I will fix this in the other branch.

robbr48:
The coefficients are from the ISO 18431-2 standard. There are other possible coefficients, Matlab for examples uses a different set. Ideally we could allow users to specify the coefficients, but that would require a lot of additional widgets and HCOM function arguments. It could even be put in the configuration XML file and accessed from the options dialog. But for now, I decided to go with he ISO standard.

HopsanGUI/Utilities/GUIUtilities.cpp
```diff
@@ -381,6 +381,12 @@ QVector< complex<double> > realToComplex(const QVector<double> &rRealVector)
 //! @param [out] Cb Frequency compensation factor
 void windowFunction(QVector<double> &data, WindowingFunctionEnumT function, double &rCa, double &rCb)
 {
+    if(data.empty()) {
```
peterNordin, line 384: Inst the problem here if data has size  1

HopsanGUI/Utilities/GUIUtilities.cpp
peterNordin: nevermind, I did not look closely enough, I though there were three out parameters rCa, rCb and rCc (but the last does note exist)

HopsanGenerator/src/generators/HopsanModelicaGenerator.cpp
peterNordin: OK

HopsanGUI/Utilities/GUIUtilities.cpp
robbr48: You are right. I changed the check so that data size must be greater than 1.

PR #1939, opened 2020-05-05
RMS frequency spectrum type

robbr48:
Computed as G_rms = sqrt(2)*abs(FFT(data))/length(data).

Implementation needs to be verified before the PR can be merged.


robbr48:
This works according with the original user request, but I have not found any reliable independent source for confirmation. Let's merge it it now and fix it later if someone discovers a problem.

PR #1943, opened 2020-05-13
Hopsan module for Julia

robbr48:
A Hopsan module for the Julia language, providing a simple API:

- [x] Loading models
- [x] Loading component libraries
- [x] Setting start time, stop time and step size
- [x] Initializing
- [x] Simulating
- [x] Fetching log data
- [x] Setting parameters

This uses a new wrapper library with a plain C interface, which can later be reused for other languages, e.g. Python or Matlab.

In the future we could support creating and modifying models, but that can be handled in a separate pull request.


hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,275 @@
...
+{
+    if(!spCoreComponentSystem) {
+        std::cout << "Error: No model is loaded.\n";
+        size = -1;
+        return nullptr;
+    }
+
+    //Parse variable string
+    hopsan::HString varStr(variable);
+    hopsan::HVector<hopsan::HString> splitSys = varStr.split('|');
```
peterNordin, line 63: We need to normalize the "syntax" for getting components,  sometimes its $ and # sometimes its | and .
Maybe not in this PR, but it would be nice to have it correct from the start.

I think that only using . as separator should be possible, just assume that each name is a component name, until its not, then check for port. Ofcourse then there is the problem if one component has the same name as a port

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+    int varId = pPort->getNodeDataIdFromName(splitVar[2]);
+    if(varId < 0) {
+        std::cout << "Error: No such variable: " << splitVar[2].c_str() << "\n";
+        size = -1;
+        return nullptr;
+    }
+
+    size = int(spCoreComponentSystem->getNumActuallyLoggedSamples());
+    double *data = new double[size];
```
peterNordin, line 107: Returning this array will lead to a memory leak since it is never deleted, also it can not be deleted in the caller (at least on Windows) since that is in a different heap, which will lead to memory corruption.

One alternative would be to return the array to this library for deletion. 
But I think that it would be better if the caller allocates the data buffer and sends in a pointer to where the data should be written. Of course the caller must then fist know the length of the data, but a function call for that information can be implemented.

Then the caller calls this function with a pointer to a double array, and the maximum allowed size.

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,275 @@
...
+    if (spCoreComponentSystem->checkModelBeforeSimulation()) {
+        std::cout << "Success!\n";
+    }
+    else {
+        std::cout << "Failed!\n";
+        printWaitingMessages(gHopsanCore, false, false);
+        return -1;
+    }
+
+    std::cout << "Initializing model... ";
```
peterNordin, line 164: I think that we have a preInitialize also, that must run

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,275 @@
...
+        std::cout << "Failed!\n";
+        printWaitingMessages(gHopsanCore, false, false);
+        return -1;
+    }
+    printWaitingMessages(gHopsanCore, false, false);
+
+    std::cout << "Simulating... ";
+    spCoreComponentSystem->simulate(stopTime);
+    std::cout << "Finished!\n";
+
```
peterNordin, line 178: Need to call finilize

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,275 @@
...
+double *getTimeVector(int &size)
+{
+    if(!spCoreComponentSystem) {
+        std::cout << "Error: No model is loaded.\n";
+        size = -1;
+        return nullptr;
+    }
+    size = int(spCoreComponentSystem->getNumActuallyLoggedSamples());
+    return spCoreComponentSystem->getLogTimeVector()->data();
```
peterNordin, line 191: In this case, returning a pointer to data is OK, since data lifetime is managed inside hopsan core.
But It might be a good idea to let the caller send in a pointer to buffer and size and copy the data, so that the caller owns the allocated memory, incase the caller would decide to store the pointer for later use. 

Or we write in the function documentation that the data must be consumed immediately, because it is no guarantee that the memory will remain valid. Then the caller can clone the data themselves.

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,275 @@
...
+            std::cout << "Error: Failed to set parameter value: " << parName.c_str() << "\n";
+            return -1;
+        }
+    }
+
+    std::cout << "Error: Wrong number of arguments.\n";
+    return -1;
+}
+
+int setLogSamples(unsigned long value)
```
peterNordin, line 266: Maybe call this, setNumLogsamples or similar, or maybe no needed

hopsanc/include/hopsanc.h
```diff
@@ -0,0 +1,20 @@
...
+extern "C" {
+    HOPSANC_DLLAPI int loadLibrary(const char* path);
+    HOPSANC_DLLAPI int loadModel(const char* path);
+    HOPSANC_DLLAPI int setParameter(const char* name, const char *value);
+    HOPSANC_DLLAPI int setStartTime(double value);
+    HOPSANC_DLLAPI int setTimeStep(double value);
+    HOPSANC_DLLAPI int setStopTime(double value);
+    HOPSANC_DLLAPI int setLogSamples(unsigned long int value);
+    HOPSANC_DLLAPI int simulate();
+    HOPSANC_DLLAPI double *getTimeVector(int &value);
```
peterNordin, line 15: C does not have references, so this is in fact a C++ API with C-style name mangling
But change the references to pointers, then its valid C

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,275 @@
...
+        }
+        else {
+            std::cout << msg.c_str() << "\n";
+        }
+    }
+}
+
+
+int loadModel(const char* path) {
+    spCoreComponentSystem = gHopsanCore.loadHMFModelFile(path, startTime, stopTime);
```
peterNordin, line 35: If a model has been previously loaded, it should be deleted before loading a new one

hopsanc/hopsanc.pro
```diff
@@ -0,0 +1,54 @@
...
+     # Add runtime search path so that dynamically loaded libraries in the same directory can be found.
+    # Note! QMAKE_LFLAGS_RPATH and QMAKE_RPATHDIR does not seem hande $$ORIGIN, adding manually to LFLAGS
+    QMAKE_LFLAGS *= -Wl,-rpath,\'\$$ORIGIN/./\'
+}
+
+# -------------------------------------------------
+# Project files
+# -------------------------------------------------
+HEADERS = \
+    include/hopsanc.h \
```
peterNordin, line 50: I think that header files should be "namespaced" with an additional sub-directory with he name of this library, to avoid name collision. For instance hopsanc_win32dll.h could then still be called win32dll.h
I know it feels silly when there  are so few files, but you avoid trouble in the future by always namespacing include files.

We should do this for new libraries

hopsanc/hopsanc.pro
```diff
@@ -0,0 +1,54 @@
+# -------------------------------------------------
+# Global project options
```
peterNordin, line 2: I would also like a CMake file, please :)
Or I can make one if you want

Scripts/Julia/hopsan.j1
```diff
@@ -0,0 +1,53 @@
+module hopsan
+  const hopsanlibpath = joinpath(ENV["HOPSANHOME"], "bin/libhopsanc")
+  const complibpath = joinpath(ENV["HOPSANHOME"], "componentLibraries/defaultLibrary/libdefaultcomponentlibrary.so")
```
peterNordin, line 3: Hard-coded .so name for library
On Windows I think the components are still built-in though, so no need to look for DLL.
But I am not sure. But on windows then we should not try to load at all.

HopsanNG.pro
```diff
@@ -1,4 +1,4 @@
 TEMPLATE = subdirs
 
 CONFIG += ordered
-SUBDIRS = HopsanCore componentLibraries SymHop Ops HopsanGenerator hopsangeneratorgui hopsanremote HopsanGUI HopsanCLI HoLC UnitTests
+SUBDIRS = HopsanCore componentLibraries SymHop Ops HopsanGenerator hopsangeneratorgui hopsanremote HopsanGUI HopsanCLI HoLC UnitTests hopsanc
```
peterNordin, line 4: I am not sure if it would work, but it would be nice if the C-API exists in the core library directly, to avoid a wrapper DLL.
But on the other hand, its strange if the HopsanCore library contains an instance of it self, so a wrapper DLL might be a good idea anyway.

hopsanc/src/hopsanc.cpp
robbr48: I changed it according to your second suggestion. It actually made the Julia code much nicer as well :)

hopsanc/src/hopsanc.cpp
robbr48: It is automatically called from within `initialize`, so we can skip it here.

hopsanc/src/hopsanc.cpp
robbr48: I modified this function so it works the same way as for `getDataVector`; the array must be pre-buffed by the caller. I skipped the size argument, since it should always equal number of samples.

hopsanc/src/hopsanc.cpp
robbr48: In that case I would call it `setNumberOfLogSamples`, but I believe the name is clear enough anyway.

hopsanc/include/hopsanc.h
robbr48: Already fixed, since the size argument was removed.

hopsanc/hopsanc.pro
robbr48: I created it, but I am no CMake export so you might want to take a look at it.

Scripts/Julia/hopsan.j1
robbr48: I added platform specific paths. It could be as you say, that we don't need to load the default library at all in Windows. I can try it, but I need to install Julia first.

HopsanNG.pro
robbr48: I prefer a separate wrapper. I think it is cleaner and easier to maintain.

I think we have discussed it before, but a future project could be to only have a C API for HopsanCore. That would require completely re-writing the interface between HopsanGUI and HopsanCore though.

hopsanc/hopsanc.pro
peterNordin: looks good but did not test, you also need to include in in the top-level cmake file, "add_subdirectory(hopsanc)
"
Also I noticed it is added as a static library, but in the qt .pro file it is a shared library. Do know what is best in this case, but I guess it should be shared, so that Julia can load it at runtime ?

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,282 @@
...
+    spCoreComponentSystem->addSearchPath(modelName+"-resources");
+    std::cout << "Loaded model: " << modelName.c_str() << "\n";
+    printWaitingMessages(gHopsanCore, false, false);
+    return 0;
+}
+
+
+//! @brief Returns specified data vector as C array
+//! @param [in] variable Variable name ("component.port.variable")
+//! @param [out] size Size of returned data array
```
peterNordin, line 55: update documentation string

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,282 @@
...
+    if(!spCoreComponentSystem) {
+        std::cout << "Error: No model is loaded.\n";
+        return -1;
+    }
+    std::cout << "Setting samples to " << size_t(value) << "\n";
+    spCoreComponentSystem->setNumLogSamples(value);
+    return 0;
+}
+
+size_t getLogSamples()
```
peterNordin, line 275: size_t is not a C type I think, (maybe i newer C standards)

hopsanc/src/hopsanc.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    std::cout << "Finalizing... ";
+    spCoreComponentSystem->finalize();
+    std::cout << "Finished!\n";
+
+    printWaitingMessages(gHopsanCore, false, false);
+    return 0;
+}
+
+int getTimeVector(double *data)
```
peterNordin, line 182: also here, maybe should have a size input argument, to avoid buffer overflow in the general case ?

hopsanc/src/hopsanc.cpp
robbr48: I thought about using a size argument, but it felt weird to have an argument that is only allowed to have a specified value. I think it is better to clearly define the requirements on the buffer size in the documentation.

hopsanc/src/hopsanc.cpp
robbr48: It was defined in C99, so I think it is safe to use it. But I will change both functions so they have the same type.

hopsanc/src/hopsanc.cpp
robbr48: I added it to the function documentation instead, like above. But if this is super important to you I can add the extra argument instead.

hopsanc/include/hopsanc.h
```diff
@@ -0,0 +1,41 @@
...
+#endif
+
+#include <stddef.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+HOPSANC_DLLAPI int loadLibrary(const char* path);
+    HOPSANC_DLLAPI int getMessage(char* buf, size_t bufSize);
```
peterNordin, line 24: Something strange with the indentation

hopsanc/src/hopsanc.cpp
peterNordin: From a security angle, if the size is known bufferoverflow can be avoided, but I guess we could ignore or missuse the size anyway, or the user could give the wrong size and buffer overflow would happen anyway.
So I am not sure if it matters really. I guess we can keep it like it is, because the the user must preallocate enough space anyway, and the getSize function tells the user how large the data will be.

PR #1944, opened 2020-05-14
Remove PythonQt and related code

peterNordin:
The built-in python scripting in HopsanGUI has not seen active work in several years, and is only available for select Linux distributions. Keeping it causes to much build related problems, so we remove it completely instead. HCOM is the command language to use inside HopsanGUI.

Python can still be used externally to call HopsanCLI or the soon to be added C-wrapper library.


peterNordin:
@robbr48 This removes Model related scripts, that were to be called as init scripts.
But I do not think that is needed anymore since we have embedded scripts for that (numhop).
But those are not so advanced, so maybe we should keep that functionality but with HCOM scripts.
On the other hand its a bit scarry to auto-run externals scripts, also the function never really worked I think.

robbr48:
Embedding Python scripts in models was a bad idea to begin with. Models should not execute scripts, scripts should execute models. NumHop is fine since it only does numerical manipulations, but Python/HCOM scripts could start simulations, create plots and all sorts of other stuff which should not be embedded in a model.

PR #1946, opened 2020-05-15
Add new component "AeroFlightRecorder" to the "AeroComponents" library.

alejandroSobron:
The new "AeroFlightRecorder" component generates .kml output files for flight or ground trajectory visualization in Google Earth. Tested in different models and scenarios, it should behave well. Documentation is included in AeroComponents/AeroFlightRecorder.md

### DIFF
--- a/componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.dae
+++ b/componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.dae
@@ -1,0 +1,761 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
+<COLLADA xmlns="http://www.collada.org/2005/11/COLLADASchema" version="1.4.1">
+    <asset>
+        <contributor>
+            <authoring_tool>Google SketchUp 7.1.4871</authoring_tool>
+        </contributor>
+        <created>2009-10-17T10:23:04Z</created>
+        <modified>2009-10-17T10:23:04Z</modified>
+        <unit meter="0.0254" name="inch" />
+        <up_axis>Z_UP</up_axis>
+    </asset>
+    <library_visual_scenes>
+        <visual_scene id="ID1">
+            <node name="SketchUp">
+                <node id="ID2" name="instance_0">
+                    <matrix>0 1 0 0.851428 -1 0 0 0.8320829 0 -0 1 11.02362 0 0 0 1</matrix>
+                    <instance_node url="#ID3" />
+                </node>
+            </node>
+        </visual_scene>
+    </library_visual_scenes>
+    <library_nodes>
+        <node id="ID3" name="component_0">
+            <instance_geometry url="#ID4">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID5">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID12">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID5">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID18">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID20">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID26">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID27">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID34">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID35">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID42">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID27">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID48">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID20">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID54">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID27">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID60">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID20">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID66">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID35">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID72">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID73">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID80">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID35">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID86">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID35">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID92">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID73">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID98">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID73">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+            <instance_geometry url="#ID104">
+                <bind_material>
+                    <technique_common>
+                        <instance_material symbol="Material2" target="#ID73">
+                            <bind_vertex_input semantic="UVSET0" input_semantic="TEXCOORD" input_set="0" />
+                        </instance_material>
+                    </technique_common>
+                </bind_material>
+            </instance_geometry>
+        </node>
+    </library_nodes>
+    <library_geometries>
+        <geometry id="ID4">
+            <mesh>
+                <source id="ID7">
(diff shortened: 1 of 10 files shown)


peterNordin:
I have not had time to look at this in detail yet, but can you cleanup the indention of the hpp file, right now it mixes spaces and tabs, use spaces. notepad++ or similar programs have functions for this, or just search replace tabs -> "four spaces" (or two is maybe better).

Also a test model would be nice, You can put it under "Models / Component Test".

Nice to see markdown documentation :)

alejandroSobron:
Absolutely! Sorry, I didn't notice the mix. I'll fix that ;)

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
```diff
@@ -0,0 +1,329 @@
...
+
+        void configure()
+        {
+            //Add ports to the component
+
+            //Add inputVariables to the component
+            addInputVariable("longitude","Longitude E","deg",15.577,&mpLongitude);
+            addInputVariable("latitude","Latitude N","deg",58.401,&mpLatitude);
+            addInputVariable("altitude","Altitude","m",0,&mpAltitude);
+            addInputVariable("phi","Roll","rad",0,&mpPhi);
```
peterNordin, line 109: Switch unit "rad" to quantity name "Angle", that way it will be possible to choose unit in graphical interface
Default unit for Angle is rad

Switch unit "m/s" to "Velocity", default unit for Velocity is m/s

Switch unit m to  Position, but that seems silly in this case, you can add an alias Altitude in Quantities.cpp line 71
Maybe its not important but to make things a bit more complicated what Altitude is this ? 
There are many different types of altitude, over ground, over ellipsoid, over geoide, pressure based altitude. 
I do not have a good idea, maybe its easiest to just call it Altitude (something that is measured in meters or feet or similar). Then it will just be an alias for Position, and the altitude reference could maybe go into the description.

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
```diff
@@ -0,0 +1,329 @@
...
+        {
+            return new AeroFlightRecorder();
+        }
+
+        void configure()
+        {
+            //Add ports to the component
+
+            //Add inputVariables to the component
+            addInputVariable("longitude","Longitude E","deg",15.577,&mpLongitude);
```
peterNordin, line 106: I realize lat,lon is easiest to work with in deg, but it would maybe better to use the Quantity Angle. But right now it is not possible to choose the display unit for the quantity to use by default so it would then show up in rad in the GUI. But it would be convertible to deg.
I am not sure what to do here right now, because if we go with deg, we wont be able to change it later as "old models" would then have the values expressed in degrees instead of the Angle base unit.

What do you think @robbr48, maybe we should use Angle here,  and later add default unit selection to the caf xml file, so that the component creator can choose the default scaling to be visible. 

If you change to Angle, then you should write deg2rad(15.577) and deg2rad(58.401) in this initialization for readability.
Those functions are likely available through the ComponentUtilities.h header (actually AuxiliarySimulationFunctions.h)

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
```diff
@@ -0,0 +1,329 @@
...
+            addInputVariable("latitude","Latitude N","deg",58.401,&mpLatitude);
+            addInputVariable("altitude","Altitude","m",0,&mpAltitude);
+            addInputVariable("phi","Roll","rad",0,&mpPhi);
+            addInputVariable("theta","Pitch","rad",0,&mpTheta);
+            addInputVariable("psi","Heading","rad",0,&mpPsi);
+            addInputVariable("v","Speed","m/s",0,&mpV);
+
+            // Add inputParameters
+            addInputVariable("dt", "Logging interval (0 = log every sample)", "s", 1, &mpDt);
```
peterNordin, line 115: Instead of s for seconds use Quantity Time
Scaling below, is scale really m, should it not be without dimension, (leave blank)

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
```diff
@@ -0,0 +1,329 @@
...
+                stopSimulation("Please specify a path for the output file.");
+                return;
+            }
+
+            //Fetch parent system name (hmf filename is unknown)
+            HString modelName = getSystemParent()->getName();
+
+            //Open file and write file headers depending on file type
+            if(mFileType == FlightPathKML) {
+                mFile.open((mFilePath.append("\\"+modelName+".kml")).c_str());
```
peterNordin, line 146: This assumes Windows, unfortunately Hopsan does not have a filepath utltiy, as far as I remember.
But you could write something like this, and use sep+modelName

`#ifdef _WIN32`
`  const HString sep = "\\"`
`#else `
`  const HString sep = "/"`
`#endif`

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
```diff
@@ -0,0 +1,329 @@
...
+            if(mFileType == FlightPathKML) {
+                mFile << "        " << longitude << "," << latitude << "," << altitude << "\n";
+            }
+
+            else if(mFileType == GroundPathKML) {
+                mFile << "        " << longitude << "," << latitude << "," << altitude << "\n";
+            }
+
+            else if(mFileType == ThreeDmodelKML) {
+                roll = (*mpPhi) * -57.29577951;
```
peterNordin, line 254: What are these magic values ?
57.29577951
3.6

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.dae
```diff
@@ -0,0 +1,761 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
```
peterNordin, line 1: Is this dae model used by the calculations in any way, or is it just an example?
If it is just and example, one could question if the file should be included at all, or if it should be located someplace else.
I dont think we have any other component that comes with data files. So this would be the first.
What does @robbr48 think? 

If this file is not needed by the component it self. Maybe it would be better to place this file together with the test model. Then the test model can also use it as the chosen model. If you move it also update the markdown documentation regarding this file.
We do have other test models that come with example data files.

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
robbr48: I guess 57.29... is 180/pi. It could perhaps be replaced with the `deg2rad` function from `ComponentUtilities.h` to make the code more readable.

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.dae
robbr48: If the component is depending on this file for input somehow, I think it should be there even if it is just an example. It should be possible for people to test this component out-of-the-box, without having to create custom files first. But then again, I am not even sure what this file actually does :)

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
alejandroSobron: I'll change that right away. This is funny: I actually had everything in "quantities" from the beginning but changed to "specific units" right before the commit because I saw that other Aero-components specified units. I think I just looked at bad/old examples... Please bear with me as it's my first time dealing with Hopsan components :)

The input Altitude was not fully defined in purpose: it's up to the user to decide if they use geometrical, pressure, QFE altitude, etc. Perhaps I will make this more clear in the documentation and stick to the Google-Earth model (WGS84 geoid). Anyway, the aircraft components are currently unable to handle differences between terrain elevation and standard (absolute) altitude when we simulate a flight, so I guess it's not a big deal just yet.

In any case, I think it's a good idea to introduce Altitude as an alias for "Position". I saw the same confusing issue in other Aero-components actually. In my opinion, it would be best to leave it "generic" and then complete the info in the documentation of a specific elevation model is required, precisely as you suggest.
I'll make these changes!

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
alejandroSobron: Yeah, this is a tricky one. I thought about restricting the input type to degrees since it's kind of standard, and this component was meant to work together with the EarthCoordinateSystem component (whose output is in deg, even though the component shows "m"(!) ).

But I think what you suggest is more reasonable: using the generic "Angle", making it clear in the documentation, and later if possible preselecting the default unit to "deg". I think I'd go for this if no better idea is proposed.

I wasn't aware that such a function (deg2rad) was available. Absolutely! I'll use it then.

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
alejandroSobron: Absolutely. You right with the "scale factors". I just wasn't sure I could leave it blank and "length" was the closest concept ;)
I'll fix this right away.

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
alejandroSobron: True! I didn't think about it. But that looks like a good solution. I'll add it!

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
alejandroSobron: @robbr48 is right, it's just a manual `rad2deg` as I wasn't aware that such a function was available. I'll change it as it is much more readable with the function.

The 3.6 is a "m/s" to "km/h". Not sure if there is a function for that, but at least I'll comment it in the code so it's clear what is going on. Or even better, I just leave it with the same units as it comes. It's not essential to get "km/h", as these values are just presented as information in the KML (possibly, in the future, they will play a role in a flight "replay" animation).

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.dae
alejandroSobron: The .dae model is only used if the user chooses to have a KML output with an embedded 3D representation of the aircraft (which is the .dae file). It is not used in the cases of "Flight path" or "Ground path" (which create just polygons and lines).
In fact, the .dae is NOT used in the calculations in any case. It's just located and referenced in the output KML file, which is the one that will try to call the .dae in when opened in Google Earth. So the Hopsan component is just making sure that the user selects a .dae model and then passes on its filepath to the KML file.

I think that the best behaviour would be that this default .dae model that I included only appears if the user selects to have a 3D KML output but he/she doesn't specify a custom .dae model. The main problem is that there is no easy way to know the absolute path to the default .dae model beforehand, and the KML file needs it.

**BUT I just got an idea: The .dae model is just a static chunk of XML code. I think that it can be included in the Hopsan-component code so it automatically generates BOTH the .kml and the .dae files at the output location when it's required. I think this is the smoothest solution: then no extra file to be included with the component :)**

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
peterNordin: Yes those other ones have not been updated, and if they are generated from mathematica, than that code must be updated first.

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
peterNordin: Yes EarthCoordinateSystem seems to have the registered unit incorrect. While we can change the C++ code this is a generated component, so the compgen code should be fixed as well. Maybe @petkr14 can help with this.

The problem here is that if the output is in degrees, we cant just connect it to an Angle input. Because the data in the model will be degrees and Angle expects rad. Any scaling or changing units is only for plotting purpose int he GUI, it does not affect the simulation data.
Of course you could put a converter gain in between.

So in this case degrees would  make sense, but then it wont be easy to change to Quantity Angle in the future. Because then old models will become incorrect. HopsanCore can not check if the "unit" is correct, a signal is only data, and the unit/quantity is only a text string.
I guess some checks could be done, at least to see if Quantity matches, but we are not doing that right now.

But maybe degrees makes sense for lat, lon. Thats is how its used most often.
Maybe lat lon would deserve its own "fictive quantity", but I dont think that is a good idea.

For now I guess you must choose, either degrees and then it will stay degrees, which is reasonable in this case. Or go with Angle but then there will be a mismatch with some other "Aerocomponents"

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.dae
peterNordin: While that would work, in this case I think it is better to leave it outside as its own file. Otherwise it will make the binary bigger and take up some RAM. Might even enter the CPU cache while simulating if its part of the component. Not sure how that works. Lets leave it where it is for now, it makes sense in this case.

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
peterNordin: Just leave a comment in the code. 3.6 = m/s to km/h, or similar

petkr14:
I think that for navigation lat and long (deg) are quit established. E.g. It is used in Google Earth so I think it is fair to yield to this. Maybe have a quantity ”Navigation”. I think in all real GPS systems you also use these (not sure....), so any software for navigation would have to deal with this anyway.
/Petter

petkr14:
The Waypoint comppnent also use longitud o lattitude in deg, so I think there is a strong case to keep deg for navigation.
/Petter

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
```diff
@@ -0,0 +1,339 @@
...
+        double *mpV;
+
+        //inputParameters pointers
+        double *mpDt;
+        double *mpScaleX;
+        double *mpScaleY;
+        double *mpScaleZ;
+
+        //inputVariables
+        double longitude;
```
peterNordin, line 69: I know that many of the other components put all of their variables here in the private part like this.
But the problem is that all of them become member variables. Even those that are only used locally inside the individual functions. Any new component should not repeat this old mistake. 

So can you please move all of the (non mName) members down to their respective functions.
That way the compiler can better optimize the code, and does not need to allocate persistent memory for these variables.
Also the variable dt is not being used.

peterNordin:
Aside from my last request, everything looks good, I tested it on Linux and seems t work, (but I did not test reading the file in Google Earth). I also have one last request, the commit history contains strange merge-commits, if possible I would like to get rid of those, so that only the content commits remain as a linear history.

I dont know what tool you are using, but if you start a git bash terminal you can give this command
`git rebase -i 35c32fd599f0979664f81aac103b2e85f6bd52d4`
- Then in the list that is shown, just replace "pick" with "drop" on the commits that you did not make.
- Then when you save and close the dialog, your branch will only contain your things.
- Then you need to `git push --force`
- If you are afraid of ruining your branch (I am sometimes), first make a backup branch
`git branch  feature/betterAeroLib_backup feature/betterAeroLib`

Typically when we work on a branch we do not merge from upstream while working. 
If we need up-to-date upstream code, we tend to rebase the entire branch. (move it as a whole) but you do not need to do that.

peterNordin:
@petkr14 Yes in the lat,lon case it makes sense with degrees, but the thing that troubles me is that in most components angles are assumed to be in rad, and here we all of a sudden have deg. So we cant assign it the quantity Angle (because it is then assumed to be rad under-the-hood). So we would need to invent a Quantity just for lat/lon or just keep it as degrees, but then not be able to change the scale in the GUI or in plots. But maybe this is not such a big issue in this particular case, because degrees is most often what you want to see.

But in the general case it is a problem, while the Hopsan Quantity scaling system needs a bit of improvement. Even if it could handle custom "under-the-hood" units it would still be odd if you allow individual components to start deciding for them selves what unit to use in their interfaces. 
I guess it might be OK if Hopsan could check and verify that a connection satisfies unit requirements, but we cant do that right now.

componentLibraries/defaultLibrary/Special/AeroComponents/AeroFlightRecorder.hpp
alejandroSobron: You are absolutely right in here. I just followed the "style" I saw in older components without even thinking twice about this issue. I'll fix it right away!

alejandroSobron:
@peterNordin I'm sorry for messing the commit history. I didn't know what was the best practice in this case (merge updates regularly from upstream or just working the branch separately). I followed your advice rebasing and dropping the other commits. I even squashed and edited my commits to leave a more understandable, linear story. I hope this branch looks cleaner now.

Thanks for all your help and sorry for the time lost during the entire reviewing. It's not only my first time writing something in Hopsan, but also my first time contributing in Git. I'm learning a lot, so next time it should be much smoother! :)

peterNordin:
@alejandroSobron for being the first time I think you did a great job, git takes "a while" to learn, and there are many different ways to use it, and we have no documentation stating our preferred method. 
We should probably add some contribution guidelines. 

Squashing is definitely a useful feature, then you can make old mistakes "go away". I use it all the time. 

Looks good now, so I am merging this.
Thank you for your contribution :)

PR #1947, opened 2020-05-20
Matlab wrapper for hopsanc

robbr48:
A Matlab class for interfacing with HopsanCore using the hopsanc library.

Depends on #1943.

### DIFF
--- a/Scripts/Matlab/hopsan.m
+++ b/Scripts/Matlab/hopsan.m
@@ -12,11 +12,16 @@ classdef hopsan
         function obj = hopsan()
             if ispc
                 path = fullfile(getenv("HOPSANHOME"), '\bin\hopsanc.dll');
+                complibpath = fullfile(getenv("HOPSANHOME"), '\componentLibraries\defaultLibrary\defaultcomponentlibrary.dll');
             elseif isunix
                 path = fullfile(getenv("HOPSANHOME"), '\bin\libhopsanc.so');
+                complibpath = fullfile(getenv("HOPSANHOME"), '\componentLibraries\defaultLibrary\libdefaultcomponentlibrary.so');
             end
             hpath = fullfile(getenv("HOPSANHOME"), '\hopsanc\include\hopsanc.h');
             loadlibrary(path,hpath);
+            if isfile(complibpath)
+                calllib('hopsanc','loadLibrary',complibpath);
+            end
             obj.checkMessages();
         end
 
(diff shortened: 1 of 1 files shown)


robbr48:
This now depends on #1952.

Scripts/Matlab/hopsan.m
```diff
@@ -0,0 +1,67 @@
...
+            bufsize=1000;
+            buf=libpointer('string',blanks(bufsize));
+            status = 0;
+            while (status == 0)
+                [status,msg] = calllib('hopsanc','getMessage',buf,bufsize);    
+                disp(msg);
+            end
+        end
+        function obj = hopsan()
+            path = fullfile(getenv("HOPSANHOME"), '\bin\hopsanc.dll');
```
peterNordin, line 13: What about Linux ? 
Also HOPSANHOME vs. HOPSAN_HOME

Scripts/Matlab/hopsan.m
```diff
@@ -0,0 +1,67 @@
...
+        function loadLibrary(obj,path)
+            calllib('hopsanc','loadLibrary',path);
+            obj.checkMessages();
+        end
+        function simulate(obj)
+            calllib('hopsanc','simulate');
+            obj.checkMessages();
+        end
+        function t = getTimeVector(obj)
+            nSamples = calllib('hopsanc','getLogSamples');
```
peterNordin, line 48: getLogSamples and setLogSamples need update due to changes in the other branch. 
"NumberOf"

Scripts/Matlab/hopsan.m
robbr48: Ok, I added Linux support, but I cannot test it since I don't have Matlab installed on Linux.

Scripts/Matlab/hopsan.m
robbr48: Fixed. But now this PR will not work in the current state, since the "NumberOf" fix is in the Python PR, which is not yet merged with master. Not a big problem though.

Scripts/Matlab/hopsan.m
peterNordin: Also need to load the default library on linux, or make the same solution as in python that load for windows also, but only if the file exists

robbr48: Ok, I added loading the default library. But again, it is untested on Linux. But it should work, since it works the same way as for Python.

Issue #1950, opened 2020-05-28
Appending to HVector is slow when vector is large


robbr48:
**Describe the bug**
`HVector::append` uses `HVector::resize`, which loops through all elements in the old vector and copies them to the new vector. This becomes very slow for large vectors.

**To Reproduce**
Use the SignalLogger component and log to a PLO file. Simulation performance will be good at begining but then become slower and slower.

**Expected behavior**
Time required for appending a value should not depend on vector size.

PR #1951, opened 2020-05-28
Replace HVector with std::vector in SignalLogger

robbr48:
Related to #1950. 

Use `std::vector` instead of `HVector` in `SignalLogger`. Resolves problem with exponentially increasing simulation time due to appending to HVector.

This does not actually solve the underlying problem, which is in the `HVector` class. There is however no reason to use HVector in this case anyway.


robbr48:
I added a clear() as you suggested.

It is not possible to preallocate memory, because components have no information of the simulation stop time. It might not even be specified in e.g. realtime simulation.

PR #1952, opened 2020-05-28
Hopsan module for Python

robbr48:
Hopsan wrapper class for Python, which loads the hopsanc library. Similar to #1943 and #1947.

### DIFF
--- a/Scripts/Python/hopsan.py
+++ b/Scripts/Python/hopsan.py
@@ -8,7 +8,8 @@ class hopsan:
         elif os.name == "nt":
             self.hdll = ctypes.cdll.LoadLibrary(os.getenv('HOPSANHOME')+"/bin/hopsanc.dll")
             libpath= os.getenv('HOPSANHOME')+"componentLibraries/defaultLibrary/defaultcomponentlibrary.dll"
-        self.hdll.loadLibrary(libpath.encode())
+        if os.path.exists(libpath):
+            self.hdll.loadLibrary(libpath.encode())
 
     def loadModel(self, path):
         self.hdll.loadModel(path.encode())
--- a/Scripts/Python/hopsan.py
+++ b/Scripts/Python/hopsan.py
@@ -2,8 +2,12 @@ class hopsan:
     def __init__(self):
         import ctypes
         import os
-        self.hdll = ctypes.cdll.LoadLibrary(os.getenv('HOPSANHOME')+"/bin/libhopsanc.so")
-        libpath= os.getenv('HOPSANHOME')+"componentLibraries/defaultLibrary/libdefaultcomponentlibrary.so"
+        if os.name == "posix":
+            self.hdll = ctypes.cdll.LoadLibrary(os.getenv('HOPSANHOME')+"/bin/libhopsanc.so")
+            libpath= os.getenv('HOPSANHOME')+"componentLibraries/defaultLibrary/libdefaultcomponentlibrary.so"
+        elif os.name == "nt":
+            self.hdll = ctypes.cdll.LoadLibrary(os.getenv('HOPSANHOME')+"/bin/hopsanc.dll")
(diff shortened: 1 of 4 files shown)


Scripts/Python/hopsan.py
```diff
@@ -0,0 +1,52 @@
+class hopsan:
+    def __init__(self):
+        import ctypes
+        import os
+        self.hdll = ctypes.cdll.LoadLibrary(os.getenv('HOPSANHOME')+"/bin/libhopsanc.so")
```
peterNordin, line 5: I would prefer HOPSAN_HOME, that would match many of the other env variables that are used in the build system

Scripts/Python/hopsan.py
```diff
@@ -0,0 +1,52 @@
...
+        libpath= os.getenv('HOPSANHOME')+"componentLibraries/defaultLibrary/libdefaultcomponentlibrary.so"
```
peterNordin, line 6: This assumes Linux, maybe support windows as well

Scripts/Python/hopsan.py
```diff
@@ -0,0 +1,52 @@
...
+        self.hdll.loadLibrary(libpath.encode())
+
+    def loadModel(self, path):
+        self.hdll.loadModel(path.encode())
+
+    def simulate(self):
+        self.hdll.simulate()
+
+    def getTimeVector(self):
+        samples = self.hdll.getLogSamples()
```
peterNordin, line 16: Hmm, I think it would be more clear, if the c-api function was called getNumberOfLogSamples, or similar. Without thinking I thought that this coped all of the samples logged. If you change it, remember the julia code.

Scripts/Python/hopsan.py
robbr48: I'm not sure really. I was inspired by `OPENMODELICAHOME`, which is used by the OpenModelica project. `HOPSANHOME` has also been used previously to find Hopsan from the OMSimulator tool

Scripts/Python/hopsan.py
robbr48: I changed it in the C  API for both get and set functions. I decided to keep the `setLogSamples` name in the Python and Julia APIs, to keep it simple.

Scripts/Python/hopsan.py
robbr48: I added Windows support. Not sure if we need to load the default library on Windows, but I added it just in case. The worst thing that can happen is that it complains that the components are already loaded.

peterNordin: True, I keep forgetting that the default library is built in on Windows.
But in this case the .dll should not even be present.
What will happen then

peterNordin: I think the best solution would be for the code to check if the defaultlibrary.dll can be found and then load it, else not try to load it. Then it would support both cases. The official release does not include that defaultlibrary dll

Scripts/Python/hopsan.py
peterNordin: I still think Hopsan should be internally consistent with naming, it is not in many cases, but adding yet an other one feels unnecessary. Or make a function getHopsanHome that checks both cases, using the first valid one ?

robbr48: No, we should only allow one variable to avoid confusion. I will keep `HOPSANHOME` for now since it was already merged in the Julia PR. If we want to change it later we can make a separate PR for all three wrappers (Julia, Python and Matlab).

Scripts/Python/hopsan.py
robbr48: I added a check for this. It also required an additional function for printing waiting messages, so that the Hopsan version information is still being printed when loading the library.

robbr48:
@peterNordin I removed the two .pyc files.

Scripts/Python/hopsan.py
peterNordin: Ok

PR #1953, opened 2020-06-02
Fix SignalLogger file stream on Windows

robbr48:
For some reason the ofstream needs to be closed and re-opened in append mode for this to work in Windows. Otherwise it only logs a few steps with a constant value. This has been tested and works on both Windows and Linux.

### DIFF
--- a/componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
+++ b/componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
@@ -193,6 +193,14 @@ namespace hopsan {
                 }
             }
 
+            //Close and re-open file in append mode
+            mFile.close();
+            mFile.open(mFilePath.c_str(),std::ios_base::app);
+            if (!mFile.is_open()) {
+                stopSimulation("Could not open file for writing: "+mFilePath);
+                return;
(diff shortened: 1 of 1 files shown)


componentLibraries/defaultLibrary/Signal/Sources&Sinks/SignalLogger.hpp
```diff
@@ -193,6 +193,14 @@ namespace hopsan {
                 }
             }
 
+            //Close and re-open file in append mode
```
peterNordin, line 196: This is very strange, I have not had time to test it my self, but is there a specific reason to do this here, after the init code. I could not see that anything was written to the file. So maybe it could be opened in append mode directly. Or well, ofcourse it needs to be cleared first. Dont know if there is a clear and append mode. Or it could be reopened directly after the first open call. Probably does not matter, just curious if that would change anything. Or move the first open call down here. 

open -> close -> open again ....

robbr48: It does write file header for column-wise CSV. The other formats need to be buffered and is written later on. 

I'm not sure if this is related to buffer size or similar, since it is different on Windows and Linux...

Issue #1794, opened 2018-11-06
Accumulator problem


MariaPettersson:
Failing accumulator in standard library?
The pressure should vary around HighPstart pressure (40 bar) due to mtion of moving wall.
Using the standard acc show variation at higher pressure.  Why?

**To Reproduce**
Run the attached model, and check the scope

![problem](https://user-images.githubusercontent.com/44434853/48051896-b4445d00-e1a6-11e8-9558-d7c53eb35755.JPG)

**Hopsan version (please complete the following information):**
*Version information is shown in the message widget when HopsanGUI is started*
 - OS: [ Windows 10]
 - Version: [e.g. HopsanCore 64-bit, Version: 2.9.0.20180912.0739
                       HopsanGUI 64-bit, Version: 2.9.0.20180915.2229]
 -

robbr48:
I don't think there is an attached model.

What do you think @petkr14?

MariaPettersson:
Best regards,
Maria Pettersson Ph.D
Senior Specialist Rock Drills
R&D Rock Drills and Rotation Units
________________________________
Epiroc
Rock Drill Product Line
Epiroc Rock Drills AB
Visitor's address: Klerkgatan 21
702 25 Örebro, Sweden
Postal address: SE-701 91 Örebro
Phone: +46196707330
E-mail: maria.pettersson@epiroc.com<mailto:maria.pettersson@epiroc.com>
http://www.epiroc.com/
Follow us at: Facebook<https://www.facebook.com/epirocgroup/> - LinkedIn<https://www.linkedin.com/company/16268827/> - Twitter<https://twitter.com/epirocgroup> - YouTube<https://www.youtube.com/channel/UCQLTt2pZVLqK1nVhjCopgeQ> - Instagram<https://www.instagram.com/epirocgroup>
________________________________
United in performance. Inspired by innovation.

peterNordin:
I do not think that you can attach/upload arbitrary files, uploading pictures seem to work though.
You have to email the model directly to Robert I think

peterNordin:
Also remember that when you replay to an email from GitHub, that email will be posted to the Issue thread and be visible to ALL. So make sure that no potentially "confidential" information is included i any such email.

peterNordin:
Also bots may harvest your email address for spam purposes (I deleted your email reply for this reason) it did not contain any useful data, the attachment had been stripped out.

robbr48:
I believe you can use zip files for arbitrary attachments:
[...]

PR #1954, opened 2020-06-03
Update accumulator components

robbr48:
This will update gas accumulator and piston accumulator components.

The gas accumulator equations are corrected.
Piston accumulator contains only variable renaming.

Changes originally made by @petkr14 .


peterNordin:
Is this related to #1794 ?

Also since these are auto generated from the mathematica files, those should be updated as well.
Otherwise I think that we should delete the .nb files related to these components and from now on treat them as "manually written components". 

It bothers me quite a lot that we have all of these autogenerated components, some with errors in them that we "can not" fix, as the fixes would be overwritten when/if they are ever regenerated.
It bothers me even more that these source files are in a proprietary format...

componentLibraries/defaultLibrary/Hydraulic/Volumes&Lines/HydraulicAckumulator.hpp
```diff
@@ -33,14 +33,15 @@
...
+//! @date Tue 27 Aug 2019 15:09:03
 //! @brief This is piston with an inertia load
 //! @ingroup HydraulicComponents
 //!
 //==This code has been autogenerated using Compgen==
-//from 
-/*{, C:, HopsanTrunk, componentLibraries, defaultLibrary, Hydraulic, \
-Volumes&Lines}/HydraulicAckumulator.nb*/
+//from
+/*{, C:, Users, petkr14, Dropbox, AircraftDesign, AeroComponents, \
```
peterNordin, line 43: I find this troublesome, that the code is generated from a source not available / synced with this code repo.
How many other components are there that we do not have updates for ?

componentLibraries/defaultLibrary/Hydraulic/Volumes&Lines/HydraulicAckumulator.hpp
```diff
@@ -157,21 +158,21 @@ here!)","",0.,&mpxmp);
      {
         //Read port variable pointers from nodes
         //Port P1
-        mpND_p1=getSafeNodeDataPtr(mpP1, NodeHydraulic::Pressure);
-        mpND_q1=getSafeNodeDataPtr(mpP1, NodeHydraulic::Flow);
-        mpND_T1=getSafeNodeDataPtr(mpP1, NodeHydraulic::Temperature);
-        mpND_dE1=getSafeNodeDataPtr(mpP1, NodeHydraulic::HeatFlow);
-        mpND_c1=getSafeNodeDataPtr(mpP1, NodeHydraulic::WaveVariable);
-        mpND_Zc1=getSafeNodeDataPtr(mpP1, NodeHydraulic::CharImpedance);
+        mpP_p1=getSafeNodeDataPtr(mpP1, NodeHydraulic::Pressure);
```
peterNordin, line 161: Not super important, but it would be even better if they were called mpP1_varName,
Then it matches the latest naming suggestion for this type of variables.

Essentially mpPortName_VariableName

componentLibraries/defaultLibrary/Hydraulic/Volumes&Lines/HydraulicAckumulator.hpp
```diff
@@ -269,7 +279,8 @@ xmp,0.1,1)*Power(V0*limit(1 - xmp,0.1,1),-1 - kappa));
           jacobianMatrix[3][4] = 0;
           jacobianMatrix[4][0] = 0;
           jacobianMatrix[4][1] = 0;
-          jacobianMatrix[4][2] = -(Zc1*onPositive(p1));
+          jacobianMatrix[4][2] = -(Zc1*dxLowLimit(c1 + \
```
peterNordin, line 282: It would be easier on the eyes of the line breaks did not occur, or at least allow longer lines, I think 160 characters, I dont remember if we decided what the maximum should be in Hopsan code.

I realize that the generator would need to be rewritten, so not super important in this PR

peterNordin:
I dont have have an opinion on the equation changes, but dont we have some sort of validation data for the testmodels for these components? Perhaps not, not sure if they are even being tested automatically.

But if we do have that, such data should be updated after these changes.

peterNordin:
@robbr48 Is there any progress on this one?
I guess we can merge the code anyway, but we must make sure to get to upstream mathematica code updated.

PR #1959, opened 2020-06-13
Kinsol solver

robbr48:
Add Sundials dependency and Kinsol solver utility. This can be used for solving algebraic equation system, either from Modelica/Mathematica models or directly from handwritten C++ components. 

Algebraic equations can be solved either using Newton iteration (for stiff problems) or fixed-point iteration (faster, for non-stiff problems). Callback functions for residuals and Jacobian matrix are added to Component class.


HopsanCore/include/ComponentUtilities/EquationSystemSolver.h
```diff
@@ -100,6 +110,30 @@ class HOPSANCORE_DLLAPI NumericalIntegrationSolver
...
+    ~KinsolSolver();
+    void solve();
+    double getState(int i);
+    void setState(int i, double value);
+    void setTolerance(double value);
+private:
+    Component *mpParentComponent;
+    void *mem;
+#ifdef USESUNDIALS
+    N_Vector y;
```
peterNordin, line 128: I dont think that sundials types should be exposed by our headers, they should be kept internal in our cpp files.
I recommend using "pimpl" solution.
Also having conditional layout of this class is a problem if you build the core with one setting (with or without) than try to link that version into an application where the macro is different for some reason. Then the class layout will mismatch, and we get memory corruption. 

In this case it seems especially unnecessary since the members are all private anyway.
Then we should not need to include the sundils.pri file everywhere as well.

componentLibraries/extensionLibrary/Components.pri
```diff
@@ -14,7 +14,8 @@ HEADERS +=  \
  $${PWD}/Mechanical/MechanicSteel.hpp \ 
  $${PWD}/Mechanical/Mechanical.h \ 
  $${PWD}/Mechanical/MechanicalSeparator.hpp \ 
- $${PWD}/Components.h
+ $${PWD}/Components.h \
+    $$PWD/Mechanical/LinearInertia.hpp
```
peterNordin, line 18: minor tab / space issue

componentLibraries/extensionLibrary/extensionLibrary.pro
```diff
@@ -15,6 +15,8 @@ DESTDIR = $${PWD}/
 # *= Means append unique
 INCLUDEPATH *= $${PWD}/../../HopsanCore/include/
 
+INCLUDEPATH *= $${PWD}/../../Dependencies/sundials-5.3.0/include/
```
peterNordin, line 18: If we hide sundiles inside hopsancore then the component library do not need to know about the library,
only the utility that we expose that wraps it

componentLibraries/extensionLibrary/extensionLibrary.xml
```diff
@@ -18,6 +18,8 @@
     <cflags>-I./someLib/include</cflags>
     <lflags>-L./someLib/bin -lsomeLib</lflags>
     <lflags os="win64">-lWs2_32</lflags> -->
+    <cflags>-I../../dependencies/sundials/include -DUSESUNDIALS</cflags>
```
peterNordin, line 21: same here, should not be needed if hopsacore wraps it internally

peterNordin:
I did not review all of the equation code, because I dont have time to refresh my understanding of TLM and such things...

peterNordin:
I also find it slightly disturbing that HopsanCore will have an (at least optional) dependency that we can not easily include when we export to FMU and such things. 
It would be nice if we could but making that work seems like a lot of work unless the code could be included with hopsan, like numhop and the csvparser.

HopsanCore/include/ComponentUtilities/EquationSystemSolver.h
robbr48: Good suggestion, that was actually the solution I was looking for. Now everything became much nicer.

robbr48:
I actually made it work for FMI export. It works in Linux, but is not yet tested on Windows. All that is needed is to place the Sundials dll/so files in the /resources directory in the FMU and add a relative search path when linking. I assume it is not harder for S-functions. Labview should work with some hands-on work. Executable export will also need to be supported.

componentLibraries/extensionLibrary/extensionLibrary.pro
peterNordin: I think this one can also be removed

dependencies/setupSundials.bat
```diff
@@ -0,0 +1,41 @@
+@ECHO OFF
+REM $Id$
```
peterNordin, line 2: For new files, remove the old Subversion $Id$ thing, also the setup scripts should be added to the setupAll.bat and .sh

peterNordin:
Also add installation of the library to the "copyInstall" script and the makeWindowsRelease script
A guess it should be handled like the fmilibrary

robbr48:
Ok, I think I got everything right. I don't think we need to include the entire directory in the installation as we do with FMILibrary though, since we users will not need to compile it.

robbr48:
I converted it to a submodule, which is compiled into HopsanCore. This should improve maintainability and eliminate the problem with external dependencies for HopsanCore.

@peterNordin Can you have a look at the CMake stuff? I need to call CMake to generate the sundials_config.h file, which is needed for compilation. In-source build is not allowed, so I had to create a separate build directory only for this purpose. Perhaps this can be done in a better way.

peterNordin:
I actually think the previous solution was better, unless we can get rid of the CMake step entirely. It is problematic if CMake is required (unless we ship it with hopsan) especially for Windows users. Also I have a feeling that I will run into problems for snap and flatpak packaging as well. 
Would it be possible to include a pregenerated header file ? Or does it depend on the current operating system and version.

I just feel that if a library require a build system of its own to work, then it should not be a required part of HospanCore.
Of-course this is what you are trying to fix here, and maybe its possible, but we must get rid of the additional cmake build step, or include cmake but then I feel that the hopsancore becomes to complex, the intention was to have no external dependencies for the core itself.

HopsanCore/HopsanCore.pro
```diff
@@ -39,6 +39,53 @@ SOURCES += $${PWD}/dependencies/libnumhop/src/VariableStorage.cpp
...
+cmake ../sundials \
+-DCMAKE_INSTALL_PREFIX=$${PWD}/../bin \
+-DCMAKE_BUILD_TYPE=RELEASE \
+-DBUILD_ARKODE=OFF \
+-DBUILD_CVODE=OFF \
+-DBUILD_CVODES=OFF \
+-DBUILD_IDA=OFF \
+-DBUILD_IDAS=OFF \
+-DBUILD_KINSOL=ON \
+-DBUILD_SHARED_LIBS=ON\
```
peterNordin, line 54: Shared libs should be OFF if we build it in I think

HopsanGenerator/src/GeneratorUtilities.cpp
```diff
@@ -687,6 +715,7 @@ QStringList listHopsanCoreIncludeFiles(const QString &hopsanInstallationPath)
     findAllFilesInFolderAndSubFolders(hopsanInstallationPath+"/dependencies", "hpp", allFiles);
     findAllFilesInFolderAndSubFolders(hopsanInstallationPath+"/HopsanCore/dependencies/rapidxml", "hpp", allFiles);
     findAllFilesInFolderAndSubFolders(hopsanInstallationPath+"/HopsanCore/dependencies/libnumhop/include", "h", allFiles);
+    findAllFilesInFolderAndSubFolders(hopsanInstallationPath+"/HopsanCore/dependencies/sundials/include", "h", allFiles);
```
peterNordin, line 718: Maybe include "HopsanCore/dependencies/sundials-build/include" here as well.
But would be better if the cmake step could be avoided and the headers be manually included

HopsanCore/HopsanCore.pro
robbr48: CMake enforces either static or shared libraries to be turned on. But it doesn't matter, since we don't actually build anything with their CMake system.

robbr48:
Ok, we can just ship the config file separetly instead to remove the CMake requirement. Only drawback is that we might need to update it if we for some reason do a submodule update. But the Sundials libraries are very mature and well tested, so it should not be needed that often.

peterNordin:
Seems resonable. Good that it worked. Now you only need to add the sundials files to our CMake system. So that they are included in the build and installed with the hopsan core source code. You can probably do it the same way as libnumhop.

robbr48:
Ok, I think I got it right. I hope I did not miss any files.

HopsanCore/CMakeLists.txt
```diff
@@ -61,3 +62,4 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/src/ DESTINATION ${HOPSANCORE_INST
 install_libnumhop_src(${HOPSANCORE_INSTALL_DST}dependencies/libnumhop)
 install_indexingcsvparser_src(${HOPSANCORE_INSTALL_DST}dependencies/indexingcsvparser)
 install_rapidxml_src(${HOPSANCORE_INSTALL_DST}dependencies/rapidxml )
+install_sundials_src({HOPSANCORE_INSTALL_DST}dependencies/sundials)
```
peterNordin, line 65: Seems to be name difference, 
install_sundials_src vs. install_libsundials_src in sundials.cmake

robbr48: Ops, it is fixed now.

peterNordin:
@robbr48 you forgot to add `add_sundials_src(hopsancore)` to HopsanCore CMakeList.txt also that function is not adding the sundials-extra/include files, both to the header list and to the include path. `$<BUILD_INTERFACE:${sundials_extra_dir}/include>`
But an even bigger problem is that there are a lot of implementation headers inside the source directory. And all of the potential src code locations must be added to the include path so that the headers can be found. For example: 
` $<BUILD_INTERFACE:${sundials_dir}/src/arkode`

But since I did not see all of this in the pro file version, or in the HopsanGEnerator code, I fail to see how this can even work at all. But the travis and appveyor builds are green, so I dont know.

Also I was thinking, are we not adding all of the sundials library parts now, or are the sundils_config suppose to de-active some parts. Right now all of the src files are added to hopsan and compiled in.

robbr48:
Ok, I fixed the missing stuff and added an explicit list of source code. I hope I got everything right, but I'm not sure if it is ok to still glob all *.h files.

The Sundials CMake system copies only the required source files to a separate build directory, depending on which part that shall be built. I guess we could do the same, but since the current method works for QMake it should probably work also for CMake.

peterNordin:
@robbr48 did you push these changes? :)

robbr48:
Ops, it required a force push due to rebasing with master. Now it is pushed :)

robbr48:
Now everything is rebased and squashed. There are still some commits that redo other commits, but the history is too complex to do more squashing I think.

peterNordin (approved):

peterNordin (changes requested):

peterNordin (approved):

PR #1962, opened 2020-06-22
Always reload models before starting a new optimization

robbr48:
This fixes a weird bug where changing a model between two optimization has no effect. For some reason the temporary optimization models were only updated before the first optimization. They should be reloaded every time, in case the model has changed.

@petkr14 This should solve the problem with your filter optimization model.


robbr48:
Yes, that is exactly what happens. I'm surprised no one has noticed this before, but perhaps people don't rebuild their models between optimization that often.

PR #1963, opened 2020-06-23
Issue 1950 reserve hvector

### LINKED ISSUE #1950: Appending to HVector is slow when vector is large

peterNordin:
Add reserve function to HVector that works similarly to the one in std::vector 
On resize() auto reserve x2 size when growing, useful in append case to avoid reallocating every time

Resolve #1950


HopsanCore/include/HopsanTypes.h
```diff
@@ -160,73 +168,70 @@ class HVector
...
+            mSize = s;
         }
-
-        // Copy old data to new array
-        for (size_t i=0; i<nCopy; ++i)
-        {
-            pNewArray[i] = mpDataArray[i];
+        else {
+            // Reserve twice the memory, to avoid reallocation on every resize in append case
+            //! @todo Multiplying with 2 may be OK for small arrays, but for very large ones this could be a problem
```
robbr48: Is multiplication by two a state-of-the-art solution used by other libraries etc?

.gitignore
```diff
@@ -1,9 +1,11 @@
 .svn/
+.vs/
 bin/
```
robbr48, line 3: Is this (and other changes to .gitignore) related to this PR?

Common.prf
```diff
@@ -18,6 +18,9 @@ CONFIG(release, debug|release) {
   QMAKE_CXXFLAGS += -O3
 }
 
+# Enable C++14
+CONFIG += c++14
+
```
robbr48, line 23: I guess this is required for the reserve functions. So this means that we can no longer build HopsanCore with old compilers? (not a problem, since we already decided on this, I just want to make sure).

HopsanCore/include/HopsanTypes.h
peterNordin: I dont know, someone told me that is what the standard library in gcc does, but I have not been able to verify that, should be quite easy now that I think about it. Maye will do at some point. 
I could not think of any other clever solution here. But if you know the size of the data before hand, then reserve is preferred.
I was also considering adding a preallocation of maybe 10, to speed up resize for small arrays, but decided not to.

.gitignore
peterNordin: No, buts its in a separate commit, I was forcing myself to use MS VisualStudio to code this, and it puts the CMake output into the source-code directory, also a .vs folder for its own files.

Common.prf
peterNordin: No this is completely unrelated, I dont remember the problem exactly, but I got linking problems due to not all parts of Hopsan using the same C++ standard version (maybe on Linux). 

Anyway, since we do "Rebase and Merge" (I do ff-only merge to keep my commit signatures) I have no problems adding some extra unrelated commits to PR like this. since they do not actually add new functionality or such things, its only cleanup och the build system files.
But I guess it would have been more "correct" to put them in a separate PR.

robbr48: Ok, then I understand. In general I prefer a separate PR for everything, but I guess some minor changes does not really matter. It is more important in projects that enforce squashing all commits when merging a PR.

PR #1971, opened 2020-08-18
Correct tab names after reload with multiple models

robbr48:
Apparently the fix in #1965 was not sufficient in all cases. This fix should work also when there are more than one text tab open.


peterNordin (approved):
I have no idea what that does but since it fixes the problem its OK,
some comment would have been nice, but not necessary to add

Issue #1976, opened 2020-08-19
logonly matches all signals on port


mrejster:
**Describe the bug**
When using hopsanCLI, the logonly is crucial for reducing output file size. However, the matching of port names do not work as expected. For example, if I wish to log only the signal:
lineRCandLosses_24#P2#Flow

I end up only saving every other value on that port, i.e.,
lineRCandLosses_24#P2#Pressure
lineRCandLosses_24#P2#Temperature
etc. resulting in around 4x larger csv files than needed (=slow)

**To Reproduce**
run a model using hopsanCLI
export data to csv using the logonly-flag and a single port name

**Expected behavior**
Only the exact match should be exported

**Hopsan version (please complete the following information):**
*Version information is shown in the message widget when HopsanGUI is started*
 - OS: Windows 10
 - Version: [e.g. HopsanCore 64-bit, Version: 2.14.1
                       HopsanGUI 64-bit, Version: 2.14.1
 - Build: Official

**Additional information**
The option to also export as .h5 would be great!

peterNordin:
Currently the logonly option specifies Ports, not variables in ports so the entire port will be logged to memory and this is not easy to change. But I guess we could choose what to export in the export step, to atleast create a smaller csv file.

.h5 support is also very time consuming to implement. Since it is a function built into the GUI, we would need to create a new "result logging" library, loaded by both the cli and gui, (like hopsangenerator). This would be nice to have, but it will take to much time.

But I guess one thing that could work (and is easier to implement) is to create a "binary csv" format, that doesn't write the values as text, I think that would be more compact and likely faster to write/read.

mrejster:
Ok, understood. I can live with the slowdown for now. I found a much bigger issue with the parameter import, I'll file a new issue for that.

robbr48:
I workaround here is to use sensor components for the important variables, and log the sensor output. Signal ports only have one variable.

peterNordin:
Also noticed a potential problem, the output uses scientific notation but the default precision of 6 significant numbers.
Increasing it would make files even larger, so maybe saving data as binary could be useful.

mrejster:
Ok, yes that could be an issue.

A possible(?) workaround for h5 export: maybe hopsan CLI can be modified to accept a script file in the the call. Then the regular sapl h5-functionality could be used. or maybe HCOM is tightly connected to hopsan GUI...

robbr48:
[...]

PR #1985, opened 2020-08-23
Issue 1976 filter individual variables with log only

peterNordin:
Related to #1976 

Enable filtering csv results on individual variables, the entire port is still logged in memory, but only the desired variables are written to file. Entire ports can still be specified like it used to work.

Test command-line:
`./hopsancli -m ../../hopsan-code/Models/Example\ Models/Sub\ System\ Example.hmf -s hmf --logonly Servo1\$Gain\#out,Pulse\#out\#Value,Servo1\$Cylinder_C\#P3\#Position,Servo1\$Translational_Mass\#P2 --resultsFullCSV out.csv`

### DIFF
--- a/HopsanCLI/ModelUtilities.cpp
+++ b/HopsanCLI/ModelUtilities.cpp
@@ -33,6 +33,7 @@
 
 #include <iostream>
 #include <fstream>
+#include <algorithm>
 
 #include "ModelUtilities.h"
 #include "version_cli.h"
@@ -42,6 +43,16 @@
 
 using namespace std;
 using namespace hopsan;
+
+namespace {
+
+template <typename ContainerT, typename ValueT>
+bool contains(const ContainerT& container, const ValueT& value) {
+    auto it = std::find(container.begin(), container.end(), value);
+    return (it != container.end());
+}
+
+}
 
 void generateFullSubSystemHierarchyName(const ComponentSystem *pSys, HString &rFullSysName)
 {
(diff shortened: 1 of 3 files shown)


HopsanCLI/main.cpp
```diff
@@ -895,20 +890,20 @@ int main(int argc, char *argv[])
                             {
                                 if (!line.empty())
                                 {
-                                    enabled_ports.push_back(line);
+                                    logOnlyPortsOrVariables.push_back(line);
```
robbr48, line 893: This variable name is really confusing, but I can't think of something better now.

HopsanCLI/ModelUtilities.cpp
```diff
@@ -43,6 +40,19 @@
 using namespace std;
 using namespace hopsan;
 
+namespace {
+
+bool findAndRemove(std::vector<std::string>& rVector, const std::string& value) {
```
robbr48, line 45: As I understand it, this function removes the variable from the filter vector, and returns true if the variable was found. Why is it necessary to remove the variable? Is it not enough to check if the filter vector contains the variable? I guess there is a good reason, but at least adding some comments would be helpful.

robbr48 (changes requested):
Looks good but could use some comments.

HopsanCLI/ModelUtilities.cpp
peterNordin: Yes, I will add some comments.
I remove the found name  so that the following searches will go faster. Probably no noticeable speed difference (especially for short lists), but just felt that it was unnecessary to keep it after it has been found since they will never be found again.

peterNordin: Never mind, I changed it to a simple contains() function, that only looks for the data and doesn't remove it from the includeFilter, also takes includeFilter as const& as it wont need to change anymore. Should be more clear and robust.

robbr48 (approved):

PR #1987, opened 2020-08-26
Complete rework of Modelica generator

robbr48:
Major improvements:
- Use Kinsol to solve algebraic equation systems
- Select lienar transform method for each model using Modelica annotations
- Apply variable limits by resetting delays instead of `dxLimit()` functions
- Support custom Hopsan functions
- Support algorithm sections and conditional statements (experimental)
- Improved symbolic simplifications for better performance


robbr48:
I know that this is very hard to review, but I created a test library with some components that demonstrates some of the new features.

[ModelicaTest.zip](https://github.com/Hopsan/hopsan/files/5128944/ModelicaTest.zip)

peterNordin:
QRandomGenerator is to new, consider using standard c++ random generator if possible

peterNordin (commented):
Yes you are right, difficult to review, lots of complex code. Just browsed through it.
Would it be possible to add a test modelica model, that can be converted using the generator (and preferably simulated, but not sure if the generator tests can do that). Not sure how to verify the generated code though.

peterNordin:
Also symhop test failes

peterNordin:
Looks good now, but I just remembered in the GeneratorTest there is a test "Generator_Modelica" that runs but says its disbaled.
Should this test be fixed or removed ?

peterNordin:
The unit test does not work with CMake, but I will try to fix it

peterNordin:
There is a build error if 	DEVELOPMENT is not defined around line 1316, in GUIContainerObject.cpp

peterNordin (changes requested):
Fix build issue when DEVELOPMENT is not defined

peterNordin (approved):

PR #1990, opened 2020-08-27
HDF5 export from HopsanCLI

robbr48:
- [x] Move export code to separate library
- [x] Call library from HopsanCLI

### DIFF
--- a/HopsanCLI/HopsanCLI.pro
+++ b/HopsanCLI/HopsanCLI.pro
@@ -69,6 +69,19 @@ CONFIG(debug, debug|release) {
   QMAKE_CXXFLAGS += -pedantic -Wno-long-long -Wconversion
 }
 
+# Set hdf5 paths
+LIBS += -L$${PWD}/../lib -lhopsanhdf5exporter$${DEBUG_EXT}
+include($${PWD}/../dependencies/hdf5.pri)
+have_hdf5(){
+  INCLUDEPATH *= $${PWD}/../hopsanhdf5exporter
+  DEFINES *= USEHDF5
+  !build_pass:message("Compiling with HDF5 support")
+} else {
+  LIBS -= -L$${PWD}/../lib -lhopsanhdf5exporter$${DEBUG_EXT}
+  !build_pass:message("Compiling without HDF5 support")
+}
+
+
 # -------------------------------------------------
 # Platform specific additional project options
 # -------------------------------------------------
--- a/HopsanCLI/HopsanCLI.pro
+++ b/HopsanCLI/HopsanCLI.pro
@@ -65,22 +65,23 @@ contains(DEFINES, HOPSANCLI_USEGENERATOR) {
 }
 #--------------------------------------------------------
 
-CONFIG(debug, debug|release) {
-  QMAKE_CXXFLAGS += -pedantic -Wno-long-long -Wconversion
-}
-
+#--------------------------------------------------------
 # Set hdf5 paths
 LIBS += -L$${PWD}/../lib -lhopsanhdf5exporter$${DEBUG_EXT}
 include($${PWD}/../dependencies/hdf5.pri)
 have_hdf5(){
   INCLUDEPATH *= $${PWD}/../hopsanhdf5exporter
   DEFINES *= USEHDF5
-  !build_pass:message("Compiling with HDF5 support")
+  !build_pass:message("Compiling HopsanCLI with HDF5 support")
 } else {
-  LIBS -= -L$${PWD}/../lib -lhopsanhdf5exporter$${DEBUG_EXT}
-  !build_pass:message("Compiling without HDF5 support")
+  LIBS -= -lhopsanhdf5exporter$${DEBUG_EXT}
+  !build_pass:message("Compiling HopsanCLI without HDF5 support")
 }
+#--------------------------------------------------------
 
+CONFIG(debug, debug|release) {
+  QMAKE_CXXFLAGS += -pedantic -Wno-long-long -Wconversion
+}
 
 # -------------------------------------------------
 # Platform specific additional project options
--- a/HopsanCLI/ModelUtilities.cpp
+++ b/HopsanCLI/ModelUtilities.cpp
@@ -36,6 +36,11 @@
 #include "CliUtilities.h"
 
 #include "HopsanEssentials.h"
+#include "HopsanTypes.h"
+
+#ifdef USEHDF5
+#include "hopsanhdf5exporter.h"
+#endif
 
 using namespace std;
 using namespace hopsan;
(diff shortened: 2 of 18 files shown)


HopsanCLI/HopsanCLI.pro
```diff
@@ -69,6 +69,19 @@ CONFIG(debug, debug|release) {
   QMAKE_CXXFLAGS += -pedantic -Wno-long-long -Wconversion
 }
 
+# Set hdf5 paths
+INCLUDEPATH *= $${PWD}/../hopsanhdf5exporter
+LIBS += -L$${PWD}/../lib -lhopsanhdf5exporter$${DEBUG_EXT}
+DEFINES *= HOPSANHDF5EXPORTER_DLLIMPORT
```
peterNordin, line 75: should not need HOPSANHDF5EXPORTER_DLLIMPORT since it is a static library

HopsanCLI/HopsanCLI.pro
```diff
@@ -69,6 +69,19 @@ CONFIG(debug, debug|release) {
   QMAKE_CXXFLAGS += -pedantic -Wno-long-long -Wconversion
 }
 
+# Set hdf5 paths
+INCLUDEPATH *= $${PWD}/../hopsanhdf5exporter
+LIBS += -L$${PWD}/../lib -lhopsanhdf5exporter$${DEBUG_EXT}
```
peterNordin, line 74: Move include path and libs into the check below, so that they are only added if have_hdf5() is true

HopsanCLI/HopsanCLI.pro
```diff
@@ -121,3 +134,5 @@ HEADERS += \
     core_cli.h \
     ModelUtilities.h \
     BuildUtilities.h
+
+message($${LIBS})
```
peterNordin, line 138: leftover

HopsanCLI/ModelUtilities.cpp
```diff
@@ -36,6 +36,9 @@
 #include "CliUtilities.h"
 
 #include "HopsanEssentials.h"
+#include "HopsanTypes.h"
+
+#include "hopsanhdf5exporter.h"
```
peterNordin, line 41: should this be conditionally included ?

HopsanGUI/LogDataHandler2.h
```diff
@@ -72,6 +72,7 @@ class LogDataHandler2 : public QObject
     void exportToCSV(const QString &rFilePath, const QList<SharedVectorVariableT> &rVariables) const;
     void exportGenerationToCSV(const QString &rFilePath, int gen) const;
     void exportToHDF5(const QString &rFilePath, const QList<SharedVectorVariableT> &rVariables) const;
+    void exportToHDF5_basic(const QString &rFilePath, const QString &rModelFileName, const QString &rToolName, const QList<QStringList> &rSystemHierarchies, const QStringList &rComponentNames, const QStringList &rPortNames, const QStringList &rVariableNames, const QStringList &rAliasNames, const QList<QVector<double> > &rDataVectors, const QStringList &rUnits, const QStringList &rQuantities) const;
```
peterNordin, line 75: what functions is this _basic, leftover ?

hopsanhdf5exporter/hopsanhdf5exporter.h
```diff
@@ -0,0 +1,25 @@
+#ifndef HOPSANHDF5EXPORTER_H
+#define HOPSANHDF5EXPORTER_H
+
+#include "hopsanhdf5exporter_win32dll.h"
+#include "HopsanEssentials.h"
+
+using namespace hopsan;
+
+class HOPSANHDF5EXPORTER_DLLAPI HopsanHDF5Exporter
```
peterNordin, line 9: No need for HOPSANHDF5EXPORTER_DLLAPI since static library

hopsanhdf5exporter/hopsanhdf5exporter.pro
```diff
@@ -0,0 +1,59 @@
...
+CONFIG += staticlib
+
+TARGET = hopsanhdf5exporter
+DESTDIR = $${PWD}/../lib
+
+INCLUDEPATH += $${PWD}/include
+
+TARGET = $${TARGET}$${DEBUG_EXT}
+
+DEFINES += HOPSANHDF5EXPORTER_LIBRARY
```
peterNordin, line 19: Not used ?

hopsanhdf5exporter/hopsanhdf5exporter.pro
```diff
@@ -0,0 +1,59 @@
+#-------------------------------------------------
+#
+# Project created by QtCreator 2020-08-27T10:38:46
+#
+#-------------------------------------------------
+include( ../Common.prf )
+
+QT       -= gui
+TEMPLATE = lib
```
peterNordin, line 9: looks like Qt is no longer used in this library (good) 
Add QT -core

hopsanhdf5exporter/hopsanhdf5exporter.pro
```diff
@@ -0,0 +1,59 @@
...
+
+TARGET = $${TARGET}$${DEBUG_EXT}
+
+DEFINES += HOPSANHDF5EXPORTER_LIBRARY
+
+# The following define makes your compiler emit warnings if you use
+# any feature of Qt which has been marked as deprecated (the exact warnings
+# depend on your compiler). Please consult the documentation of the
+# deprecated API in order to know how to port your code away from it.
+DEFINES += QT_DEPRECATED_WARNINGS
```
peterNordin, line 25: Not needed if no Qt used

hopsanhdf5exporter/hopsanhdf5exporter.pro
```diff
@@ -0,0 +1,59 @@
...
+#DEFINES += QT_DISABLE_DEPRECATED_BEFORE=0x060000    # disables all the APIs deprecated before Qt 6.0.0
+
+SOURCES += \
+        hopsanhdf5exporter.cpp
+
+HEADERS += \
+        hopsanhdf5exporter.h \
+        hopsanhdf5exporter_win32dll.h
+
+unix {
```
peterNordin, line 56: non of the other .pro files have installation specified, so maybe remove this
I am trying to fix this in a cross-platform cross-compiler way in the CMake build system

hopsanhdf5exporter/hopsanhdf5exporter_win32dll.h
```diff
@@ -0,0 +1,53 @@
+/*-----------------------------------------------------------------------------
```
peterNordin, line 1: This file should not be needed since static library

HopsanCore/src/ComponentSystem.cpp
```diff
@@ -882,6 +882,24 @@ std::vector<HString> ComponentSystem::getSubComponentNames() const
...
+//! @brief Returns a list of pointers to all components the system and all sub systems
+//! @param [out] rComponents Pointers to all components
+void ComponentSystem::getSubComponentsRecursively(std::vector<Component*> &rComponents) const
+{
+    //! @todo for now create a vector of the component names, later maybe we should return a pointer to the real internal map
+    SubComponentMapT::const_iterator it;
+    for (it = mSubComponentMap.begin(); it != mSubComponentMap.end(); ++it)
+    {
+        if(it->second->isComponentSystem()) {
+            reinterpret_cast<ComponentSystem*>(it->second)->getSubComponentsRecursively(rComponents);
```
peterNordin, line 895: This will not include the system itself, but I assume it should be included since it is also a component, which may have ports.

hopsanhdf5exporter/hopsanhdf5exporter.cpp
```diff
@@ -0,0 +1,187 @@
...
+        // We need this to avoid massive exception casting when creating directories as
+        // group names will be repeated, also we need to create one group depth at a time
+        // The map will sort the group names as unique values in the correct order
+        std::map<HString, char> varhier;
+        for(size_t i=0; i<mSystemHierarchies.size(); ++i) {
+
+            //Split system hierarchy string to a vector of sub strings
+            std::vector<HString> sysnames;
+            while(!mSystemHierarchies[i].empty()) {
+                sysnames.push_back(mSystemHierarchies[i].substr(0, mSystemHierarchies[i].find("."))); // token is "scott"
```
peterNordin, line 67: "scott" ? dott ?
Also same thing further down

hopsanhdf5exporter/hopsanhdf5exporter.cpp
```diff
@@ -0,0 +1,187 @@
...
+                if(!mSystemHierarchies[i].containes(".")) {
+                    mSystemHierarchies[i].clear();
+                }
+                else {
+                    mSystemHierarchies[i].erase(0,mSystemHierarchies[i].find(".") + 1);
```
peterNordin, line 72: This seems like its erasing text from the strings in each item in mSystemHierarchies, but this variable is reused further down (line 112), but is not all the text gone in that case?

peterNordin (changes requested):

peterNordin:
Also cmake build is missing, but I will add on my other branch where I have added hdf5 inclusion.
It makes no difference on this branch.

hopsanhdf5exporter/hopsanhdf5exporter.cpp
robbr48: You are right. I didn't realize any problems though, but I will change to a temp variable.

hopsanhdf5exporter/hopsanhdf5exporter.cpp
robbr48: Ops, that came from stackoverflow :)

HopsanGUI/LogDataHandler2.h
robbr48: Yes, it was my temporary implementation before creating the static library. I will remove it.

HopsanCLI/HopsanCLI.pro
robbr48: Actually this is a problem. The exporter lib needs to be linked before the hdf libs, or there will be undefined references. I will need to think some more about this.

peterNordin: But if hdf5 is not available, then they CLI should not link withbthe exporter, because it wont exist.

hopsanhdf5exporter/hopsanhdf5exporter.cpp
peterNordin: Did you test with a subsystem example model, I try to do that whenever possible to detect this kind of things.

HopsanCLI/HopsanCLI.pro
robbr48: I "solved" it by always adding libs, and then removing it if hdf5 was not found. Kind of ugly, but it works.

HopsanCLI/CMakeFiles/CMakeDirectoryInformation.cmake
```diff
@@ -0,0 +1,16 @@
+# CMAKE generated file: DO NOT EDIT!
```
peterNordin, line 1: You accidentally added and committed the CMake build cache, it should be removed

peterNordin: I pushed a fixup commit to remove these files, but you need to squash them

HopsanCLI/HopsanCLI.pro
peterNordin: Aha, I realize now why it did not work, the "have_hdf5" function not only checks, it adds link flags as well. So they end up before hopsanhdf5exporter unless you add the link flags before. I guess we have to live with this solution in the qmake build system.

peterNordin:
@robbr48  I think that I was able to solve the build issues when hdf5 is not available I have pushed some commits, two that you need to squash with previous work, and two new ones. The last is technically unrelated but it adds a nice speed improvement to the qmake build (no need to wait for subprojects to build one at a time) and since you added a new subproject I think it make sens to throw it in with this PR.

peterNordin:
Hmm apperantly not quite done yet, CI still fail, possibly due to some misstake in.my speedup commit.

hopsanhdf5exporter/hopsanhdf5exporter.cpp
robbr48: Yes, I have tested it for the subsystem example and listed the contents using h5ls utility. It appears to work for both final and full results.

peterNordin (approved):

Issue #1988, opened 2020-08-26
Separate repository for development libraries


larsviktorlarsson:
There should be an official repository for Hopsan development libraries, that is separated from the Hopsan repo. In this repo, users can have their own public libraries with components that are still under construction. 

This way, it is possible to share libraries with others and also allow others to give feedback before trying to merge new components into Hopsan master. This repo could also contain temporary libraries used e.g. in courses and teaching.

It should be noted that this repo is for development libraries, which means that the users should be aware of this and thus cannot expect the components to work perfectly if working at all.

Please add additional requirments/desires in this issue before the repo is created.

larsviktorlarsson:
Perhaps this repo could contain an example library, with example components, xml files, md files and proper naming of the different files?

peterNordin:
The example library already exists in the default repo and is included in the hopsan release, so it should stay there.
Any library that should be shipped with hopsan "must" be in the main repository. Or one may use git submodules, but lets avoid that if possible.

larsviktorlarsson:
> The example library already exists in the default repo and is included in the hopsan release, so it should stay there.
> Any library that should be shipped with hopsan "must" be in the main repository. Or one may use git submodules, but lets avoid that if possible.

Ok, but then the information in the readme maybe should include a path to the example library in hopsan?

PR #1991, opened 2020-08-27
Moved devLibraries to external repository, see issue 1988

peterNordin:
New repo here [https://github.com/Hopsan/development-libraries](https://github.com/Hopsan/development-libraries)
Related to #1988


robbr48 (approved):
Looks good. I believe there are some components in default library, especially under "Special", that should be moved as well, either to the external repository or to extensionLibrary. But that can be done later in a separate PR.

larsviktorlarsson (approved):
Looks good!

Perhaps the README file should contain a brief guide on how to use devLibraries (file structure, file names)? I think the example library should be mentioned here as well (if not moved to this repo). I was unaware of the existance of the example library until @peterNordin mentioned it in a previous comment, so I think it should be highlighted. Personally, I prefer to look at examples rather than reading guides.

peterNordin:
@larsviktorlarsson I have updated the README in https://github.com/Hopsan/development-libraries

PR #1993, opened 2020-09-03
Fix release script after various updates to dependencies

peterNordin:


makeWindowsRelease.py
```diff
@@ -26,7 +26,7 @@
 gTemporaryBuildDir = r'C:\temp_release'
 
 # External programs
-inkscapeDirList = [r'C:\Program Files\Inkscape', r'C:\Program Files (x86)\Inkscape']
+inkscapeDirList = [r'C:\Program Files\Inkscape\bin', r'C:\Program Files\Inkscape', r'C:\Program Files (x86)\Inkscape']
```
robbr48, line 29: Should we also add a bin path to the 32-bit version?

robbr48 (approved):
Looks good, just one (probably not important) comment.

makeWindowsRelease.py
peterNordin: Yes I will add that

PR #1994, opened 2020-09-08
Display highest generation command in HCOM

robbr48:
Get current highest generation in HCOM:
- HCOM "dihg" command
- HCOM "hg()" functionoid


peterNordin (approved):
OK, but one thing that is missing is a re-generated documentation for hcom, but I can add that to the release branch later. I need to do that in case there were other changes as well.

Issue #1966, opened 2020-08-14
System crash when opening animation dialogue if the movable index does not start at 0


larsviktorlarsson:
**Describe the bug**
*Hopsan crashes when animation dialogue is opened and the index of the movables does not start at 0*

**To Reproduce**
*If the bug can be reproduce, please give the steps taken to reproduce the behavior:*
1. Make a component which has animation features with several movables
2. Let the index of the first movable be idx="1" (movable idx="1")
3. Drag the component to a model
4. Go into animation mode
5. Double click on the movable to open the animation dialogue
6. Hopsan crashes 

**Expected behavior**
*Perhaps this should be checked as the library is loaded, and generate a warning? Alternatively, one could have an error dialogue that opens instead of crashing the gui*

robbr48:
Confirmed. Hopsan accesses the movables by the order they are loaded from XML, which likely also causes problems if the movables are not sorted according to their indices.

I was able to fix this for the dialogs, but the animated components also uses the weird indexing. I need to think of a solution for this.

peterNordin:
@robbr48 did you merge some partial fix for this ?

robbr48:
No, but I can do that tomorrow I think.

robbr48:
After looking some more into this, I don't think this is a problem anymore. The movables are updated in the order they are specified in the XML file, but it doesn't matter as long as each movable is updated with its own parameters. Also, they seem to be sorted in Z-level according to their index, so that the highest index is always on top.

PR #1996, opened 2020-09-10
Correct indexing of movables in animation dialogs

robbr48:
Related to #1966.

This fixes the crash when opening properties dialog of animation icons with wrong index. For correct animation, zero-indexed movables is however still required. Everything in the implementation is based on accessing vector elements using indices. We should change this to maps, but it is used on hundreds of places so it will have to wait for now.


peterNordin (approved):
As you said, more work is needed, but this is at least a small improvement, I wont add it to the release notes though, as the problem still seem to be there

PR #2001, opened 2020-09-25
Parallel simulation option for HopsanCLI

### LINKED ISSUE #1999: Multicore support for hopsanCLI
Multicore seems unsupported (or at least undocumented) for the hopsanCLI application?

Add the option to use multicore simulation with the hopsanCLI application

Manually firing up multiple hopsanCLI instances, but using the complex optimization method requires each run to be finished before updating parameters. For other techniques it might be possible, if different evaluations can be run simultaneously.

**Additional context**
-

robbr48:
New option for parallel simulation in HopsanCLI. Resolves #1999.
```
   -p <integer>,  --parallel <integer>
     Enable parallel simulation with specified number of threads. 0 threads
     means auto-detect number of procssors.
```


peterNordin:
Does this enable parallel optimization ?
I think that was mentioned in the issue description.

robbr48:
Parallel optimization using the Ops algorithms are already supported. I think the issue has external algorithms in mind, e.g. Matlab or Python. This can be achieved by launching multiple hopsancli instances, assuming of course they read from different parameter files and write to different result files.

PR #2004, opened 2020-10-06
Restore PLO v2 format export to use one less num data columns

### LINKED ISSUE #2003: Crash when importing PLO files
**Describe the bug**
Hopsan crashes when exporting data to PLO and then trying to import the same file. I reads number of columns but then increases it by 1 for adding X-vector (time/frequency). However, the X-vector is already included in the number of columns number. As a result, it goes out of range and crashes.

**To Reproduce**
1. Plot something
2. Export to PLO
3. Improt same PLO file
4. The program will crash

**Expected behavior**
It should not crash :unamused: 

Confirmed in Hopsan 2.15, probably related to latest PLO fixes.

**Additional information**
Example of erroneous PLO-file:

```
    'VERSION'
    2
    'plotest.PLO'
    3    2048
    'Time',    'Position_Sensor#out#Value',    'Step#out#Value'
  Time  Position  Position
  0.000000e+00  0.000000e+00  2.000000e-01
  1.500000e-03  6.999712e-09  2.000000e-01
  2.900000e-03  9.386945e-06  2.000000e-01
[...]
```
Example of correct PLO-file:

```
    'VERSION'
    2
    'plotest.PLO'
    2    2048
    'Time',    'Position_Sensor#out#Value',    'Step#out#Value'
  Time  Position  Position
  0.000000e+00  0.000000e+00  2.000000e-01
  1.500000e-03  6.999712e-09  2.000000e-01
  2.900000e-03  9.386945e-06  2.000000e-01
[...]
```

peterNordin:
Resolves #2003

### DIFF
--- a/HopsanCore/src/ComponentUtilities/PLOParser.cpp
+++ b/HopsanCore/src/ComponentUtilities/PLOParser.cpp
@@ -126,15 +126,16 @@ bool PLOParser::readFile(std::iostream &rFileStream)
     }
 
     // Read column names
+    const bool isPloV1orV2 = (mPloVersion == 1) || (mPloVersion == 2);
 
-    mDataNames.reserve(mNumDataCols+1); //One extra in case plo v1 and Time in first column
+    mDataNames.reserve(mNumDataCols+1); // Reserve one extra in case plo v1 or v2 and Time in first column
     for (size_t i=0; i<mNumDataCols; ++i)
     {
         rFileStream >> tmp;
         HString name(tmp.c_str());
 
-        if ( (i==0) && (mPloVersion == 1) && ((name=="'Time',") || (name=="'Frequency',")) )
-        {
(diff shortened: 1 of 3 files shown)


peterNordin:
I downside with this change is that any previous plo v2 file exported that works with the HopsanCore plo parser will now stop working. Fixing the file manually is easy if you know what to do though. (Reduce number of columns with 1).

Maybe it would be possible to add a check after reading the number of columns (the column names) and compare with the specifed number of columns to give a warning message, but I did not try to implement that.

robbr48 (approved):
It seems to work. A solution for backwards compatibility could be to verify that the number of columns equals the specified number. But then one could of course argue that specifying the number of columns is pointless.

peterNordin:
We can look at this again for PLO v3, I have a separate branch to finish that

Issue #1964, opened 2020-07-03
Installation on archlinux


Xwang1976:
**Is your improvement request related to a problem? Please describe.**
I'm trying to create a PKGBUILD to install hopsan natively in archlinux.
I've followed the build procedure, but I've noticed that there isn't any "make install" and that at the end of the building procedure all the executables and libraries are mixed in the hopsan/bin folder.

**Describe the solution you'd like**
I would like to have a configure, make, make install procedure which will create the /usr/bin and /usr/lib folders with executables and libraries correctly separated and maybe with the possibility to install in /opt instead of /usr as some other open source code projects permit to do.

**Describe alternatives you've considered**
I suspect that the debian packages rules require to do something similar when a deb package is created. If so, maybe something similar to what I'm asking for is already available.

**Additional context**

peterNordin:
You may be able to use the copyInstallHopsan.sh script located in the packaging subdirectory.

It copies the files that were written back to the source code directory. And some files that were there to begin with.
It takes two arguments, src and dst directories.
Deb packaging uses this script.

CMake support is a work in progress, once it is finished, you will be able to make install as you would expect.

PR #2006, opened 2020-10-13
Issue 1964 full cmake build and installation

peterNordin:
- Add CMake build support to all Hopsan libraries (Excluding HoLC)
- Add CMake doxygen generator for user and developer (full) documentation without using external scripts
- Enable build with MS Visual C++ Compiler (MSVC) 
  _Note!_ Generator library still does not support building code with MSVC
- Works with GCC, MinGW and MSVC, but not yet Clang for MacOS

Known problems:
- CMake still uses non-standard installation paths (but at least it does not install back into the source code directory)
- CMake build system is still considered experimental, the produced binaries have not been properly tested (but seems to work)
- Debug mode is not supported when using MSVC
- News feed does not work in the MSVC compiled version

Related to #1964


robbr48 (commented):
Compilation works fine, and actually seems to be faster than qmake on Windows. However I am not able to run the executables. They are not moved to the bin folder, and running them from the build directory does not work since it cannot find the required dlls.

peterNordin:
Did you "build" the install target, or run "cmake --build . --target install" 
It will no longer install stuff back into the source code directory, which makes it more complicated to use in QtCreator unfortunately.

I am working on new build instructions for CMake, will add in a separate PR

peterNordin:
I added new "manual" build instructions to this PR

Issue #1801, opened 2018-11-18
Improve Quantity and Unit system


peterNordin:
### Issues with the current implementation
From the comments in PR #1793, the following issues have been identified.

- If no Quantity have been registered, and a unit that is a base-unit for two different quantities is given. The automatic Quantity selection does not work. E.g. Frequency and Angular Velocity both have rad/s as the base unit
- There is no easy way to change base unit for a quantity, then all loaded models must be updated, and likely all model code must also be. The latter is not possible to solve in hopsan. 
- Unit scaling is only known by HopsanGUI, Hopsan core only works with "base-units" (by design). But if unit scaling was known by the core, then maybe changing base unit would not be so complicated
- It would be nice (not sure I agree) to be able to use non-base-units in the component code. Technically you can use whatever unit you like, but it is confusing if the majority of components use un-scaled SI (or derived) base units, and then someone comes along and use something different. Ex: millimeter for a position instead of meter. But if the core new about unit scales, then maybe invalid (unit-wise) connections could be detected and prevented. See for example the FMI 2.0 specification.
- It is unclear how to set a unit or quantity on a variable/parameter in a component. And it is difficult to ensure that a correct value is used since a string is the parameter type. A user can write anything in the string. Example:  Ohm instead of ohm (actually one should write the quantity name Resistance). If Ohm is written it will not be possible change the unit scaling.

Out of battery will update later

MarkusBagge:
Very nice summary. I didn't notice until now you had put it in.

peterNordin:
Yes I was going to tell you about it when I was done editing it, but I never finished it, I will continue to improve and clarify this issue. And add some check boxes with tasks, and document decisions as they are made.

peterNordin:
@robbr48 @MarkusBagge I have updated the issue description as best I can based on earlier discussions. Feel free to suggest improvements or request clarification. I have added my current opinion (which may change) to each decision item. @robbr48 You can add yours when you have time. when/if we agree I can add tasks.  
This will obviously take a very long time to complete, but one step at a time will lead to some improvements. 
Also requesting feedback (as comments) (if you have have time for it) from @petkr14 @magse @marthoch  I believe that you may have valuable input on this sort of thing.

MarkusBagge:
Again, excellent break down! 
Regarding "Component developement/Point 1/Answer 2";
>But maybe the default display value should be possible to choose, maybe in the component appearance XML.

[...]

PR #2016, opened 2020-11-26
Pumps with external leakage

samuelkarnell:
Added pumps with external leakage ports in the extended library. The difference between VariableMachine and VariableMachineW is that the latter has a mechanical port.


robbr48 (changes requested):
- `VariableMachine` only computes the pressures (`p=c+Zc*q`) when there is cavitation. They should always be computed. The second component does this correctly.
- Default parameters for the leakage coefficients should not be zero. It is very hard for the user to figure out which order of magnitude to use for these parameters if there is no default value to compare with.

OwenDavid:
Does the internal and external leakage in these components correspond to the leakage coefficient in the fixed displacement pumps and motors from the standard library?

If so the following issue resolved the leakage units to be m^3/s/pa. 
https://github.com/Hopsan/hopsan/issues/2025

Should the same units be used for the Variable Machines too? I'm not sure what the academic precedent is here but the standard library has components with m^3/s/pa.

*Ideally, there would be one standard leakage unit across all similar components, i.e. displacement pumps/motors.*

OwenDavid:
What are your thoughts on the following proposal regarding a leakage coefficient units converter?

https://github.com/Hopsan/hopsan/issues/2040

samuelkarnell:
@OwenDavid I think it could be a nice feature, and of course the principle can be applied on other parameters as well. However, I'm not sure how the unit conversion is implemented today (it is done outside of the component code), so I'm not really the guy to answer. Anyway, I found a related discussion #1801.

OwenDavid:
I have now answered my own question on this and have a working implementation of scaling the leakage coefficients.

It is very trivial so have a look at the issue again if you are interested.

robbr48 (approved):

PR #2019, opened 2020-12-06
Fix various copy paste and undo stack related bugs

peterNordin:
- Fix various copy paste and undo stack related bug
- Various  cleanup in related GUI code


robbr48 (changes requested):
Good that the bug with getting stuck in drag mouse mode is finally solved. The rest looks good. Only problem I can find is that components appear in the wrong location if pasted while pasting after dragging with left mouse button pressed (rubber band dragging or dragging the workspace). It is related to the feature that moves components only vertically or horizontally if ctrl key is pressed. It can easily be resolved by not allowing pasting when left mouse button is pressed (add it to the check on line 1577 in GUIContainerObject.cpp). I could fix it, but I am not allowed to push to your PR.

peterNordin:
- I added one commit as you suggested.
- Hmm strange that I can push to yours. 
- The hand getting stuck can actually still happen. IT happens if I run HopsanGUI through the debugger and I get stuck at a breakpoint while panning. But I think that ordinary users will not be affected by this.

robbr48 (approved):
Looks good now!
Breakpoints in event-driven GUIs can always cause problems I think. It is not really a bug if you need to use a debugger to trigger it.

PR #2021, opened 2020-12-10
Attempt to fix progress bar related crash

peterNordin:
By not removing the dialog and progress worker object between runs.
Also simplified signal/slot usage

I could not recreate the original crash, so I can not be sure that this works.
But I think that the risk should be much smaller


robbr48 (approved):
The code looks good, but I can't recreate the actual problem so I'm not sure if it is working either.

PR #2024, opened 2020-12-22
Load text files internally in Hopsan with HCOM edit command

robbr48:
HCOM "edit" command shall open supported files in Hopsan text editor instead of external editor (which quite often is not found anyway).

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -5566,7 +5566,15 @@ void HcomHandler::executeEditCommand(const QString cmd)
     QString path = args[0];
     path.remove("\"");
     path.prepend(mPwd+"/");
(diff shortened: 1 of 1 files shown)


HopsanGUI/HcomHandler.cpp
```diff
@@ -5566,7 +5566,15 @@ void HcomHandler::executeEditCommand(const QString cmd)
     QString path = args[0];
     path.remove("\"");
     path.prepend(mPwd+"/");
-    QDesktopServices::openUrl(QUrl::fromLocalFile(path));
+
+    QString suffix = QFileInfo(path).suffix();
+    QStringList textFileSuffixes = QStringList() << "hcom" << "hpp" << "cpp" << "h" << "c" << "cc" << "cci" << "xml";
```
peterNordin, line 5571: Maybe add  "plo", "csv" and "txt"

PR #2032, opened 2021-01-03
Update Inkscape splash screen commands for version 1.0

peterNordin:
- Fixes invalid splash screen in release
- Inkscape 1.0 or newer is now required on Windows
- Adjust alignment for two major version numbers


robbr48 (approved):
Looks good. I may need to convince our IT department to provide Inkscape 1.0, but that is a minor issue.

peterNordin:
Well, its not really required, only if you build a release, and if you do not have it I think it will just not generate the splash screen

PR #2033, opened 2021-01-03
Turn off rotation for various signal icons

### LINKED ISSUE #2026: Default Library : rotation of 'SUM' signal icon
The summation signal icon displays a capital Sigma ( Σ )

Rotation or mirroring of this icon changes the sigma to read as either a 'W', 'M', or '3'.

The 'W' and 'M' cases are particularly clear and can lead to confusion in the signal path.

*Modify icon to prevent rotation of graphic when rotating or mirroring ports.*

peterNordin:
Where the symbol would look odd when rotated or flipped.
Round and square ones, rectangular ones can not be fixed right now.

Resolves #2026


peterNordin:
This one also includes SignalSign which is actually rectangular, but since it is nearly square, rotation looks reasonable anyway.
What do you think @robbr48 ? There are some other rectangular (nearly square) icons that we could turn off rotation for.
Or we could improve the code, but that takes more time.

robbr48 (approved):
Looks good. I changed the sign component icon to a circular shape, to match the other components in the arithmetic subfolder. Then the disabled rotation should never be a problem.

peterNordin:
Yes round makes sense

PR #2034, opened 2021-01-04
Correct units and consistent default values for leakage coefficients

### LINKED ISSUE #2025: Default Library : Q_type_Fixed_Displacement_Motor : missing parameter units
If we open the properties for this object we can see that the leakage coefficient, *c_lm* has no associated units.

If we compare this to the Q type pump, we see that a leakage coefficient unit of *(m^3/s)/Pa*

Experimenting with values in this unit, such as, 10e-12, yields reasonable results in the motor.

*Confirm the units of motor leakage coefficient*
*Add the correct units to the default library object*

robbr48:
- Added missing unit to leakage parameter for all pump/motor components
- Set default value for leakage parameter to 1e-12 for all pump/motor components

Resolves #2025.

### DIFF
--- a/componentLibraries/defaultLibrary/Hydraulic/LinearActuators/HydraulicCylinderC.hpp
+++ b/componentLibraries/defaultLibrary/Hydraulic/LinearActuators/HydraulicCylinderC.hpp
@@ -95,14 +95,14 @@ class HydraulicCylinderC : public ComponentC
             mpP3 = addPowerPort("P3", "NodeMechanic");
 
             // Add input variables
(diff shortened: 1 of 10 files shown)


peterNordin:
I searched for "Leakage coefficient" and also found missing units under linear actuators and machine parts.
Maybe do them as well ?

I was also considering if we need  "Quantities" for these values, but i dont know if it makes sens to change "unit" for a leakage coefficient.

peterNordin (approved):
Approved, but maybe add for the missing LinearActuators and Machine part (piston)

PR #2036, opened 2021-01-04
Remove fade effect on red error flash in terminal

### LINKED ISSUE #2028: Terminal Window : Simulation Error / Red Flash Timing
In the event of a simulation error two things happen,

1. The error message is printed to the terminal in a red font
2. There is a red flash to the terminal background which fades over time.

The combination of the red font and red flash make it impossible to read the error message until the flash has faded. The flash seems that it takes a long time to fade after an error has occurred and you are waiting to read the message...

The ability to read the error message instantly please.

Suggestions,

1.  Reduce time of red flash fade
2.  Change red flash color so the error message font is readable sooner
3.  Consider flashing the simulation background instead with a lighter color
4. Put an error widget on the terminal button strip and animate this

robbr48:
When an error occurs, show a solid red background for 400 ms instead of the fading effect. This should be more robust, instead of updating the background color once every 6 ms for ~1 s, as before. Now we only need to trigger a single event once.

Alternative solution could be to use a more coarse-grained fading effect, where the color is updated less often.

Please try and see what you think.

Resolves #2028.

### DIFF
--- a/HopsanGUI/Widgets/HcomWidget.cpp
+++ b/HopsanGUI/Widgets/HcomWidget.cpp
@@ -239,8 +239,6 @@ TerminalConsole::TerminalConsole(TerminalWidget *pParent)
     mpCompleter->setCaseSensitivity(Qt::CaseInsensitive);
     QObject::connect(mpCompleter, SIGNAL(activated(QString)),
                      this, SLOT(insertCompletion(QString)));
-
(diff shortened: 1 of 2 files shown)


peterNordin (changes requested):
Works for me, I added a code cleanup commit that you can squash into your commit before merging. 
So Approved after squash

PR #2037, opened 2021-01-05
Elites parameter in genetic algorithm

robbr48:
The best individuals are always carried over to the next generation without crossover and mutation. This helps avoid degradation of solution quality between generations.

### DIFF
--- a/HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
+++ b/HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
@@ -169,6 +169,10 @@ OptimizationScriptWizard::OptimizationScriptWizard(SystemObject* pSystem, QWidge
     mpMPLineEdit = new QLineEdit("0.1", this);
     mpMPLineEdit->setValidator(new QDoubleValidator());
 
+    mpElitesLabel = new QLabel("Number of elites: ");
+    mpElitesLineEdit = new QLineEdit("4", this);
+    mpElitesLineEdit->setValidator(new QIntValidator());
+
     mpNumModelsLabel = new QLabel("Number of models: ");
     mpNumModelsLineEdit = new QLineEdit(QString::number(qMax(1,gpConfig->getIntegerSetting(CFG_NUMBEROFTHREADS))), this);
     mpNumModelsLineEdit->setValidator(new QIntValidator());
@@ -246,6 +250,8 @@ OptimizationScriptWizard::OptimizationScriptWizard(SystemObject* pSystem, QWidge
     pSettingsLayout->addWidget(mpCPLineEdit,           row++, 1);
     pSettingsLayout->addWidget(mpMPLabel,              row,   0);
     pSettingsLayout->addWidget(mpMPLineEdit,           row++, 1);
+    pSettingsLayout->addWidget(mpElitesLabel,              row,   0);
+    pSettingsLayout->addWidget(mpElitesLineEdit,           row++, 1);
     pSettingsLayout->addWidget(mpNumModelsLabel,       row,   0);
     pSettingsLayout->addWidget(mpNumModelsLineEdit,    row++, 1);
     pSettingsLayout->addWidget(mpMethodLabel,          row,   0);
@@ -627,6 +633,8 @@ void OptimizationScriptWizard::setAlgorithm(int i)
     mpCRLineEdit->setVisible(false);
     mpCPLabel->setVisible(false);
     mpCPLineEdit->setVisible(false);
+    mpElitesLabel->setVisible(false);
+    mpElitesLineEdit->setVisible(false);
     mpMPLabel->setVisible(false);
     mpMPLineEdit->setVisible(false);
     mpLengthLabel->setVisible(false);
@@ -704,6 +712,8 @@ void OptimizationScriptWizard::setAlgorithm(int i)
         mpCPLineEdit->setVisible(true);
         mpMPLabel->setVisible(true);
         mpMPLineEdit->setVisible(true);
+        mpElitesLabel->setVisible(true);
+        mpElitesLineEdit->setVisible(true);
     case Ops::ParameterSweep:
         break;
     default:
@@ -1172,6 +1182,7 @@ void OptimizationScriptWizard::generateGeneticScript()
 
     templateCode.replace("<<<cp>>>", mpCPLineEdit->text());
     templateCode.replace("<<<mp>>>", mpMPLineEdit->text());
+    templateCode.replace("<<<elites>>>", mpElitesLineEdit->text());
     templateCode.replace("<<<partol>>>", mpEpsilonXLineEdit->text());
     templateCode.replace("<<<nmodels>>>", mpNumModelsLineEdit->text());
 
--- a/Ops/src/OpsWorker.cpp
+++ b/Ops/src/OpsWorker.cpp
@@ -37,6 +37,7 @@
 #include <iostream>
 #include <time.h>
 #include <algorithm>
+#include <random>
 
 //#include <QDebug>
 #include "OpsWorker.h"
@@ -364,7 +365,10 @@ size_t Worker::getCurrentNumberOfIterations()
 
 double Worker::opsRand()
 {
-    return double(rand())/double(RAND_MAX);
+    std::random_device device;
+    std::default_random_engine generator(device());
+    std::uniform_real_distribution<> distribution(0.0, 1.0);
+    return distribution(generator);
 }
 
 bool Worker::aborted()
--- a/Ops/src/OpsWorkerGenetic.cpp
+++ b/Ops/src/OpsWorkerGenetic.cpp
@@ -36,6 +36,9 @@
 #include "OpsMessageHandler.h"
 #include <math.h>
 #include <random>
+#include <iostream>
+#include <algorithm>
+
 
 using namespace Ops;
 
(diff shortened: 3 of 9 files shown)


HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
```diff
@@ -169,6 +169,10 @@ OptimizationScriptWizard::OptimizationScriptWizard(SystemObject* pSystem, QWidge
     mpMPLineEdit = new QLineEdit("0.1", this);
     mpMPLineEdit->setValidator(new QDoubleValidator());
 
+    mpElitesLabel = new QLabel("Number of elites: ");
+    mpElitesLineEdit = new QLineEdit("4", this);
+    mpElitesLineEdit->setValidator(new QDoubleValidator());
```
peterNordin, line 174: DoubleValidator used here, but the number is not a double, should be int ?

Ops/src/OpsWorkerGenetic.cpp
```diff
@@ -71,30 +74,23 @@ void WorkerGenetic::run()
...
     for(; mIterationCounter<mnMaxIterations && !mpMessageHandler->aborted(); ++mIterationCounter)
     {
-        for(size_t i=0; i<mNumCandidates; i+=2) {
+        mCandidatePoints.clear();
+        std::vector<size_t> ids = getIdsSortedFromWorstToBest();
+        for(int i=0; i<mNumElites; ++i) {
+            mCandidatePoints.push_back(mPoints[ids.back()]);
+            ids.pop_back();
+        }
+        while(mCandidatePoints.size() < mNumPoints) {
```
peterNordin, line 83: When selecting parents for crossover, will the elites be used again ? And if that is that case it is a problem or maybe as it should be? I dont remember. Maybe there is a risk for the elites to have to much "say" since both they them selves and their crossover-children will survive in this case ? OR maybe this is indented ?

Ops/src/OpsWorkerGenetic.cpp
```diff
@@ -147,66 +143,91 @@ void WorkerGenetic::setMutationProbability(double value)
...
-    for(size_t i=0; i<mNumPoints; ++i) {
-        double value = mObjectives[i]*opsRand();
-        if(value < bestValue) {
-            mParent1 = i;
-            bestValue = value;
+    std::vector<size_t> sortedIds = getIdsSortedFromWorstToBest();
+
+    // Select parent 1
+    std::vector<size_t> competitors;
+    competitors.push_back(opsRand()*(mNumPoints));
```
peterNordin, line 159: Small (very) risk that: `opsRand()*(mNumPoints)` will be equal to mNumPoints, in which case the competitor (I assume its an index) will be out of bounds. If opsRand() returns "exactly" 1.

peterNordin: Maybe make another help function, that takes an argument for the max allowed value.
But in that case consider the rounding.
In this case the rounding to size_t will always be by truncation. So its unfair since the max value will almost never be reached. (But in the current code that is a good thing to avoid out of bounds)

Ops/src/OpsWorkerGenetic.cpp
```diff
@@ -147,66 +143,91 @@ void WorkerGenetic::setMutationProbability(double value)
...
+    competitors.push_back(opsRand()*(mNumPoints));
+    while(competitors.size() < 2) {
```
peterNordin, line 161: Since there are two push_backs above, this loop will never run

Ops/src/OpsWorkerGenetic.cpp
peterNordin: Aha, just noticed that parents are chosen randomly, so maybe not an issue.

Ops/src/OpsWorkerGenetic.cpp
```diff
@@ -147,66 +143,91 @@ void WorkerGenetic::setMutationProbability(double value)
...
+        competitors.push_back(opsRand()*(mNumPoints));
+    }
+    size_t bestCompetitor = 0;
+    for(auto competitor : competitors) {
```
peterNordin, line 165: For readability, break this out into a help function `findMaxOf(stuff)`   or similar.
It is also used below, or maybe there is a std::algorithm available

Ops/src/OpsWorkerGenetic.cpp
```diff
@@ -147,66 +143,91 @@ void WorkerGenetic::setMutationProbability(double value)
...
-        }
-        else
+        for(size_t i=0; i<mNumParameters; ++i)
         {
+            double x1 = mPoints[mParent1][i];
+            double x2 = mPoints[mParent2][i];
+
             double mean = (x1+x2)/2.0;
-            double dev = fabs(x2-x1)/2.0;
+            double dev = fabs(x2-x1)/10.0;
```
peterNordin, line 206: What is this magic value 10.0 that was previously 2.0

Ops/src/OpsWorkerGenetic.cpp
```diff
@@ -147,66 +143,91 @@ void WorkerGenetic::setMutationProbability(double value)
...
-    mCandidatePoints[mParent1] = mChild1;
-    mCandidatePoints[mParent2] = mChild2;
 }
 
 
 //! @brief Auxiliary function for mutating children
 void WorkerGenetic::mutate()
 {
+    std::vector<size_t> eliteIds = getIdsSortedFromWorstToBest();
+    while(eliteIds.size() > mNumElites) {
```
peterNordin, line 219: Minor issue: OK, I guess this works, but maybe use std::copy instead. To copy the mNumElites last variables, (using iterators)
If you change it, Remember to pre-allocate the destination, or to use std::backinserter

Or maybe use a loop that loops in reverse and copies from "allSortedIds" into eliteIds

Ops/src/OpsWorkerGenetic.cpp
```diff
@@ -147,66 +143,91 @@ void WorkerGenetic::setMutationProbability(double value)
...
     for(size_t i=0; i<mNumPoints; ++i) {
+        if(std::find(eliteIds.begin(),eliteIds.end(),i) != eliteIds.end()) {
+            continue;
+        }
         for(size_t j=0; j<mNumParameters; ++j) {
             double doMutation = opsRand();
             if(doMutation < mMutationProbability) {
                 double mean = mCandidatePoints[i][j];
-                double dev = (mParameterMax[j]-mParameterMin[j])/2.0;
+                double dev = (mParameterMax[j]-mParameterMin[j])/100.0;
```
peterNordin, line 230: Again, unclear (for me) magic value.

peterNordin (changes requested):
Also, does this work with optimization using the CLI or dies that need additional code

Ops/src/OpsWorkerGenetic.cpp
robbr48: There are different ways to implement elitism, but when reading about it online, it seems like the most common approach is that elites should also be able to be selected as parents. There are also different opinions on whether or not the elites should undergo mutation or not. But to begin with we should just stick to the most simple approach I think.

HopsanGUI/Dialogs/OptimizationScriptWizard.cpp
robbr48: Fixed!

Ops/src/OpsWorkerGenetic.cpp
robbr48: I replaced the rand() function with C++11 generator, which generates a value in [0,1[ instead of [0,1]. This should eliminate the risk.

Ops/src/OpsWorkerGenetic.cpp
robbr48: Not sure where the 2 comes from. I changed it so both parents are chosen in the same way.

Ops/src/OpsWorkerGenetic.cpp
robbr48: Replaced with the `max_element` function.

Ops/src/OpsWorkerGenetic.cpp
robbr48: It is the standard deviation of the normal distribution used for crossover. It gives a slightly higher chance of producing offspring close to the average value of the parents. The value 10 just makes it a little narrower. Perhaps it should be a tunable parameter, but it would be quite tricky for the user to find a decent value anyway.

Ops/src/OpsWorkerGenetic.cpp
robbr48: I changed it to a simpler solution by just creating a new vector in the range `[sortedIds.end()-mNumElites, sortedIds.end()]`. Maybe not the most efficient code, but better than the previous and quite readable I think.

Ops/src/OpsWorkerGenetic.cpp
robbr48: Same as above, it makes the distribution for mutation more narrow around the average value.

robbr48:
You are right, the CLI was not supported. I added an `elites` argument.

Ops/src/OpsWorker.cpp
```diff
@@ -364,7 +365,11 @@ size_t Worker::getCurrentNumberOfIterations()
 
 double Worker::opsRand()
 {
-    return double(rand())/double(RAND_MAX);
+    std::random_device device;
+    std::mt19937 generator(device());
+    std::uniform_real_distribution<> distribution(0.0, 1.0);
+    std::cout << distribution(generator);
```
peterNordin, line 371: leftover debug cout

Ops/src/OpsWorker.cpp
```diff
@@ -364,7 +365,11 @@ size_t Worker::getCurrentNumberOfIterations()
 
 double Worker::opsRand()
 {
-    return double(rand())/double(RAND_MAX);
+    std::random_device device;
+    std::mt19937 generator(device());
```
peterNordin, line 369: Any particular reason for using the mt1993 generator? (I dont know anything about these things), but there is also:
`std::default_random_engine generator()`, slightly simpler code, but does not really matter.

Ops/src/OpsWorkerGenetic.cpp
peterNordin: hmm, yes it should probably be tunable, but I guess its not critical right now

Ops/src/OpsWorkerGenetic.cpp
peterNordin: I think this solution is the same as what I intended but much simpler

Ops/src/OpsWorkerGenetic.cpp
peterNordin: OK, but should probably be tunable

Ops/src/OpsWorker.cpp
robbr48: I changed it do the default engine. I got the code from a suggestion on some forum. I have no idea which one is best, but for our purposes they should all be good enough I think.

peterNordin (approved):

robbr48:
@peterNordin any ideas why the ghactions tests are failing? It fails to download Qt for some reason.

peterNordin:
Yes, the Qt download servers crashed or something. If appveyor is green in this case, go ahead and merge

PR #2049, opened 2021-01-15
Support DCP standard

robbr48:
Support for the [Distributed Co-simulation Protocol (DCP)](https://dcp-standard.org/) standard. This enables communication between Hopsan and other DCP supporting tools, as well as with physical devices. Hopsan is hard-coded to use UDP protocol for now, but TCP/IP can alsobe supported in the future as well as CAN and Bluetooth.

It is possible to both export Hopsan models and simulate them in other tools, as well as importing and simulating models from other tools. Due to the lack of open source DCP tools, I have only been able to test simulations between Hopsan and Hopsan.

The implementation is based on the [DCPLib](https://github.com/modelica/DCPLib) subproject, which in turn requires three new dependencies; Asio, Xerces and libzip.


dependencies/dependencies.xml
```diff
@@ -87,6 +87,24 @@
     </releasefile>
   </dependency>
 
+  <dependency name="asio">
+    <releasefile sha1="bac61183b633a6d91ef7ae31836434bdf4f2b0d0">
+      <url>https://sourceforge.net/projects/asio/files/asio/1.12.1 (Stable)/asio-1.12.1.zip</url>
```
peterNordin, line 92: Are the exact versions listed here required? Otherwise it would be "better"  to install them from the package repository on Linux systems. But i guess having them as a fall-back in-case system is to old might make sense.

More use full would be the downloads (and build scripts) for Windows

dependencies/setupAsio.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+# $Id$
```
peterNordin, line 2: We should stop adding `# $Id$` snf remove it when copy pasting

dependencies/setupAsio.sh
```diff
@@ -0,0 +1,27 @@
...
+./download-dependencies.py ${name}
+
+# Copy code to build dir, not sure if out-of-source build is possible
+mkdir -p $builddir
+pushd $builddir
+cp -a $codedir/* .
+
+# Generate makefiles
+chmod u+x ./configure
+./configure --prefix=$installdir --without-boost
```
peterNordin, line 22: This (configure) will be problematic on Windows, I hope asio and the others have cmake support
I have just recently been able to get rid of the msys2 build dependency, I do not want it back.

dependencies/setupLibzip.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Include general settings
+source setHopsanBuildPaths.sh
+
+# Create build dir and enter it
+mkdir -p $builddir
+cd $builddir
+
+# Generate makefiles
+cmake -Wno-dev -DCMAKE_INSTALL_PREFIX=$installdir ${codedir}
```
peterNordin, line 23: cmake good :)

hopsandcp/dcpmaster.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include "dcp/model/pdu/DcpPduFactory.hpp"
+#include "dcp/xml/DcpSlaveDescriptionReader.hpp"
+#include "dcp/driver/ethernet/udp/UdpDriver.hpp"
+#include "dcp/logic/DcpManagerMaster.hpp"
+
+
+DcpMaster::DcpMaster(const std::string host, u_short port)
+    : stdLog(std::cout)
+{
+    driver = new UdpDriver(host, port);
```
peterNordin, line 18: Maybe should have called this mpDriver and mpManager, but not such a big deal

hopsandcp/dcpmaster.cpp
```diff
@@ -0,0 +1,259 @@
...
+    connection.fromVr = fromVr;
+    connection.toVrs = toVrs;
+    connection.toSlaves = toSlaves;
+    connections.push_back(connection);
+}
+
+void DcpMaster::start() {
+    std::thread b(&DcpManagerMaster::start, manager);
+    std::chrono::seconds dura(1);
+    std::this_thread::sleep_for(dura);
```
peterNordin, line 68: Sleeping random time is not safe, sometimes its enough sometime not. Is there some other way to "wait long enough"

hopsandcp/dcpmaster.cpp
```diff
@@ -0,0 +1,259 @@
...
+    manager->addLogListener(std::bind(&OstreamLog::logOstream, stdLog, std::placeholders::_1));
+    manager->setGenerateLogString(true);
+}
+
+DcpMaster::~DcpMaster() {
+    delete driver;
+    delete manager;
+}
+
+void DcpMaster::addSlave(string &filepath)
```
peterNordin, line 41: const string&, unless filepath is input/output (then call it rFilepath)

hopsandcp/dcpmaster.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include <memory>
+
+#include "dcp/helper/LogHelper.hpp"
+#include "dcp/model/pdu/DcpPduFactory.hpp"
+#include "dcp/xml/DcpSlaveDescriptionReader.hpp"
+#include "dcp/driver/ethernet/udp/UdpDriver.hpp"
+#include "dcp/logic/DcpManagerMaster.hpp"
+
+
+DcpMaster::DcpMaster(const std::string host, u_short port)
```
peterNordin, line 15: Calling stuff Master/Slave is discouraged nowdays. But since they are called master/slave in the dcp library I understand why you did. Not sure if we should call them something else or just leave the names.
Maybe Master/Worker, not sure

hopsandcp/dcpslave.cpp
```diff
@@ -0,0 +1,244 @@
+#include "dcpslave.h"
+#include "utilities.h"
+
+#include <dcp/helper/Helper.hpp>
```
peterNordin, line 4: If dcp is bundled inside Hopsan, then maybe use #include "path/here"  instead of < >, not a big deal

hopsandcp/utilities.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+ The full license is available in the file LICENSE.
+ For details about the 'Hopsan Group' or information about Authors and
+ Contributors see the HOPSANGROUP and AUTHORS files that are located in
+ the Hopsan source code root directory.
+
+-----------------------------------------------------------------------------*/
+
+//!
+//! @file   utilities.cpp
```
peterNordin, line 26: we really need a common library for all of these functions

peterNordin:
Made some comments, but have not had time to test, have no obvious ideas, seems like a reasonable approach

hopsandcp/dcpmaster.cpp
robbr48: You are right, I actually did not hear of this before. It is not only the DCP library that uses it. This terminology is very strong in the co-simulation field. The DCP standard uses the word "slave" 497 times in their specification, and the FMI standard 130 times. But I guess we should call it something else. I don't really like "worker", since it is actually a connection to an something external. Could we call it "DCPServer"? Or maybe "DCPConnection"?

robbr48: Apparantly "Master" is also discouraged, but I guess it depends more on the context. We could call it "DCPManager" instead, which is actually more accurate description of what it does.

robbr48:
Just so you know, I am going to remove the command line interface and convert hopsandcp into a library. Then the gui and cli can link against this library directly. The important thing is that hopsancore should not depend on DCPLib.

hopsandcp/utilities.cpp
robbr48: That would be great. But like I mentioned below, I will take away the CLI functionality from hopsandcp anyway.

hopsandcp/dcpmaster.cpp
peterNordin: Unfortunately I think it would be best to keep the master/slave names to avoid confusion. If out classes match DCP master/slave classes. Otherwise rename them. But yes maybe Manager/Worker, I dont know.

hopsandcp/dcpmaster.cpp
robbr48: I don't like this either but can't think of a better solution. This code snippet is taken directly from the examples provided by DCPLib.

dependencies/dependencies.xml
robbr48: I found out that Asio does not need to be compiled, it can be used as header-only. This makes things much easier.

Issue #2023, opened 2020-12-21
Matlab/Julia/Python: Improved data logging functions


robbr48:
**Is your improvement request related to a problem? Please describe.**
Right now data variables can only be fetched from hopsanc interfaces (Matlab/Julia/Python) through the `getDataVector` function, which works for only one variable. This is not very nice if you want to log all variables from the simulation.

**Describe the solution you'd like**
New functions:
- `getAllDataVectors`
- `disableLogging`
- `enableLogging`

The two latter functions should enable or disable logging for specified variables, preferably using wildcards.

OwenDavid:
While using the python API I have noticed that the variable port names are hard to determine correctly.

For example, if we use the getDataVector() function incorrectly, the error states,

>Component name, port name and variable name must be specified.

If the component name is incorrect then all the components in the model are printed for the user. This is useful.

An improvement would be that in the event that the port is incorrect that all the ports for the component are listed. Same with the variable.

In other words, if the user enters an incorrect component - list all components. If the user enters a correct component but incorrect port, list the ports for the correct component, and if the component and port are correct but the variable is not, then list all the available variables for the given component port.

PR #2050, opened 2021-01-15
List alternatives when quering for illegal ports or variables

robbr48:
Provide a list of legal port names or variables names if an illegal name is provided to hopsanc API. This already works for component names.

Related to #2023.


peterNordin (approved):
Did not test, but looks OK

PR #2051, opened 2021-01-15
Hydraulic differential pressure sensor

### LINKED ISSUE #2030: New Component: Differential Pressure Gauge
A differential pressure gauge is a useful real world tool to identify the pressure drop across a point of interest in the system.

Could we implement a new object with two pressure inputs and simple code to return the difference or perhaps abs(difference)?

This would prevent the use of several signal blocks and a scope to calculate the same.

Otherwise, would it be trivial to add this as an output variable to relevant objects in the library (restrictors, orifices, etc) so that it would feature in the plot list or would it increase the computation cost for all simulations with only a marginal return, i.e. if the output was used/plotted?

**Additional context**

robbr48:
Differential pressure component with two ports. Outputs the difference between the ports (P1.p - P2.p). 

Created by @OwenDavid.
Resolves #2030


componentLibraries/defaultLibrary/Hydraulic/Sensors/HydraulicDifferentialPressureSensor.hpp
```diff
@@ -0,0 +1,83 @@
...
+//! @date   5/1/2021
+//!
+//! @brief Contains a Hydraulic Differential Pressure Sensor Component
+//!
+//$Id$
+
+#ifndef HYDRAULICDIFFERENTIALPRESSURESENSOR_HPP_INCLUDED
+#define HYDRAULICDIFFERENTIALPRESSURESENSOR_HPP_INCLUDED
+
+#include <iostream>
```
peterNordin, line 37: iostream not needed, (but does not really matter)

peterNordin (approved):

peterNordin:
I added linked to issue #2030, I assume that is the correct one

PR #2052, opened 2021-01-20
Replace (m^3/s)/Pa with quantity LeakageCoefficient

### LINKED ISSUE #2040: Standard Library Components: Leakage Coefficient Unit Conversion
In the GUI some component parameter windows can access a unit converter. For example, for we can set a flow rate in l/min or m^3/s.

Could a new converter be added for leakage coefficients?

**Proposed Leakage Units**
*Default - m^3/s/pa (matching existing components)*
*Alternative - m^5/Ns*
*Realworld/datasheets - l/min/bar*

If I am right in thinking that the default and alternative are equivalent then the conversion factor would be 1. The change of name would accommodate the user familiar with these units.

The conversion factor for l/min/bar:

1 l/min =1.66667e-5 m^3/s
1 bar = 1e5 Pa

1 m^3/s/Pa = 1.66667e-5 / (1/1e5)

1 / 1.66667e-10 l/min/bar = 5999,988e6 l/min/bar

*finally, 1 m^3/s/Pa = 5999,988e6 l/min/bar*

As a test, a typical motor leakage coefficient of 1.067e-12 m^3/s/Pa = 0.0064 l/min/bar, so at 100 bar, this gives 0.64 l/min.

This would be relevant to many of the hydraulic components, i.e. pumps, motors, linear actuators, and hydraulic machines and assist users with industrial backgrounds or those referring to datasheets.

peterNordin:
And add scale factors GUI

Resolves #2040

### DIFF
--- a/componentLibraries/defaultLibrary/Hydraulic/Valves/PressureControlValves/HydraulicOverCenterValve.hpp
+++ b/componentLibraries/defaultLibrary/Hydraulic/Valves/PressureControlValves/HydraulicOverCenterValve.hpp
@@ -80,8 +80,8 @@ namespace hopsan {
             addOutputVariable("xv", "Equivalent spool position", "", &mpXv);
 
             addConstant("tao", "Time Constant of Spool", "s", 0.01, mTao);
-            addConstant("k_cs", "Steady State Characteristic due to Spring", "(m^3/s)/Pa", 0.00000001, mKcs);
-            addConstant("k_cf", "Steady State Characteristic due to Flow Forces", "(m^3/s)/Pa", 0.00000001, mKcf);
+            addConstant("k_cs", "Steady State Characteristic due to Spring", "LeakageCoefficient", 0.00000001, mKcs);
+            addConstant("k_cf", "Steady State Characteristic due to Flow Forces", "LeakageCoefficient", 0.00000001, mKcf);
             addConstant("q_nom", "Flow with Fully Open Valve and pressure drop p_nom", "m^3/s", 0.001, mQnom);
             addConstant("p_nom", "Nominal pressure drop", "Pa", 7e6, mPnom);
         }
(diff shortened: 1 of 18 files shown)


componentLibraries/defaultLibrary/Hydraulic/Valves/PressureControlValves/HydraulicOverCenterValve.hpp
```diff
@@ -80,8 +80,8 @@ namespace hopsan {
             addOutputVariable("xv", "Equivalent spool position", "", &mpXv);
 
             addConstant("tao", "Time Constant of Spool", "s", 0.01, mTao);
-            addConstant("k_cs", "Steady State Characteristic due to Spring", "(m^3/s)/Pa", 0.00000001, mKcs);
```
peterNordin, line 83: I am not sure if it makes sense to call it LeakageCoefficient in these cases (valves), but since the unit is the same that is what it will get labeled as anyway. What do you thing @robbr48

robbr48: I am not really sure what quantity this parameter should have. It is technically behaving as a leakage, but the pressure in these parameters might not always be the pressure drop. 

Actually I think unit conversion should only work when you specify a quantity. If a specific unit is given, it should not automatically try to identify the quantity. That makes sense only for trivial units and quantities. I know we discussed this before, with "rad/s" being both a frequency and angular velocity unit. But I guess this would require a significant amount of re-coding.

robbr48 (approved):
Looks good. I think there are more similar quantities we can add later.

componentLibraries/defaultLibrary/Hydraulic/Valves/PressureControlValves/HydraulicOverCenterValve.hpp
peterNordin: Turning of the automatic lookup of quantity should be fairly easy, but it has to be done in a couple of places.
Maybe later....

PR #2054, opened 2021-01-29
Reconfigurable FMI wrapper component

robbr48:
- New `reconfigure` function in Component class
- New `FMIWrapper` component
  - Loads any FMU with a string parameter
  - Adds ports and parameters based on `modelDescription.xml`
  - Graphical ports are autoplaced by HopsanGUI
  - Temporary folders are automatically created and deleted when component is configured/reconfigured/deconfigured
- For simplicity I added C++17 support, not sure which complications this may have

I placed the FMIWrapper component in extentionLibrary for now. It should be tested more before we move it to the defaultLibrary. Then this component can replace the existing FMI import features. The main benefit is that we don't need to recompile anything when importing an FMU. This makes import a lot faster, and you don't need a compiler.


Common.prf
```diff
@@ -18,8 +18,8 @@ CONFIG(release, debug|release) {
   QMAKE_CXXFLAGS += -O3
 }
 
-# Enable C++14
-CONFIG += c++14
+# Enable C++17
```
peterNordin, line 21: The problem with c++17 is that it wont work on older compilers, typically available in older Linux OS versions.
Also not available in the version of MinGW we are currently shipping on windows.
Either the entire FMIWraper must be optional depending on available C++ version, and the build system must choose automatically. Or the functionality must be rewritten with nasty ifdefs and either posix or MSC runtime calls.

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
```diff
@@ -0,0 +1,355 @@
...
+        callbacks.logger = jmLogger;
+        callbacks.log_level = jm_log_level_debug;
+        callbacks.context = static_cast<jm_voidp>(this);
+
+        context = fmi_import_allocate_context(&callbacks);
+
+        addInfoMessage("Loading FMU from "+mFmuPath+"...");
+
+        //Generate random temporary path
+        mTmpPath = std::filesystem::temp_directory_path().c_str();
```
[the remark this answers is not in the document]
peterNordin: Since only C++17 is used to crate and remove a temp directory, maybe it would not be so difficult to use "native" C code for that instead. Possibly a new utility in HopsanCore that wraps such functions, (it could internally use ifdefs to check for c++17).
I actually recently added "dirent" a MSVC implementation third party header-only library, that is used in the RemoteHopsan libraries to enable dirent based filesystem access also with MSVC. 
Of course lookup of temp directory must be made with native calls.

Not sure whats easier, to add ifdefs with native code (Windows, Posix and MacOS(might work with posix)) directly. Or add a HopsanCore utility which includes dirent for MSVC.

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
```diff
@@ -0,0 +1,355 @@
...
+        std::random_device rd;
+        std::default_random_engine gen(rd());
```
peterNordin, line 181: This may also be a problem, is it available pre c++11 (I did not check)?
HopsanCore "must" be compilable with C++03 or the feature in this case must be optional, maybe could use old-style c random generator if pre c++11.

componentLibraries/extensionLibrary/extensionLibrary.xml
```diff
@@ -28,6 +28,8 @@
   <component>Hydraulic/HydraulicVariableVolumeLength.hpp</component>
   <component>Hydraulic/HydraulicVariableVolumeLengthMP.hpp</component>
   <component>Electrical/ElectricInductanceC.hpp</component>
+  <component>Connectivity/FMIWrapper.hpp</component>
+  <component>Signal/FibonacciNumbers.hpp</component>
```
peterNordin, line 32: I guess Fibanatcci was some component that were already present or is it by mistake ?

peterNordin (changes requested):
I know my comments are vague, but I fell like this one should wait until after 2.17 is released. 
And c++17 should not be used. In components c++03 is required unfortunately, unless we can make it an optional component, that you get when your compiler is new enough.

HopsanGUI/CoreAccess.cpp
```diff
@@ -475,6 +476,8 @@ bool CoreSystemAccess::setParameterValue(QString componentName, QString paramete
 {
     hopsan::Component *pComponent = mpCoreComponentSystem->getSubComponent(componentName.toStdString().c_str());
     bool retval = pComponent->setParameterValue(parameterName.toStdString().c_str(), value.toStdString().c_str(), force);
+    pComponent->reconfigure();
```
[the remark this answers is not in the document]
robbr48: Actually I want it to be called also when loading. If you load a model with an existing FMI component, it should reconfigure the component properly to make sure all ports and parameters exist. We could add some information to the CAF file so that only certain parameters calls reconfigure, but won't that introduce even more overhead? Calling an empty function is not so time consuming. And the components can have an internal check to make sure they only perform a reconfiguration when a certain critical parameter has changed.

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
```diff
@@ -1096,6 +1096,9 @@ bool VariableTableWidget::setStartValues()
             // Parameter has changed, add to undo stack and set the parameter
             if( cleanAndVerifyParameterValue(value, qobject_cast<ParameterValueSelectionWidget*>(cellWidget(row, int(VariableTableWidget::Value)))->getDataType()) )
             {
+                if(!mpModelObject->hasParameter(name)) {
```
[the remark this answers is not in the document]
robbr48: Reconfiguration can add or remove parameters. If a reconfiguration that is triggered from within the properties dialog removes a parameter, the dialog will still try to set the parameter value. I don't remember now if this caused an endless iteration or a crash, but it needs to be handled in either case.

HopsanGUI/GUIConnector.cpp
```diff
@@ -1002,9 +1002,9 @@ void Connector::setUnHovered()
 
 
 //! @brief Asks the parent system to delete the connector
-void Connector::deleteMe(UndoStatusEnumT undo)
+void Connector::deleteMe(UndoStatusEnumT undo, bool force)
```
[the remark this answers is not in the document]
robbr48: Yes, this should be used with caution. I have not seen any problems in this case. If we don't want to reconfigure directly from the properties dialog we would need some `reconfigureLater()` function, that runs directly after the parameter is changed. This would on the other hand make the "Apply" button useless for the reconfiguration parameter.

HopsanGUI/GUIObjects/GUIComponent.cpp
```diff
@@ -139,8 +139,60 @@ QString Component::getTypeCQS() const
 //! @brief Set a parameter value to be mapped to a System parameter
 bool Component::setParameterValue(QString name, QString value, bool force)
 {
+    if(this->getTypeName() == "FMIWrapper" && name == "path") {
```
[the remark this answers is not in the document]
robbr48: I'm thinking of a more general solution where a component can have different `<configuration>` tags in the CAF file. Then it would be possible to switch between them using a drop-down menu. This could example change a pump between fixed and variable, or enable/disable friction/leakage etc. But the FMIWrapper is still a special case, since we can't know in advance which parameters and ports we want. We need to first parse the FMI model description and then auto place the ports.

HopsanGUI/GUIObjects/GUIComponent.cpp
```diff
@@ -139,8 +139,60 @@ QString Component::getTypeCQS() const
...
+                inputs << port;
+            }
+            else if("WritePortType" == type) {
+                outputs << port;
+            }
+        }
+
+        //Auto-place input ports
+        for(int i=0; i<inputs.size(); ++i) {
+            SharedPortAppearanceT app(new PortAppearance());
```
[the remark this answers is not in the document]
robbr48: You are right, there is a `refreshExternalPortsAppearanceAndPosition()` function. It is only available in `SystemObject` though. Perhaps we can move this function to `ModelObject` and use it also for FMI components. I will have a look.

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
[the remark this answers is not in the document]
robbr48: FMI import always requres a temp directory, where the FMU is unzipped by FMILibrary. The previous FMI import generates and compiles a new component for each imported FMU. The temp folder is then created directly in the folder for that component library. In the new version a generic component is used for all FMUs. Therefore every component instance needs to create its own temp directory.

We can probably use C++03 to create the temp folder, only drawback is that it requires far more code. The only problem is that Linux systems have no API function for getting the temp directory. The existing functions in C++17 and Boost checks some envirnoment variables (like "TEMPDIR", "TEMPPATH", "TEMP" and so on), and fallbacks to "/tmp" if nothing else is found. Perhaps we can just always assume "/tmp" on Linux systems.

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
```diff
@@ -0,0 +1,355 @@
...
+        if (status == jm_status_error) {
+            stopSimulation("Failed to load .dll/.so file (error: "+HString(fmi2_import_get_last_error(fmu))+")");
+            return;
+        }
+
+        addDebugMessage("Successfully loaded .dll/.so file");
+
+        //Instantiate FMU
+        HString instanceName = getName();
+        fmi2_string_t fmuResourceLocation = "file:///home/robbr48/Documents/Hopsan/import/FMU/hopsan_tutorial/resources";
```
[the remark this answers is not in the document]
robbr48: Ops, that needs to change. It should probably point to the resource folder in the temp directory for the FMU.

Common.prf
robbr48: We can stick with old C++ code then I guess. I just need to replace the temp directory stuff and some loops and auto-variables.

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
robbr48: Unfortunately not, I think it is actually C++14. We can use `rand()` instead. `rand() % 1000000` should generate a random number between 0 and 999999, which is what we need for a six-digit random number in the temp directory. I believe this is sufficient to avoid name collisions.

componentLibraries/extensionLibrary/extensionLibrary.xml
robbr48: Right, my bad. I made that component first only to test the reconfiguration stuff with something simple.

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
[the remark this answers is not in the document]
robbr48: I added a utility class for temp folders. It creates a temp folder in constructor and removes in in the destructor, using C++17 if available, otherwise C++03. It has only two member functions, one for fetching the path and one for checking if the temp directory is valid.

I first thought about having a general file manipulation utility. But then I realized that a function for `deleteFolderAndAllSubfoldersRecursively` is quite of dangerous. The new utility class will only delete the same folder it created, which is much safer.

HopsanGUI/CoreAccess.cpp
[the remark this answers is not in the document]
peterNordin: "Tagging" a parameter must be in the core, so that it can reconfigure when run from CLI or as an exported FMU or similar, then CAF files are not used.

In case a component have reconfigure logic, it will run every time you change a parameter, or in this case, every time you close a properties dialog, once for each parameter and once for each start value. (Unless the dialog box ignores values not changed, but I do not think so).

Currently reconfigure is only handled by the GUI, it would be better if the core could decide when to call it maybe. Then applications would not need extra logic to decide.

In the load case, it could be called once for each component after load is complete.

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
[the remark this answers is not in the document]
peterNordin: OK, makes sense

HopsanGUI/GUIConnector.cpp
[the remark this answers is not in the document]
peterNordin: Hmm I guess force delete is acceptable for connectors.
But I dont know what happens if you have multiple properties boxes open for different components, and then  a connector connecting them disappears.  Likely the startvalue case is handled by the other check (haveParameter).

And what happens if you have the port location editor open. Not sure if can handle removed ports, must test that.

HopsanGUI/GUIObjects/GUIComponent.cpp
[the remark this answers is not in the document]
peterNordin: Reconfiguration would need to be related to the core somehow, so that it works in non GUI applications.
IS the "reconfigure" parameter a special parameter, in that case we could add it as a new parameter type. (then we would get automatic "tagging" that the parameter would need to trigger reconfigure. Then if the CAF have multiple configuration options listed, then those only represent the graphics (port placements and such) for the cases already implemented inside the component and selectable from this "special parameter" similar to the one called "conditional" now ?

HopsanGUI/GUIObjects/GUIComponent.cpp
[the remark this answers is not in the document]
peterNordin: yes or we can add it later

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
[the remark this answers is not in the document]
peterNordin: yes that would have been unsafe, did you check the env variables, before assuming /tmp ?

Common.prf
peterNordin: Yes for all components we should assume that it may run on old system.

componentLibraries/extensionLibrary/CMakeLists.txt
```diff
@@ -8,13 +8,15 @@ endif()
 
 set(targetname extensionlibrary)
 
+include(${CMAKE_CURRENT_LIST_DIR}/../../dependencies/fmilibrary.cmake)
```
peterNordin, line 11: Is it possible to recompile extensionlibrary inside HopsanGUI ? otherwise you need to add bunch of compilation flags to the extensionlibrary.xml. Shold be documented somewhere how to do that.

componentLibraries/extensionLibrary/extensionLibrary.xml
```diff
@@ -18,6 +18,8 @@
     <cflags>-I./someLib/include</cflags>
     <lflags>-L./someLib/bin -lsomeLib</lflags>
     <lflags os="win64">-lWs2_32</lflags> -->
+	<cflags>-I../../dependencies/fmilibrary/include</cflags>
```
peterNordin, line 21: Hmm a bit non-portable. But I guess there is no easy fix for that. We would need to know the HopsanInstallation directory automatically somehow.

HopsanCore/include/ComponentUtilities/TempDirectoryHandle.h
```diff
@@ -0,0 +1,61 @@
...
+ The full license is available in the file LICENSE.
+ For details about the 'Hopsan Group' or information about Authors and
+ Contributors see the HOPSANGROUP and AUTHORS files that are located in
+ the Hopsan source code root directory.
+
+-----------------------------------------------------------------------------*/
+
+//!
+//! @file   SecondOrderTransferFunction.h
+//! @author Björn Eriksson <bjorn.eriksson@liu.se>
```
peterNordin, line 27: Copy-paste file header (not sure if we even need these headers any longer, at least not adding date and such.
At least not for internal core files. It makes sense for components.

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+#if __cplusplus >= 201703L
+        mPath = std::filesystem::create_directory(mPath.c_str());
+#else
+        mIsValid = createDirectory(mPath);
+#endif
+}
+
+TempDirectoryHandle::~TempDirectoryHandle()
+{
+    std::cout << "Test\n";
```
peterNordin, line 82: test

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+
+HString TempDirectoryHandle::getTempDirectory() {
+#if __cplusplus >= 201703L
+        mPath = std::filesystem::temp_directory_path().c_str();
+#else
+#ifdef _WIN32
+    char tempdirbuff[MAX_PATH+1];
+    GetTempPathA(MAX_PATH+1, tempdirbuff);
+    return HString(tempdirbuff);
```
peterNordin, line 104: should this set mPath and mIsValid ?

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+        return val;
+    }
+    return "/tmp";
+#endif
+#endif
+}
+
+HString TempDirectoryHandle::generateRandomNumericString()
+{
+#if false//__cplusplus >= 201103L
```
peterNordin, line 129: if false or if c++14 (or 11)

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+        std::random_device rd;
+        std::default_random_engine gen(rd());
```
peterNordin, line 131: Dont thing you need to pass random_device in the constructor to default_random_engine. 
Not a big deal

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+#endif
+        std::stringstream ss;
+        ss << std::setw(6) << std::setfill('0') << random;
+        return ss.str().c_str();
+}
+
+bool TempDirectoryHandle::createDirectory(HString path)
+{
+#if __cplusplus >= 201703L
+        return std::filesystem::create_directory(mPath.c_str());
```
peterNordin, line 145: should it use path or mPath ?

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+#else
+    HString currentLevel = "";
+    std::string level;
+    if(!path.empty() && path[0] == '/') {
+        //path.erase(0,1);
```
peterNordin, line 150: commended code

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+#endif
+
+        currentLevel += "/"; // don't forget to append a slash
+    }
+#endif
+    return folderExists(path);
+}
+
+
+bool TempDirectoryHandle::folderExists(HString &foldername)
```
peterNordin, line 177: Please call it directory instead of "folder". It feels so MS Windows

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+        return std::filesystem::create_directory(mPath.c_str());
+#else
+    HString currentLevel = "";
+    std::string level;
+    if(!path.empty() && path[0] == '/') {
+        //path.erase(0,1);
+    }
+    std::stringstream ss(path.c_str());
+
+    while(std::getline(ss, level, '/')) {
```
peterNordin, line 154: Will / work on windows, or do you need to check for \

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+        if(!currentLevel.empty()) {
+            if (!folderExists(currentLevel)) {
+                if(mkdir(currentLevel.c_str(), 0755) != 0) {
+                    return false;
+                }
+            }
+        }
+#endif
+
+        currentLevel += "/"; // don't forget to append a slash
```
peterNordin, line 170: does it work on windows or is \\ needed

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+    struct stat st;
+    stat(foldername.c_str(), &st);
+    return st.st_mode & S_IFDIR;
+}
+
+bool TempDirectoryHandle::removeDirectory(HString path)
+{
+    std::cout << "Trying to remove directory: " << path.c_str() << "\n";
+#if __cplusplus >= 201703L
+    std::filesystem::remove_all(mTmpPath.c_str());
```
peterNordin, line 188: path instead of mTmpPath ?

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+    std::wstring stemp = std::wstring(tempStr.begin(), tempStr.end());
+    StringCchCopyW(szDir, MAX_PATH,STRSAFE_LPCWSTR(stemp.c_str()));
+    int len = lstrlenW(szDir);
+    szDir[len+1] = 0; // Null termination
+
+    //Delete the folder including files and subfolders
+    fos.wFunc = FO_DELETE;
+    fos.pFrom = szDir;
+    fos.fFlags = FOF_NO_UI;
+    return (0 == SHFileOperationW( &fos ));
```
peterNordin, line 203: Wow, wonderful Windows code :(

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+#else
+    DIR *dir = opendir(path.c_str());
+    size_t path_len = path.size();
+    bool success = false;
+
+    if(dir) {
+        std::cout << "Found directory: " << path.c_str() << "\n";
```
peterNordin, line 211: This and in other places, debug putput

HopsanGUI/GUIConnector.cpp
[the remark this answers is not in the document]
robbr48: A solution could be to make the connector broken instead of removing it? Port location editor could still be a problem though.

HopsanGUI/GUIObjects/GUIComponent.cpp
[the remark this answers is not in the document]
robbr48: Reconfiguration can be triggered by different type of parameters I would say. It could be a string for a FMU file, an enumerable for changing between configurations or an integer for choosing e.g. number of outputs. We will need a tag in the general parameter class I think, with some `doesTriggerReconfiguration()` function.

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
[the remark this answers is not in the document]
robbr48: No, I'm not sure if it was necessary. I did some research on this, and some people suggest only assuming "/tmp" is enough. I guess the folder always exists on Linux systems, and the only reason to use the environment variables is if you really want to put temp files somewhere else. But if you think it is needed I can add it?

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
robbr48: I have tested it on Windows and confirmed that it works.

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
robbr48: No I prefer to delete them to make the code cleaner. We can easily add new outputs later if needed.

componentLibraries/extensionLibrary/extensionLibrary.xml
robbr48: We should use the $HOPSANHOME environment variable in such cases I think.  But this will do for now.

HopsanCore/include/ComponentUtilities/TempDirectoryHandle.h
robbr48: I corrected it for now. Maybe we can remove these headers later.

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
robbr48: I changed to `path`, as was intended.

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
robbr48: I think I copied it from an example I found somewhere, but it doesn't matter I guess.

HopsanGUI/CoreAccess.cpp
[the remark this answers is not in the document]
robbr48: I added an  "is reconfigurable" flag to `Parameter` class, with an `addReconfigurationParameter()` method in the component API. This way, each component can notify the core that a parameter triggers a reconfiguration. All reconfiguration except graphical appearance is then handled by the core.

Perhaps the name of the function should be changed, as it does not actually add the parameter. You need to first add it using `addConstant/InputVariable` and then call the new function as well to make it reconfigurable. I did not want to bloat the existing add functions with a new boolean parameter.

HopsanGUI/GUIConnector.cpp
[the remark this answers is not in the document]
robbr48: Done. Connectors now become broken if port is removed when changing "path" parameter. It seems to work. I have tried with having the other component properties dialog open, and it did not crash. When port position dialog is open parameters cannot be changed anyway.

HopsanGUI/GUIObjects/GUIComponent.cpp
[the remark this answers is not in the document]
robbr48: Yes, that code was quite different as well. It looks on the position of the internal model objects in the container and places the ports based on this. But perhaps some parts of it can still be abstracted out and re-used.

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
robbr48: We could clean up some variable names perhaps. The main problem is that the Windows API uses the WCHAR datatype. We must then convert HString -> const char* -> std::string -> std::wstring -> WCHAR, and also manually ensure the WCHAR is null terminated. The rest is basically just a function call.

HopsanCore/include/Parameters.h
```diff
@@ -82,6 +85,7 @@ class HOPSANCORE_DLLAPI ParameterEvaluator
     size_t mDepthCounter;
     ParameterEvaluatorHandler* mpParameterEvaluatorHandler;
     std::vector<HString> mConditions;
+    bool mTriggersReconfiguration = false;
```
peterNordin, line 88: Umm I think this should have failed on non c++03 export, but it did not ?
I dont think initializing variables directly in header should work.

HopsanCore/include/ComponentUtilities/TempDirectoryHandle.h
```diff
@@ -0,0 +1,63 @@
...
+    public:
+        TempDirectoryHandle(const HString &rPrefix);
+        ~TempDirectoryHandle();
+        const HString &path() const;
+        bool isValid() const;
+        bool addSubDirectory(const HString &rName);
+
+    private:
+        HString mPath;
+        bool mIsValid = false;
```
peterNordin, line 53: Init in header should/may not work with c++03

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
peterNordin: this still looks odd, the c++17 case assigns mPath, and does not return (from a function returning HString)
In thr other case it returns a Hstring, Also return HString should not be const, since it returns by value

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
peterNordin: no need

HopsanGUI/GUIObjects/GUIComponent.cpp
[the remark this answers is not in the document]
peterNordin: so what to do in this case, it is still hard coded for `this->getTypeName() == "FMIWrapper" && name == "path"`

componentLibraries/extensionLibrary/extensionLibrary.xml
peterNordin: Should the file be removed from this list?
The file it self was not added

HopsanGUI/GUIObjects/GUIComponent.cpp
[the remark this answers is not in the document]
robbr48: I think we can keep the component-specific code for now. I guess the GUI could check whether or not a reconfiguration was triggered and update ports if it was, but do we always want ports to be placed in the same way for all reconfigurable components? For some components reconfiguration should probably turn on or off a port, with a fixed position specified in the CAF file. It should even be possible to change the component icon depending on the reconfiguration, e.g. changing a pump from fixed to variable.

HopsanCore/include/ComponentUtilities/TempDirectoryHandle.h
robbr48: Ok, I moved it to constructor. Perhaps compilers are permissive for some reason.

HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
robbr48: I changed it to a return value. I guess I didn't notice, since you have to manually turn on C++17 compilation.

componentLibraries/extensionLibrary/extensionLibrary.xml
robbr48: Right. I missed it since the library XML file is not in the Qt project, so search function did not find it. Maybe we should add all library/component XML files as auxiliary files?

peterNordin (approved):

PR #2055, opened 2021-01-30
Issue 2053 autocomplete drives mrejster nuts

### LINKED ISSUE #2053: HCOM Autocomplete drives me nuts!

Autocomplete is a great function in HCOM. But it should be activated using a key (=TAB) not add whatever it finds suiting when pressing enter. In the case below, I wish to get the 1st generation. But if pressing enter at this stage I get the "102mm_90513810.A1" 'th generation, which doesn't even make sense. Unless I press space before enter, but this is not natural for my muscle memory from most other software, and I constantly find myself having to erase long autocompleted lines.
 
image

*A clear and concise description of what you want to happen.*
Activation of autocomplete with TAB key (as is the case already), but NOT by default when pressing enter.

*A clear and concise description of any alternative solutions or features you've considered. (optional)*

**Additional context**

peterNordin:
Includes the following changes to auto-complete
But more work is required to make it better

- Do not trigger auto complete popup after @ 
- Always show the auto complete popup, but only activate (create selection), if the user press Tab, (or use arrow keys or mouse to select in the list). If you press enter without first pressing Tab (or selecting in some other way). It will be as if the auto complete list was not even there.
- Fix autocomplete not recognizing . in names such as Step.out.y. This greatly increase the usefulness for components with many ports.
- Autocomplete now replaces the entire completer prefix, to ensure that case is corrected. Example: step.out.y -> Step.out.y

Resolve #2053

### DIFF
--- a/HopsanGUI/Widgets/HcomWidget.cpp
+++ b/HopsanGUI/Widgets/HcomWidget.cpp
@@ -568,17 +568,30 @@ void TerminalConsole::keyPressEvent(QKeyEvent *event)
             QTextEdit::keyPressEvent(event);
 
             // Compute prefix letters before cursor
+            QString fullPrefix, prevCharacter;
             QTextCursor tc = textCursor();
-            auto curretPos = tc.position();
-            tc.select(QTextCursor::WordUnderCursor);
-            QString prefix = tc.selectedText();
-            // Restore cursor position and anchor
-            tc.setPosition(curretPos);
+            while(true) {
+                auto curretPos = tc.position();
+                tc.select(QTextCursor::WordUnderCursor);
+                QString prefix = tc.selectedText();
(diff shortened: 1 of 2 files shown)


robbr48 (changes requested):
Looks good overall, but there is one problem. If I type `chpv mass.P1.`,  the autocomplete list will not pop up. If I skip the final dot however (`chpv mass.P1`), it will give me a list of available variables. I cannot see the problem in the code right now, but I think it should be fixed  before merging.

peterNordin:
Yes, the cursor manipulation does not work as I expected. I made a fix that seem to work. Only tested on Linux not sure how it works on Windows. Must test that as well. Hopefully text cursor works the same way there.

peterNordin:
Works the same on Windows

PR #2058, opened 2021-02-11
Clarify plot curve invert UI elements and tooltips

### LINKED ISSUE #2057: Invert curve in Plotwindows
When inverting a variable in a plotwindow it does become un-inverted when a simulation is performed. Then one need to invert it again. It is likely that the user wants to keep it inverted i.e. the same visual benefit with inverting the variable is likely still there after another simulation. 

**To Reproduce**
1. Simulate a system
2. Plot a varilble or more
3. Invert at leat one of the plotted curves
4. Simulate the model again
5. The invert will be un-ticked and the variable is thereby also un-inverted

**Expected behavior**
The invert setting in plotwindows should be preserved when simulations are ran. 

- OS: Windows 10 ver 1809 OS Build 17763.973
- Version: Info: HopsanCore 64-bit, Version: 2.16.0.20201108.1336 (Compiled in debug mode)
- Build: compile it myrself

peterNordin:
- For updates to latest generation, remember previous plot curve invert setting
- If there is a gap in the generations use toggle state from model (forget local state)
- If model toggle state changes, use that to override local choice (forget local state)
- If you toggle invert in one plot all curves for that generation will also be toggled in other plots (like Hopsan was already doing). This is to prevent confusion when plotting the same data in different plots.

Resolve #2057

### DIFF
--- a/HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
+++ b/HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
@@ -885,7 +885,7 @@ VariableTableWidget::VariableTableWidget(ModelObject *pModelObject, QWidget *pPa
     columnHeaders.append("Unit");
     columnHeaders.append("Value");
     columnHeaders.append("Quantity");
-    columnHeaders.append("PlotSettings");
+    columnHeaders.append("PlotLabel");
     columnHeaders.append("Port");
     this->setHorizontalHeaderLabels(columnHeaders);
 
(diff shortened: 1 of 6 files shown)


bjoer37 (approved):

robbr48 (approved):
Seems to work. 

It is kind of weird that the setting will affect all future generations, but not existing older generations of the same variable. But I guess it kind of makes sense to use the setting for the current generation as a "default setting" for every new generation. 

I don't like the fact that the invert function also inverts curves in other plot windows though. You may want to invert it in one window so you can compare it with some other variable, while you want to keep it uninverted in another window. But this problem already exists and is out of scope for this PR.

peterNordin:
The reason for also inverting all other instances of the curve/data in other plots was a feature request from a user.
It is confusing and easy to make mistakes when you have multiple plots including the same data, but they differ.
So flipping all curves/data in the same generation in all plot windows dimultaneously is intentional.
It will even flipp if the data is used as x-axis data.

If you want to only scale the local curve, I think that you can add a local curve-scale -1 inside the curve scale/offset settings. (but this will also add -1 the the unit display).

I dont want to touch old data "automatically" once a data generation has been "left behind". So any changes to old generations must be manually done.

I tried multiple approaches to this, and the currently implemented one is the one that is the least strange in my opinion.
Typically if you want a curve to be allways flipped, then you should set that in the component properties.
The curve control flip is just a convenience function.

PR #2062, opened 2021-02-17
Some fixes to FMI wrapper component

robbr48:
- Support all parameter types except enum
- Load start values for parameters
- Correct calling sequence (`fmi2_reset` in `finalize` and `fmi2_terminate` in `deconfigure`)
- Fix issue with temporary directories sometimes not being created on Linux

### DIFF
--- a/HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
+++ b/HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
@@ -140,7 +140,8 @@ bool TempDirectoryHandle::createDir(const HString &rName) const
 #ifdef _WIN32
         return _mkdir(rName.c_str()) == 0;
 #else
-        return mkdir(rName.c_str(), 0777) == 0;
+        int ret = mkdir(rName.c_str(), 0777) == 0;
+        return (ret != 0);
 #endif
 }
 
@@ -160,7 +161,9 @@ bool TempDirectoryHandle::createPath(const HString &rPath) const
                 return false;
             }
         }
-        currentLevel += "/"; // don't forget to append a slash
+        if(currentLevel.empty() || currentLevel[currentLevel.size()-1] != '/') {
+            currentLevel += "/"; // don't forget to append a slash
+        }
     }
 #endif
     return directoryExists(rPath);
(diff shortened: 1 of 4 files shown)


HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
```diff
@@ -170,8 +173,13 @@ bool TempDirectoryHandle::createPath(const HString &rPath) const
 bool TempDirectoryHandle::directoryExists(const HString &rName) const
 {
     struct stat st;
-    stat(rName.c_str(), &st);
-    return st.st_mode & S_IFDIR;
+    if(stat(rName.c_str(), &st) != 0) {
+        return false;
+    }
+    else if(st.st_mode & S_IFDIR) {
```
peterNordin, line 179: You could have keept: (no big deal)
`return st.st_mode & S_IFDIR;`

peterNordin (approved):
Did not test, but changes look good

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
```diff
@@ -414,6 +422,25 @@ class FMIWrapper : public ComponentSignal
 
         delete mpTempDir;
     }
+
+
+    //! @brief Removes all illegal characters from the string, so that it can be used as a variable name.
+    //! @param [in out] rName String that will be modified
```
peterNordin, line 428: I guess you copied this, in which case the original is also wrong. 
The rName is actually only input and should be called name
Also the todo note is not relevant in this case

peterNordin:
We could add this to 2.17.1, since we need to make that release once the optimization is fixed
Since it significantly improve that component

componentLibraries/extensionLibrary/Connectivity/FMIWrapper.hpp
robbr48: Should we not use the "r" prefix also for const references? It is still a reference. The `[in out]` is wrong though, and the `@returns` is missing. I will fix that.

robbr48:
Yes, including this in 2.17.1 would be really good. I am actually working on FMI cross checks, but I cannot publish them unless the FMI changes have been officially released.

PR #2063, opened 2021-02-17
Filepath parameter type 

robbr48:
Added a HFileType class, derived from HString without changes (just like HTextBlock).

Component parameters of this type automatically get a browse button in the parameter dialog, which opens a file browser dialog. It behaves just like a string parameter in all other aspects.

### DIFF
--- a/HopsanCore/src/Parameters.cpp
+++ b/HopsanCore/src/Parameters.cpp
@@ -749,7 +749,14 @@ bool ParameterEvaluatorHandler::evaluateInComponent(const HString &rName, HStrin
     bool success = false;
     for(size_t i = 0; i < mParameters.size(); ++i)
     {
-        if ( (mParameters[i]->getName() == rName) && (mParameters[i]->getType() == rType) )
+        if ( (mParameters[i]->getName() == rName) &&
+             ((mParameters[i]->getType() == rType) ||
+              ((mParameters[i]->getType() == "string") && (rType == "filepath")) ||
+              ((mParameters[i]->getType() == "string") && (rType == "textblock")) ||
+              ((mParameters[i]->getType() == "filepath") && (rType == "string")) ||
+              ((mParameters[i]->getType() == "filepath") && (rType == "textblock")) ||
+              ((mParameters[i]->getType() == "textblock") && (rType == "string")) ||
+              ((mParameters[i]->getType() == "textblock") && (rType == "filepath"))))
         {
             success = mParameters[i]->evaluate(rEvaluatedParameterValue);
         }
--- a/HopsanCore/src/Parameters.cpp
+++ b/HopsanCore/src/Parameters.cpp
@@ -750,13 +750,7 @@ bool ParameterEvaluatorHandler::evaluateInComponent(const HString &rName, HStrin
     for(size_t i = 0; i < mParameters.size(); ++i)
     {
         if ( (mParameters[i]->getName() == rName) &&
-             ((mParameters[i]->getType() == rType) ||
-              ((mParameters[i]->getType() == "string") && (rType == "filepath")) ||
-              ((mParameters[i]->getType() == "string") && (rType == "textblock")) ||
-              ((mParameters[i]->getType() == "filepath") && (rType == "string")) ||
-              ((mParameters[i]->getType() == "filepath") && (rType == "textblock")) ||
-              ((mParameters[i]->getType() == "textblock") && (rType == "string")) ||
-              ((mParameters[i]->getType() == "textblock") && (rType == "filepath"))))
+             (mParameters[i]->getType() == rType) )
         {
             success = mParameters[i]->evaluate(rEvaluatedParameterValue);
         }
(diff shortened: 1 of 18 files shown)


HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
```diff
@@ -1793,6 +1802,14 @@ void ParameterValueSelectionWidget::openValueEditDialog()
     }
 }
 
+void ParameterValueSelectionWidget::openFileBrowserDialog()
+{
+    QString filePath = QFileDialog::getOpenFileName(gpMainWindowWidget, "Select file", gpDesktopHandler->getDocumentsPath());
+    if(!filePath.isEmpty()) {
+        mpValueEdit->setText(filePath);
```
peterNordin, line 1809: Maybe we should open an additional dialog to let the user choose if the path should be absolute or relative to the model file. Or automatically make it relative to the model file automatically. I am not sure. If its absolute then if you "send" the model to someone else, paths may be wrong. But I think that relative paths should work. Provided that the same relative path is available on the other side.

Maybe make it relative automatically if the "data file" is on the same level deeper down the directory structure from where the model is located.

peterNordin:
I was going to add that you need to add test cases for this new parameter type, just like the "textbox" one. But apparently the existing tests are also failing. I have not been able to find why yet.

peterNordin:
I think its due to a failure in initialize in one of the changed components

peterNordin:
Ok found it, inside the "unittest_model", inside the subsubsystem 1DLookuptable it is trying to resolve "self.filename" as the comment character. It should resolve to the letter K (which is the parameter value entered as filename) but it is not able to do so and the actual value is "self.filename". So you are missing some code somewhere that is not handling the new parameter type as it should.

robbr48:
I made a fix that is kind of ridiculous, but we need to compare all six possible combinations of string, textblock and filepath when evaluating the component's own parameters. The problem actually already existed for textblocks, but this case was not included in the tests.

robbr48:
Can we also include this in 2.17.1? It makes the FMIWrapper component much more user-friendly.

I think I solved the issue with the FMI export test failing. Filepath parameters were not included when looking for model assets, so the csv file was not added to the exported FMU.

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
```diff
@@ -1804,9 +1804,14 @@ void ParameterValueSelectionWidget::openValueEditDialog()
...
 {
-    QString filePath = QFileDialog::getOpenFileName(gpMainWindowWidget, "Select file", gpDesktopHandler->getDocumentsPath());
+    QString defaultPath = gpDesktopHandler->getDocumentsPath();
+    if(!mpValueEdit->text().isEmpty()) {
+        defaultPath = QFileInfo(mpValueEdit->text()).absolutePath();
+    }
+    QString filePath = QFileDialog::getOpenFileName(gpMainWindowWidget, "Select file", defaultPath);
     if(!filePath.isEmpty()) {
-        mpValueEdit->setText(filePath);
+        QDir relDir(gpDesktopHandler->getExecPath());
```
peterNordin, line 1813: I think that it would be better to be relative to the model, if it is possible to get the model path here

peterNordin:
Not in 2.17.1, but I can just call the new release 2.18.0 instead, so yes we can include it.

HopsanCore/src/Parameters.cpp
```diff
@@ -749,7 +749,14 @@ bool ParameterEvaluatorHandler::evaluateInComponent(const HString &rName, HStrin
     bool success = false;
     for(size_t i = 0; i < mParameters.size(); ++i)
     {
-        if ( (mParameters[i]->getName() == rName) && (mParameters[i]->getType() == rType) )
+        if ( (mParameters[i]->getName() == rName) &&
+             ((mParameters[i]->getType() == rType) ||
+              ((mParameters[i]->getType() == "string") && (rType == "filepath")) ||
```
peterNordin, line 754: Hmm I was thinking about this madness, maybe we should not allow referring to parameters of other types.
Sure all of these are strings, but still.

In any case I dont think its the right place to solve this problem inside this function.

But I think the best solution is not to allow mixing types. So in that case the test model must be changed. The test that failed. Testing if string parameters can refer to other strings parameters in self, must be solved in some other way. (or temporarily disabled, since we want to merge this soon)

It would be nice with a better error message if the user try to refer a parameter to another armater that has a different type. Right now it fails in unclear way. But maybe that can wait as well.

robbr48: I removed this code and updated the tests. There is a small risk for backward compatibility problem if someone is already referencing filepath parameters of string type, but as you say it is very unlikely.

HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
robbr48: Ok, I changed it.

For some reason the model file info is stored in the top level system and not the model widget. I think we should move that, but not now.

PR #2065, opened 2021-02-24
Make random generators static to avoid getting same random value on every call

### LINKED ISSUE #2064: Optimization fails to start on Windows
**Describe the bug**
When clicking start optimization, nothing happens and the program freezes.

**Hopsan version**
It happens on Windows with 2.17.0, but seems to work fine on Linux. When I downgrade to 2.16 it works again on Windows.

I will do some more testing. Unfortunately my debugger does not work on Windows at the moment, so it is difficult to trace the error. It could be related to changes to the progress bar. I tried with progress bar both enabled and disabled though, and there were no difference. I also tried both multithreaded and singlethreaded simulation with no effect.

peterNordin:
- Make random generators static so that they are not reset on ever call
- Switch to mt19937 because its recommended everywhere (as @robbr48 initial intended)

I had problems with the same sequence of "random" values being generated every time the program was run, according to some stackoverflow post switching to mt19927 would solve that problem, (but I do not think that it did).

Resolves #2064

### DIFF
--- a/HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
+++ b/HopsanCore/src/ComponentUtilities/TempDirectoryHandle.cpp
@@ -40,6 +40,7 @@
 #endif
 #if __cplusplus >= 201103L
 #include <random>
+#include <chrono>
 #endif
 #include <sstream>
 #include <iomanip>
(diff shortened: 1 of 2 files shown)


robbr48:
I think this GCC bug is related:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85494

Someone suggest seeding the generator with current time:
`static_cast<long unsigned int>(std::chrono::high_resolution_clock::now().time_since_epoch().count())`

I think it's quite important that running multiple optimizations with the same settings will generate different results. It is a common approach to run multiple optimizations and picking the best result.

robbr48 (approved):
Seems to work! Also both Nelder-Mead and Complex-RF produce different results on every optimization.

peterNordin:
I added seeding as you suggested as a special case for MinGW

PR #2069, opened 2021-03-19
Use interpolation-based downsampling to reduce vector size

### LINKED ISSUE #2068: Frequency analysis does not use the chosen time frame
**Describe the bug**
*If you make a simulation from -10 s to 10 s and choose to make a bode plot from 0 s to 10 s, the resulting bode plot is not the same as if one simulates from 0 to 10, although the exact same data is shown in the plot window.*

**To Reproduce**
*Test in the provided model:*
1. Simulate from 0 to 10 s
2. Make bode plot in the scope  (use standard settings)
3. The resonance frequency in the bode plot is approximately 18.5 rad/s (correct)
4. Redo simulation from -10 to 10 instead
5. The resonance in the bode plot is now around 8.2 rad/s (incorrect)

**Expected behavior**
*The frequency response calculation should be based on the data in the plot window, which means that both settings above should result in the same bode plot. I should note that this could be caused by my incompetence when it comes to frequency analysis...*

**Screenshots, Model or Data files**
*If applicable, add screenshots, models or data files, that display or cause the bug.*

*Version information is shown in the message widget when HopsanGUI is started*
 - OS: Windows 10
 - Version: [e.g. HopsanCore 64-bit, Version: 2.18.0.20210225.2108
                       HopsanGUI 64-bit, Version: 2.18.0.20210225.2108]
 - Build: Is this an **official** build downloaded from the web page
[freqBug.zip](https://github.com/Hopsan/hopsan/files/6143129/freqBug.zip)

**Additional information**
*Add any other context/information about the problem here.*

robbr48:
There was an error in `reduceVectorSize` function due to integer division. This resovles the problem by replacing the previous closest-index method with a more accurate downsampling algorithm based on linear interpolation. It will affect all frequency analysis plots (spectrum, bode and nyquist).

Resolves #2068.

### DIFF
--- a/HopsanGUI/Utilities/GUIUtilities.cpp
+++ b/HopsanGUI/Utilities/GUIUtilities.cpp
@@ -508,18 +508,24 @@ void FFT(QVector< complex<double> > &data)
 }
 
 
-//! @brief Reduces number of log samples of a data vector to specified value
-//! @param vector Reference to vector that will be reduced
+//! @brief Resample a data vector to specified size using linear interpolation
+//! @param [in,out] vector Reference to vector that will be resampled
 //! @param newSize New size of vector
-void reduceVectorSize(QVector<double> &vector, int newSize)
+void resampleVector(QVector<double> &vector, int newSize)
 {
     int oldSize = vector.size();
 
-    QVector<double> tempVector;
+    QVector<double> tempVector(newSize);
 
-    for(int i=0; i<newSize; ++i)
-    {
-        tempVector.append(vector.at(oldSize/newSize*i));
+    for(int i=0; i<newSize; ++i) {
+        double pos = double(oldSize-1)/double(newSize-1)*double(i);
+        int prev = int(floor(pos)); //Previous index in original vector
+        int next = min(oldSize-1, int(ceil(pos)));  //Next index in original vector
+        double x = 0;
+        if(next != prev) {
+            x = (pos-prev)/(next-prev);  //Fraction
+        }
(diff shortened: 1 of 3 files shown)


HopsanGUI/Utilities/GUIUtilities.cpp
```diff
@@ -519,7 +519,14 @@ void reduceVectorSize(QVector<double> &vector, int newSize)
...
     {
-        tempVector.append(vector.at(oldSize/newSize*i));
+        double pos = double(oldSize)/double(newSize)*double(i);
+        int prev = int(floor(pos)); //Previous index in original vector
+        int next = int(ceil(pos));  //Next index in original vector
+        double x = 0;
+        if(next != prev) {
+            x = (pos-prev)/(next-prev);  //Fraction
+        }
+        tempVector.append((1.0-x)*vector.at(prev) + x*vector.at(next));
```
peterNordin, line 529: Since newSize is known, the vector memory should be reserved before the loop

I know that this function is meant to reduce vector, but what if newSize is >=oldSize, then I think that vector.at() will throw
Maybe protect against that with in initial if check, and just do no do anything if newSize is not smaller

robbr48: Ok, I added preallocation and made some modifications to prevent going out of bounds. It should now work even if newSize is larger.

robbr48: I also renamed the function to `resampleVector`, to make clear that it works for both upsampling and downsampling.

peterNordin (approved):

PR #2070, opened 2021-03-22
Improvements and cleanup of generated Simulink code

robbr48:
- Use mdlStart function for initialization
- Fix indentation
- Avoid confusing temporary variables
- Remove unused code parts
- Make sure system parameters works correctly
- Add more comments
- Use sample time properly

### DIFF
--- a/HopsanGenerator/templates/simulinkWrapperTemplate.cpp
+++ b/HopsanGenerator/templates/simulinkWrapperTemplate.cpp
@@ -24,6 +24,7 @@
 
 #define S_FUNCTION_NAME <<<name>>>
 #define S_FUNCTION_LEVEL 2
+#define MDL_START
 
 #include <sstream>
 #include <string>
@@ -38,7 +39,7 @@ HopsanEssentials gHopsanCore;
 ComponentSystem* pComponentSystem;
 bool isOkToSimulate = false;
 
-#define NUMPARAMS 0//<<<numparams>>>
+#define NUMPARAMS 0
 
 <<<15>>>
 static void mdlInitializeSizes(SimStruct *S)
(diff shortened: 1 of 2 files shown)


robbr48:
A side effect of these improvements is that Hopsan S-functions now also support the "fast restart" option in Simulink. This can save a lot of simulation time when running batch simulations.

HopsanGenerator/templates/simulinkWrapperTemplate.cpp
```diff
@@ -94,25 +84,25 @@ static void mdlInitializeSizes(SimStruct *S)
...
     pComponentSystem->setDesiredTimestep(<<<timestep>>>);
+
 <<<5>>>}
 
 
 static void mdlInitializeSampleTimes(SimStruct *S)
 {
-    //ssSetSampleTime(S, 0, 0.001);
-    //ssSetOffsetTime(S, 0, 0.0);
+    ssSetSampleTime(S, 0, 0.001);
```
peterNordin, line 94: I am unsure how this works, is 0.001 the default step time ? 
I assume that the user can change this from simulation somehow ?

peterNordin (approved):

HopsanGenerator/templates/simulinkWrapperTemplate.cpp
robbr48: It is actually hard-coded, depending on the step size in the exported Hopsan model. Making this a tunable parameter is not possible, because sample time is defined only once before reading any other variables in a Simulink simulation. We used to have this commented out, but I noticed it could cause problems if running Simulink with variable step-size. As I interpret it, this will force Simulink to call this block at exactly the specified interval. Not perfect, but it should not be worse than before either.

PR #2074, opened 2021-03-24
Support variable aliases in hopsanc

robbr48:
Make it possible to get data vectors by alias in Python, Julia and Matlab.

Usage:
```
getDataVector(alias)
getDataVector(system1|alias)
getDataVector(system1|system2|alias)
```

### DIFF
--- a/hopsanc/src/hopsanc.cpp
+++ b/hopsanc/src/hopsanc.cpp
@@ -100,10 +100,6 @@ int getDataVector(const char* variable, double *data)
     hopsan::HVector<hopsan::HString> splitSys = varStr.split('|');
     hopsan::HVector<hopsan::HString> splitVar = splitSys.last().split('.');
     splitSys.resize(splitSys.size()-1);
-    if(splitVar.size() < 3) {
-        printMessage("Error: Component name, port name and variable name must be specified.");
-        return -1;
-    }
 
     //Find system
     hopsan::ComponentSystem *pSystem = spCoreComponentSystem;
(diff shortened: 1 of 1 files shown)


hopsanc/src/hopsanc.cpp
```diff
@@ -115,6 +111,24 @@ int getDataVector(const char* variable, double *data)
...
 
+    //Check for alias if splitVar is of size one
+    if(splitVar.size() == 1 && pSystem->getAliasHandler().hasAlias(splitVar[0])) {
+        hopsan::HString compName, portName;
+        int varId;
+        pSystem->getAliasHandler().getVariableFromAlias(splitVar[0], compName, portName, varId);
+        hopsan::Component *pComp = pSystem->getSubComponent(compName);
+        hopsan::Port *pPort = pComp->getPort(portName);
+        std::vector< std::vector<double> > *pLogData = pPort->getLogDataVectorPtr();
+        for (size_t t=0; t<spCoreComponentSystem->getNumActuallyLoggedSamples(); ++t) {
```
peterNordin, line 122: Maybe this should be pSystem, in case the number of logged samples would be different from the main system.
Not sure if that can happen.

robbr48: The hopsanc interface can only set number of log samples for the entire model. Is it even possible to have different number of samples in a sub system? Didn't we take away that feature, or was that only from the gui?

peterNordin: But still, since we take the logdata vector from a pPort in a component in pSystem, it makes sens to ask pSystem for the actual number of log samples. Even though they should be the same as the main system.

robbr48: That's true. I changed it now.

peterNordin (approved):

PR #2079, opened 2021-04-07
Add moment of inertia quantity

robbr48:
This is a common quantity, that seemed to have the same unit but different formatting on many places. Not sure if we also need to add alternative units.


peterNordin (approved):
Can add other units later if needed

PR #2081, opened 2021-04-20
SymHop expressions in HCOM terminal

robbr48:
Enables defining and evaluating SymHop expressions in HCOM. Most functions in the SymHop API that makes sense in this context is supported.

Example:
```
>> a=expr(sin(x)+y)
sin(x)+y
>> b=der(a,expr(x))
cos(x)
>> x=5
Assigning scalar x with 5
>> y=2
Assigning scalar y with 2
>> eval(a)
1.0410757253368614
>> eval(b)
0.28366218546322625
```


HopsanGUI/HcomHandler.cpp
```diff
@@ -381,6 +381,21 @@ HcomHandler::HcomHandler(TerminalConsole *pConsole) : QObject(pConsole)
     registerInternalFunction("x", "Returns the X-vector of the specified variable.","Usage: x(vector)");
     registerInternalFunction("td", "Converts variable to time domain." "Usage: y = td(x)");
     registerInternalFunction("fd", "Converts variable to frequency domain." "Usage: y = fd(x)");
+    registerInternalFunction("expr", "Defines a new SymHop expression" "Usage: e = expr(...)");
```
peterNordin, line 384: Maybe call it something else then "SymHop" expression. It is not clear to the end-user what SymHop is. Maybe call it mathematical expression or something similar. Also in all other places, and related documentation. 
I did this mistake whit NumHop, and I intend to replace all mention of NumHop with "Embedded script" since it makes more sense to the end users. We should do the same with SymHop. NumHop and SymHop are only the name of the libraries implementing the functionallity.

HopsanGUI/HcomHandler.cpp
```diff
@@ -7134,6 +7158,211 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
             return;
         }
     }
+    else if(desiredType != Scalar && desiredType != DataVector && isHcomFunctionCall("expr", expr)) {
+       QStringList args = extractFunctionCallExpressionArguments(expr);
+       if(args.size() != 1) {
+           HCOMERR("Wrong number of arguments for expr function.\n"+mLocalFunctionDescriptions.find("ddt").value().second);
```
peterNordin, line 7164: why add `+mLocalFunctionDescriptions.find("ddt").value().second`to the error message, in particular `ddt`

HopsanGUI/HcomHandler.cpp
```diff
@@ -7134,6 +7158,211 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
+           mAnsType = Undefined;
+           return;
+       }
+       mAnsType = Expression;
+       return;
+    }
+    else if(isHcomFunctionCall("eval", expr)) {
+        QStringList args = extractFunctionCallExpressionArguments(expr);
+        if(args.size() != 1) {
+            HCOMERR("Wrong number of arguments for eval function.\n"+mLocalFunctionDescriptions.find("ddt").value().second);
```
peterNordin, line 7183: again `ddt` ?

HopsanGUI/HcomHandler.cpp
```diff
@@ -7134,6 +7158,211 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
...
+        evaluateExpression(args[1]);
+        if(mAnsType != Expression) {
+            HCOMERR("Failed to evaluate expression \""+args[1]+"\".");
+            mAnsType = Undefined;
+            return;
+        }
+        expr2 = mAnsExpression;
+        if(getFunctionName(expr) == "der") {
+            bool ok;
+            mAnsExpression = mLocalExpressions[args[0]].derivative(mAnsExpression,ok);
```
peterNordin, line 7231: Should we not use the expr1 and expr2  in some way this section of the code ?

HopsanGUI/HcomHandler.cpp
robbr48: Good point. It is obvious to me what it means, but not the end user. I changed it to "symbolic expression".

HopsanGUI/HcomHandler.cpp
robbr48: Copy-paste error. It is intended to print the help string for the function, but of course for the <i>correct</i> function. Perhaps this can leter be wrapped into some more beautiful auxiliary function.

HopsanGUI/HcomHandler.cpp
robbr48: Yes, that's true. It is fixed now.

PR #2082, opened 2021-04-20
Monitor text files for external changes

### LINKED ISSUE #2059: Files loaded into a Hopsan text editor should monitor files for external changes
When a file loaded into a Hopsan text editor window changes from the outside, the file:
-  Is not reloaded automatically
- A notification is not shown that the file is out-of-date
- There is no way to refresh (typically pressing F5)

Either of these would work

- Auto refresh if no local unsaved changes
- Or some kind of notification that file is out-of-date and need to be refreshed. How to handle unsaved local changes I do not know

Also:
- Add refresh file functionality

**Additional context**
This happened when I tried to add an existing xml component to a library and had already opened the xml file for editing in Hopsan. The file was changed but I did not see it and had to close/open the file again

robbr48:
- Monitor text files for external changes
- Let user choose to reload file or ignore changes
- Support text files with existing revert action
- Add warning dialog before reverting a file

Resolves #2059.

Possible future work would be to also monitor hmf files for changes, but that is both less important and more complicated.

### DIFF
--- a/HopsanGUI/Widgets/TextEditorWidget.cpp
+++ b/HopsanGUI/Widgets/TextEditorWidget.cpp
@@ -136,9 +136,11 @@ void TextEditorWidget::fileChanged(QString filePath)
     qDebug() << "File changed: " << filePath;
     //We need to add the path again, because some text editors removes the actual file and
     //creates a new one when saving (e.g. gedit). This makes the watcher stop watching.
-    qobject_cast<QFileSystemWatcher*>(sender())->addPath(filePath);
-
-    mpFileChangeNotificationWidget->setVisible(true);
+    QFileSystemWatcher *pWatcher = qobject_cast<QFileSystemWatcher*>(sender());
+    if(pWatcher != nullptr) {
+        pWatcher->addPath(filePath);
+        mpFileChangeNotificationWidget->setVisible(true);
+    }
 }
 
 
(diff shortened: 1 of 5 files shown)


HopsanGUI/Widgets/TextEditorWidget.cpp
```diff
@@ -112,6 +131,16 @@ void TextEditorWidget::find(QString text, QTextDocument::FindFlags flags)
     mpEditor->find(text,flags);
 }
 
+void TextEditorWidget::fileChanged(QString filePath)
+{
+    qDebug() << "File changed: " << filePath;
+    //We need to add the path again, because some text editors removes the actual file and
+    //creates a new one when saving (e.g. gedit). This makes the watcher stop watching.
+    qobject_cast<QFileSystemWatcher*>(sender())->addPath(filePath);
```
peterNordin, line 139: While tight now, sender is always a QFileSystemWatcher, there should be a pointer check on the result of qobject_cast, just in case someone else sends the signal, or the code will crash due to nullptr.

peterNordin (changes requested):
This works find except for one thing, the notification shows up in multiple tabs when I change one file. I opened two files, (one view library file, the .cc/.cpp file in an external library) and one manually opened script file. Now when I make changes in one file, the "File changed widget comes up in both tabs even if I change only one of them).

So when a file change is detected, the TextEditorWidget should check that the filePath that comes in the signal matches its own file before displaying to notification.

HopsanGUI/Widgets/TextEditorWidget.cpp
robbr48: That's true. This is actually a problem on many places already. It works for now, but only until someone changes something sometimes in the future...

robbr48:
Ops, my bad. The intention was of course that it should compare the paths. I just forgot to add that check. It is fixed now.

PR #2083, opened 2021-04-20
"Always visible" setting for GUI components

### LINKED ISSUE #1958: The "toggle signal components visibility" function hides too much
*It always frustrates me when I have a signal component that has animation, and when I want to hide only the signal lines (for instance when I want to show an interactive animation without a lot of signal lines), my animation component is hidden as well*

*I would like to be able to override the "toggle signal components visibility", for instance by including an "Always show" option in the component.*

*An alternative component type (e.g. Animation) could be used to ensure that the component is not hidden when I choose to hide signal components*

robbr48:
This setting prevents signal components from being hidden when using the "toggle signal component visibility" feature.

Resolves #1958.

### DIFF
--- a/HopsanGUI/GUIObjects/GUIContainerObject.cpp
+++ b/HopsanGUI/GUIObjects/GUIContainerObject.cpp
@@ -115,7 +115,7 @@ void SystemObject::commonConstructorCode()
     mIsCreatingConnector = false;
     mShowSubComponentPorts = gpMainWindow->mpTogglePortsAction->isChecked();
     mShowSubComponentNames = gpMainWindow->mpToggleNamesAction->isChecked();
-    mSignalsHidden = gpMainWindow->mpToggleSignalsAction->isChecked();
+    mSignalsHidden = !gpMainWindow->mpToggleSignalsAction->isChecked();
     mLossesVisible = false;
     mUndoEnabled = true;
     mGfxType = UserGraphics;
--- a/HopsanGUI/GUIObjects/GUIContainerObject.cpp
+++ b/HopsanGUI/GUIObjects/GUIContainerObject.cpp
@@ -115,7 +115,7 @@ void SystemObject::commonConstructorCode()
     mIsCreatingConnector = false;
     mShowSubComponentPorts = gpMainWindow->mpTogglePortsAction->isChecked();
     mShowSubComponentNames = gpMainWindow->mpToggleNamesAction->isChecked();
-    mSignalsHidden = !gpMainWindow->mpToggleSignalsAction->isChecked();
+    mSignalsVisible = gpMainWindow->mpToggleSignalsAction->isChecked();
     mLossesVisible = false;
     mUndoEnabled = true;
     mGfxType = UserGraphics;
@@ -239,7 +239,7 @@ void SystemObject::makeMainWindowConnectionsAndRefresh()
     // Update the main window menu and toolbar actions that are system specific
     gpMainWindow->mpTogglePortsAction->setChecked(mShowSubComponentPorts);
     gpMainWindow->mpToggleNamesAction->setChecked(mShowSubComponentNames);
-    gpMainWindow->mpToggleSignalsAction->setChecked(!mSignalsHidden);
+    gpMainWindow->mpToggleSignalsAction->setChecked(mSignalsVisible);
     gpMainWindow->mpEnableUndoAction->setChecked(mUndoEnabled);
     gpMainWindow->mpUndoAction->setEnabled(mUndoEnabled);
     gpMainWindow->mpRedoAction->setEnabled(mUndoEnabled);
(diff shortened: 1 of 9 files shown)


HopsanGUI/GUIObjects/GUIModelObject.cpp
```diff
@@ -1398,6 +1407,11 @@ QAction *ModelObject::buildBaseContextMenu(QMenu &rMenu, QGraphicsSceneContextMe
         mpParentSystemObject->getUndoStackPtr()->newPost();
         this->flipHorizontal();
     }
+    else if (selectedAction == pAlwaysVisbleAction)
+    {
+        mpParentSystemObject->getUndoStackPtr()->newPost();
```
peterNordin, line 1412: Not sure if `mpParentSystemObject->getUndoStackPtr()->newPost();` is needed. Since no undo data seems to be added ?

HopsanGUI/GUIObjects/GUIModelObject.cpp
```diff
@@ -1824,6 +1838,13 @@ QGraphicsSvgItem *ModelObject::getIcon()
     return mpIcon;
 }
 
+void ModelObject::setAlwaysVisible(const bool visible)
+{
+    mAlwaysVisible = visible;
+
+    mpIcon->setVisible(!mpParentSystemObject->areSignalsHidden());
```
peterNordin, line 1845: Should this be plain and simply set to "true" in this case.
I did not realize, but you can actually interact with the components when they are hidden. (But that is not a problem). So when I check setAlwaysVisible in hidden state I would expect the icon to show up.

HopsanGUI/GUIObjects/GUIModelObject.cpp
```diff
@@ -1055,6 +1056,9 @@ QDomElement ModelObject::saveGuiDataToDomElement(QDomElement &rDomElement)
     QPointF cpos = this->getCenterPos();
     appendPoseTag(xmlGuiStuff, cpos.x(), cpos.y(), rotation(), this->mIsFlipped, 10);
 
+    // Save the alwasys visible setting
+    xmlGuiStuff.setAttribute("visible", mAlwaysVisible);
```
peterNordin, line 1060: Maybe call the attribute "always_visible" to avoid confusion. Since components with this property set to "false" will still be visible in the normal case.

HopsanGUI/GUIObjects/GUIContainerObject.cpp
```diff
@@ -115,7 +115,7 @@ void SystemObject::commonConstructorCode()
     mIsCreatingConnector = false;
     mShowSubComponentPorts = gpMainWindow->mpTogglePortsAction->isChecked();
     mShowSubComponentNames = gpMainWindow->mpToggleNamesAction->isChecked();
-    mSignalsHidden = gpMainWindow->mpToggleSignalsAction->isChecked();
+    mSignalsHidden = !gpMainWindow->mpToggleSignalsAction->isChecked();
```
peterNordin, line 118: Maybe it would have been better to rename the variable to mSignalsVisible (instead of having bool with negative meaning), but I guess this works to. Odd that we have not seen this bug before. 
No need to change unless you want to, then all places where this is used must be changed to.

HopsanGUI/loadFunctions.cpp
```diff
@@ -280,6 +280,8 @@ ModelObject* loadModelObject(const QDomElement &domElement, SystemObject* pSyste
     parsePoseTag(guiData.firstChildElement(HMF_POSETAG), posX, posY, target_rotation, isFlipped);
     target_rotation = normDeg360(target_rotation); //Make sure target rotation between 0 and 359.999
 
+    bool alwaysVisible = parseAttributeBool(guiData, "visible", false);
```
peterNordin, line 283: attribute name should maybe be "always_visible"

peterNordin (changes requested):
Works OK, but consider my comments.

HopsanGUI/GUIObjects/GUIModelObject.cpp
robbr48: I solved it the proper way instead, and added an undo post.

HopsanGUI/GUIObjects/GUIModelObject.cpp
robbr48: Not simply to true, but it should be visible if signals are not hidden _or_ if the visibility setting is true.

HopsanGUI/GUIObjects/GUIModelObject.cpp
robbr48: I renamed it to "alwaysvisible", without the underscore. Most existing two-word attributes are written without underscore.

HopsanGUI/GUIObjects/GUIContainerObject.cpp
robbr48: Agreed. This has been confusing for a long time. If the action is toggled, the boolean should become true.

PR #2084, opened 2021-04-21
Static linking of MinGW DLLs for FMI export on Windows

robbr48:
Link stdc++ and pthread statically on Windows. This resolves issues with FMUs not being able to find `libstdc++-6.dll` and `libwinpthread-1.dll`.


peterNordin (approved):
Hmm, I wonder why this is not needed on Linux, maybe it will find the other libraries because they are available in the system. But then we may have problems if we send a FMU build on a newer version to an older version system. We should investigate that. Those libs might be required on Linux as well.

robbr48:
Yes, I think that might be necessary. But I leave this for another PR.

PR #2087, opened 2021-04-23
More Modelica fixes

robbr48:
- Improved performance
- Cleanup of generated code
- Added a `preventCavitation` function (performs recalculation when a negative pressure is detected)
- Some improvements to example components

### DIFF
--- a/componentLibraries/ModelicaExampleLib/Modelica22Valve.mo
+++ b/componentLibraries/ModelicaExampleLib/Modelica22Valve.mo
@@ -1,14 +1,14 @@
 model Modelica22Valve "2/2 Valve"
-    annotation(hopsanCqsType = "Q", linearTransform="bdf2")
+    annotation(hopsanCqsType = "Q", linearTransform="trapezoid")
 
     input Real x;
     NodeHydraulic P1;
     NodeHydraulic P2;
 
     output Real xv;
-    Real vv(start=0);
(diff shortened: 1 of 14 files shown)


componentLibraries/ModelicaExampleLib/Modelica22Valve.mo
```diff
@@ -1,32 +1,33 @@
...
+    Ks = K1*xv;
+    der(der(xv))/omega/omega + der(xv)*2*B/omega + xv = x;
     P2.q = turbulentFlow(P1.c,P2.c,P1.Zc,P2.Zc,Ks);
     P1.q+P2.q=0;
     P1.p = P1.c + P1.Zc*P1.q;
     P2.p = P2.c + P2.Zc*P2.q;
-    
-algorithm
+    preventCavitation(P1.p,P1.c,P1.Zc);
+    preventCavitation(P2.p,P2.c,P2.Zc);
```
peterNordin, line 31: Is there any documentation about what "hopsan/modelica" help functions are available and what their arguments are ?

robbr48: No, not at the moment. They don't even show up with autocomplete I think, so the only way to find them is by looking in the example library. The Modelica generator is still quite experimental, so I want to try things out before writing too much documentation.

PR #2088, opened 2021-04-25
Expect derivative input in PID2 to be a derivative and not a delta

### LINKED ISSUE #2080: Possible error in SignalPID2
Maybe I'm mistaken here, but I think this is wrong. If the derivative input is not connected, the derivative is approximated according to the following code:
```
if (mUseDeInput)
{
    dErr = (*mpDerr);
}
else
{
    dErr = err-mLastErr;
    mLastErr = err;
}
```
Should not dErr be divided by mTimestep in the second case?

peterNordin:
Resolves #2080


robbr48 (approved):
Looks good!

PR #2089, opened 2021-04-27
Convert SymHop expressions to LaTeX

robbr48:
Can be tested by using the `latex` function in HCOM:
```
>> f=expr((a+b)/(c+d))
(a+b)/(c+d)
>> latex(f)
\dfrac{a+b}{c+d}
```

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -396,6 +396,7 @@ HcomHandler::HcomHandler(TerminalConsole *pConsole) : QObject(pConsole)
     registerInternalFunction("euler", "Transforms derivatives in a symbolic expression using the forward Euler method", "Usage: e = euler(f)");
     registerInternalFunction("bdf1", "Transforms derivatives in a symbolic expression using the first order backward differentiation formula (implicit Euler)", "Usage: e = bdf1(f)");
     registerInternalFunction("bdf2", "Transforms derivatives in a symbolic expression using the second order backward differentiation formula", "Usage: e = bdf2(f)");
+    registerInternalFunction("latex", "Prints a symbolic expression with LaTeX syntax", "Usage: latex(e)");
 
     //Setup local function pointers (used to evaluate expressions in SymHop)
     registerFunctionoid("aver", new HcomFunctionoidAver(this), "Calculate average value of vector", "Usage: aver(vector)");
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -7261,7 +7261,7 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
             isHcomFunctionCall("removeDivisors", expr) || isHcomFunctionCall("left", expr) ||
             isHcomFunctionCall("right", expr) || isHcomFunctionCall("trapezoid", expr) ||
             isHcomFunctionCall("euler", expr) || isHcomFunctionCall("bdf1", expr) ||
-            isHcomFunctionCall("bdf2", expr)) {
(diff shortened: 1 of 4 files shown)


HopsanGUI/HcomHandler.cpp
```diff
@@ -7261,7 +7261,7 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
             isHcomFunctionCall("removeDivisors", expr) || isHcomFunctionCall("left", expr) ||
             isHcomFunctionCall("right", expr) || isHcomFunctionCall("trapezoid", expr) ||
             isHcomFunctionCall("euler", expr) || isHcomFunctionCall("bdf1", expr) ||
-            isHcomFunctionCall("bdf2", expr)) {
+            isHcomFunctionCall("bdf2", expr) || isHcomFunctionCall("latex", expr)) {
```
peterNordin, line 7264: Are any of these symhop/hcom functions part of the hcom documentation?

SymHop/src/SymHop.cpp
```diff
@@ -1279,6 +1279,203 @@ QString Expression::toString() const
     return ret;
 }
 
+QString Expression::toLaTeX() const
+{
+    QString ret;
+
+    if(this->isSymbol())
+    {
+        QStringList greekSymbols = {"alpha","beta","gamma","Gamma","delta","Delta","epsilon","varepsilon","zeta","eta","theta","vartheta","Theta","iota","kappa","lambda","Lambda","mu","nu","xi","Xi","pi","Pi","rho","varrho","sigma","Sigma","tau","upsilon","Upsilon","phi","varphi","Phi","chi","psi","Psi","omega","Omega"};
```
peterNordin, line 1288: Some of these have only lower case version, maybe intended, I dont know

SymHop/src/SymHop.cpp
```diff
@@ -1279,6 +1279,203 @@ QString Expression::toString() const
...
+            ret = QString::number(mString.toDouble());
+        }
+        else {
+            ret = mString;
+        }
+    }
+    else if(this->isFunction())
+    {
+        QStringList builtinFunctions = {"sin","cos","tan","arcsin","arccos","arctan","cot","sec","csc","sinh","cosh","tanh","lg","ln","log","min","max"     };
+        if(mFunction == "pi" && mArguments.size() == 0) {
```
peterNordin, line 1303: Is pi a function ?

SymHop/src/SymHop.cpp
```diff
@@ -1279,6 +1279,203 @@ QString Expression::toString() const
...
+        {
+            powerStr.prepend("(");
+            powerStr.append(")");
+        }
+        if(powerStr.size() > 1) {
+            powerStr.prepend("{");
+            powerStr.append("}");
+        }
+        ret = baseStr+"^"+powerStr;
+        //ret = "pow("+baseStr+","+powerStr+")";
```
peterNordin, line 1469: comment

peterNordin (approved):
I just browsed through the code, look good.
I guess this is helpful when writing the markdown documentation ?
Pasting the c++ equations into hcom and converting to latex ?

SymHop/src/SymHop.cpp
robbr48: Yes, it is intentional. Many upper case greek letters are equivalent to latin letters, and those don't have latex commands.

SymHop/src/SymHop.cpp
robbr48: Yes, in SymHop, but not in Latex. Therefore `pi()` must be converted to `\pi`.

HopsanGUI/HcomHandler.cpp
robbr48: No, you are right. They should be registered as internal functions. I will fix that.

PR #2090, opened 2021-05-03
Replace foreach loops with range-based C++ loops

robbr48:
We need to remove  the Qt foreach loops since they are deprecated and will be removed in future Qt versions. 
This is probably hard to review, but most changes are only syntactic. The only functional changes are actually the goto expressions in SymHop.

- Replace all `Q_FOREACH` loops with range-based C++ loops
- Replace all `foreach` loops with range-based C++ loops
- Make some regular loops range-based
- Fix some bracket styles
- Remove some comments
- Replace some goto expressions with loops
- Some minor fixes (e.g. replace `size_t` with `int` to avoid warnings)

### DIFF
--- a/HopsanGUI/PlotTab.cpp
+++ b/HopsanGUI/PlotTab.cpp
@@ -144,9 +144,7 @@ void PlotTab::openTimeOffsetDialog()
 //! @brief Toggles the axis lock on/off for the enabled axis
 void PlotTab::toggleAxisLock()
 {
-    PlotArea *pArea;
-    Q_FOREACH(pArea, mPlotAreas)
-    {
+    for(PlotArea *pArea : mPlotAreas) {
         pArea->toggleAxisLock();
     }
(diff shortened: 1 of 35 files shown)


HopsanGUI/PlotTab.cpp
```diff
@@ -1687,7 +1643,7 @@ void PlotTab::openCreateBodePlotDialog()
                 else if(pWindowingComboBox->currentIndex() == 1) {
                     function = HannWindow;
                 }
-                else if(pWindowingComboBox->currentIndex() == 2) {
+                else/* if(pWindowingComboBox->currentIndex() == 2)*/ {
```
peterNordin, line 1646: Everything looks OK, but why this change ?

peterNordin: Aha, there are only three choices

PR #2103, opened 2021-11-04
Issue 2098 losscalculation

### LINKED ISSUE #2098: Loss calculation shows wrong value for domain-specific losses
**Describe the bug**
When a component for example only has mechanical ports, the "mechanical losses" should be the same as "total losses". It is however much smaller. It seems like only the last log interval is considered for domain specific losses.

**To Reproduce**
1. Simulate a model
2. Click on "Calculate Losses"
3. All domain-specific losses will be wrong

**Expected behavior**
The domain-specific losses should sum up to the total losses in each component.

**Additional information**
Related to #540.

peterNordin:
Resolves #2098

### DIFF
--- a/HopsanGUI/GUIObjects/GUIModelObject.cpp
+++ b/HopsanGUI/GUIObjects/GUIModelObject.cpp
@@ -431,6 +431,7 @@ void ModelObject::showLosses()
     QTime time;
 
     mTotalLosses = 0.0;
+    mDomainSpecificLosses.clear();
 
     QString unit = " J";
     double div = 1;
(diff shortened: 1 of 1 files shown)


peterNordin:
I am not sure about multi ports. If there even is a problem. When only one port is connected to the multi port it seems to work.
If I connect multiple ports I do not know what to expect.


Also I am not sure, but is are the losses not calculated twice. In both components sharing a port connection.
Maybe this is the intended behavior ?

Example Position Servo, I was looking at the Cylinder connected to the valve. Hydraulic losses are calculated on both sides since they share the same ports. (Cylinder has multi port so actual data is in the connected port). But this is also the case with normal ports, they always share data. And I guess added energy is also calculated twice ?

robbr48:
It should be computed twice; the output power from one of the components will be input power to the other one. The losses is the difference between all input and output power in each component.

Multiports should be treated as multiple ports. The input or output power should be computed for each connection.

robbr48 (approved):
Looks good!

PR #2106, opened 2021-11-26
Support elseif and nested if-statements in HCOM

robbr48:
Enables writing if-statements in this format in HCOM scripts:
```
if(x<1)
  if(y>42)
    print "Condition 1a"
  elseif(y<40)
    print "Condition 1b"
  else
    print "Condition 1c"  
  endif
elseif(x>3)
  print "Condition 2a"
else
  print "Condition 2c"
endif
```

This PR also includes minor bugfixes to `alias()` and `lt()` functions.

### DIFF
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -3032,7 +3032,10 @@ void HcomHandler::executeDefineAliasCommand(const QString cmd)
     QString alias = splitCmd[1];
     variable.remove("\"");
     toLongDataNames(variable);
-    if(mpModel->getViewContainerObject()->setVariableAlias(variable, alias))
+    QStringList hierarchy = variable.section("#",0).split("$");
+    hierarchy.removeLast();
+    SystemObject *pSystem = searchIntoSubsystem(mpModel->getViewContainerObject(),hierarchy);
+    if(pSystem->setVariableAlias(variable, alias))
     {
         HCOMINFO(QString("Sucessfully assigned variable alias %1").arg(alias));
     }
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -7900,7 +7900,7 @@ QString HcomHandler::runScriptCommands(QStringList &lines, bool *pAbort)
                     break;
                 }
             }
-            if(!oneConditionWasTrue)
+            if(!oneConditionWasTrue && !elseCode.isEmpty())
             {
                 QString gotoLabel = runScriptCommands(elseCode, pAbort);
                 if(*pAbort)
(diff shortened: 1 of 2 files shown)


HopsanGUI/HcomHandler.cpp
```diff
@@ -8881,7 +8881,9 @@ void HcomHandler::executeLtBuiltInFunction(QString fnc_call)
             pVar2 = mAnsVector;
         }
 
-        LogDataHandler2 *pLogDataHandler = mpModel->getViewContainerObject()->getLogDataHandler().data();
+        LogDataHandler2 *pLogDataHandler = 0;
+        if(mpModel)
+            pLogDataHandler = mpModel->getViewContainerObject()->getLogDataHandler().data();
```
peterNordin, line 8889: pLogDataHandler is used on line 8907 and 8917, so if it is nullptr it will still crash
But maybe those cant happen if no model i opened, (pVar1,2) will be nullptr

HopsanGUI/HcomHandler.cpp
```diff
@@ -7820,10 +7820,13 @@ QString HcomHandler::runScriptCommands(QStringList &lines, bool *pAbort)
...
             }
-            QString condition = args.front();
-            QStringList ifCode;
+            QList<QString> conditions;
+            conditions.append(args.front());
+            QList<QStringList> ifCodes;
+            ifCodes.append(QStringList());
             QStringList elseCode;
             bool inElse=false;
+            int depth=1;
```
peterNordin, line 7832: Is see that only depth one is processed, but just want to see if i understand correctly.
Nested if statements (depth >=2) would be interpreted runScriptCommands(ifCodes[i] is called.
But every time depth will be at most one.

peterNordin (commented):
Have not had time to test but looks good, if I understood correctly. See other comments.
Maybe add a test case (with nested if) to HcomTest.hpp

HopsanGUI/HcomHandler.cpp
robbr48: I cannot happen, because evaluating arg1 and arg2 will not result in data vector answers unless there is an open model. But I will add checks anyway to get some redundancy.

HopsanGUI/HcomHandler.cpp
```diff
@@ -7833,43 +7836,68 @@ QString HcomHandler::runScriptCommands(QStringList &lines, bool *pAbort)
...
+                    }
+                    if(!gotoLabel.isEmpty())
+                    {
+                        return gotoLabel;
+                    }
+                    break;
                 }
             }
-            else
+            if(!oneConditionWasTrue)
```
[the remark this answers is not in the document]
robbr48: If elseCode is empty there will be nothing to call anyway. It will loop over an empty QStringList. But if you want to I can add a check anyway?

HopsanGUI/HcomHandler.cpp
robbr48: Yes, exactly. Nested loops will be treated iteratively in the next evaluation. This makes it possible to have an infinite depth of nested statements.

PR #2108, opened 2021-11-26
Ignore file change warning after saving the file

robbr48:
Not an ideal solution, but this will make sure the "file has changed on disk" warning does not appear when the file has been saved by Hopsan itself. 

Disconnecting the signal from the slot before saving and then reconnecting afterwards does not work, since there may be a small time delay before Hopsan is notified of the change by the file system.


peterNordin:
I can not reproduce the original problem (KUbuntu) what OS did you run to get the problem.
I assume the problem is saving a file (like a hcom script) inside hopsan and it gets flagged as externally changed on disk?

robbr48:
Yes, exactly. It happens every time you save a file on Windows. I guess different OS use different methods for keeping track on changed files.

PR #2109, opened 2021-11-26
Remember loading HMF components as internal systems when saving

### LINKED ISSUE #2105: Loading HMF components as internal subsystems is forgotten when saving model
**Describe the bug**
If a HMF component is loaded as an internal subsystem, it is reverted to an external subsystem when saving model.

**To Reproduce**
1. Add a HMF component to a model
2. Enter the subsystem of HMF component
3. Click "Load as internal subsystem"
4. Save the model
5. Close the model
6. Load the model again
7. The HMF component is no longer an internal subsystem

**Expected behavior**
The HMF component should remain an internal subsystem after reloading the model.

robbr48:
The problem was that the subsystem component kept its subtypename. When loading the model, the HMF component was then loaded as an external subsystem, ignoring the subsystem contents specified in the HMF file.

Resolves #2105.

### DIFF
--- a/HopsanGUI/Widgets/ModelWidget.cpp
+++ b/HopsanGUI/Widgets/ModelWidget.cpp
@@ -945,6 +945,7 @@ void ModelWidget::openCurrentContainerInNewTab()
             pContainer->setModelFile("");
             lockModelEditingFull(true);
             mpExternalSystemWarningWidget->setVisible(false);
+            pContainer->setSubTypeName("");
             break;
         }
     }


HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -945,6 +945,7 @@ void ModelWidget::openCurrentContainerInNewTab()
             pContainer->setModelFile("");
             lockModelEditingFull(true);
             mpExternalSystemWarningWidget->setVisible(false);
+            pContainer->setSubTypeName("");
```
peterNordin, line 948: I have not looked deep into this problem. But it feels like it would be better to do this in the code that loads and embeds the subsystem. Not here in the code that seems to open it in a new tab. But maybe I need to look deeper.

robbr48: That is actually this code. The name of the function is a bit weird though. As far as I can see it does not open any new tab. It just enables editing, sets the model file to an empty string and hides the external system widget.

Did we at some point change the behavior from opening external models in a new tab to embedding them in the model?

PR #2114, opened 2021-12-21
Import/export parameter data to SSV

robbr48:
- Export model parameters to System Structure Parameter Values (SSV)
- Import model parameters from SSV
- Units are automatically converted whenever possible
- Added SI-unit exponentials to all default base units in Hopsan, according to SSP and FMI standards.

### DIFF
--- a/HopsanGUI/Configuration.cpp
+++ b/HopsanGUI/Configuration.cpp
@@ -1009,17 +1009,17 @@ QString Configuration::getBaseUnit(const QString &rQuantity)
     return  mUnitScales.value(rQuantity, QuantityUnitScale()).baseunit;
 }
 
-void Configuration::getBaseUnitSIExponents(const QString &rQuantity, int &kg, int &m, int &s, int &A, int &K, int &mol, int &cd, int &rad)
+void Configuration::getBaseUnitSIExponents(const QString &rQuantity, int &rKg, int &rM, int &rS, int &rA, int &rK, int &rMol, int &rCd, int &rRad)
 {
     QuantityUnitScale us = mUnitScales.value(rQuantity, QuantityUnitScale());
-    kg = us.kg;
-    m = us.m;
-    s = us.s;
-    A = us.A;
-    K = us.K;
-    mol = us.mol;
-    cd = us.cd;
-    rad = us.rad;
+    rKg = us.kg;
+    rM = us.m;
+    rS = us.s;
+    rA = us.A;
+    rK = us.K;
+    rMol = us.mol;
+    rCd = us.cd;
+    rRad = us.rad;
 }
 
 bool Configuration::isRegisteredBaseUnit(const QString &rUnitName) const
--- a/HopsanGUI/Configuration.cpp
+++ b/HopsanGUI/Configuration.cpp
@@ -249,6 +249,14 @@ void Configuration::saveToXml()
             {
                 xmlQuantity.setAttribute(CFG_BASEUNIT, baseunit);
             }
+            xmlQuantity.setAttribute(CFG_KG, qit.value().kg);
+            xmlQuantity.setAttribute(CFG_KG, qit.value().m);
+            xmlQuantity.setAttribute(CFG_KG, qit.value().s);
+            xmlQuantity.setAttribute(CFG_KG, qit.value().A);
+            xmlQuantity.setAttribute(CFG_KG, qit.value().K);
+            xmlQuantity.setAttribute(CFG_KG, qit.value().mol);
+            xmlQuantity.setAttribute(CFG_KG, qit.value().cd);
+            xmlQuantity.setAttribute(CFG_KG, qit.value().rad);
 
             // Save the default selected display unit
             QMap<QString, QString>::iterator itdu = mSelectedDefaultUnits.find(qit.key());
@@ -665,6 +673,14 @@ void Configuration::loadUnitSettings(QDomElement &rDomElement, bool tagAsBuiltIn
     {
         QString quantity = xmlQuantity.attribute(HMF_NAMETAG);
         QString baseunit = xmlQuantity.attribute(CFG_BASEUNIT);
+        int kg = parseAttributeInt(xmlQuantity, CFG_KG, 0);
+        int m = parseAttributeInt(xmlQuantity, CFG_M, 0);
+        int s = parseAttributeInt(xmlQuantity, CFG_S, 0);
+        int A = parseAttributeInt(xmlQuantity, CFG_A, 0);
+        int K = parseAttributeInt(xmlQuantity, CFG_K, 0);
+        int mol = parseAttributeInt(xmlQuantity, CFG_MOL, 0);
+        int cd = parseAttributeInt(xmlQuantity, CFG_CD, 0);
+        int rad = parseAttributeInt(xmlQuantity, CFG_RAD, 0);
         QString deafdisplayunit = xmlQuantity.attribute(CFG_DEFAULTDISPALYUNIT, baseunit);
         if (!deafdisplayunit.isEmpty())
         {
@@ -681,6 +697,14 @@ void Configuration::loadUnitSettings(QDomElement &rDomElement, bool tagAsBuiltIn
         if (!baseunit.isEmpty() && !qit.value().builtInUnitconversions.contains(baseunit))
         {
             qit.value().baseunit = baseunit;
+            qit.value().kg = kg;
+            qit.value().m = m;
+            qit.value().s = s;
+            qit.value().A = A;
+            qit.value().K = K;
+            qit.value().mol = mol;
+            qit.value().cd = cd;
+            qit.value().rad = rad;
             qit.value().customUnits.insert(baseunit, UnitConverter(quantity, baseunit, "1.0", ""));
             if (tagAsBuiltIn)
             {
@@ -983,6 +1007,19 @@ QStringList Configuration::getQuantitiesForUnit(const QString &rUnit) const
 QString Configuration::getBaseUnit(const QString &rQuantity)
 {
     return  mUnitScales.value(rQuantity, QuantityUnitScale()).baseunit;
+}
+
+void Configuration::getBaseUnitSIExponents(const QString &rQuantity, int &kg, int &m, int &s, int &A, int &K, int &mol, int &cd, int &rad)
+{
+    QuantityUnitScale us = mUnitScales.value(rQuantity, QuantityUnitScale());
+    kg = us.kg;
+    m = us.m;
+    s = us.s;
+    A = us.A;
+    K = us.K;
+    mol = us.mol;
+    cd = us.cd;
+    rad = us.rad;
 }
 
 bool Configuration::isRegisteredBaseUnit(const QString &rUnitName) const
--- a/HopsanGUI/Configuration.cpp
+++ b/HopsanGUI/Configuration.cpp
@@ -249,14 +249,14 @@ void Configuration::saveToXml()
             {
                 xmlQuantity.setAttribute(CFG_BASEUNIT, baseunit);
             }
-            xmlQuantity.setAttribute(CFG_KG, qit.value().kg);
-            xmlQuantity.setAttribute(CFG_KG, qit.value().m);
-            xmlQuantity.setAttribute(CFG_KG, qit.value().s);
-            xmlQuantity.setAttribute(CFG_KG, qit.value().A);
-            xmlQuantity.setAttribute(CFG_KG, qit.value().K);
-            xmlQuantity.setAttribute(CFG_KG, qit.value().mol);
-            xmlQuantity.setAttribute(CFG_KG, qit.value().cd);
-            xmlQuantity.setAttribute(CFG_KG, qit.value().rad);
+            xmlQuantity.setAttribute(cfg::siunits::kg, qit.value().kg);
+            xmlQuantity.setAttribute(cfg::siunits::m, qit.value().m);
+            xmlQuantity.setAttribute(cfg::siunits::s, qit.value().s);
+            xmlQuantity.setAttribute(cfg::siunits::A, qit.value().A);
+            xmlQuantity.setAttribute(cfg::siunits::K, qit.value().K);
+            xmlQuantity.setAttribute(cfg::siunits::mol, qit.value().mol);
+            xmlQuantity.setAttribute(cfg::siunits::cd, qit.value().cd);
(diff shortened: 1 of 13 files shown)


HopsanGUI/GUIObjects/GUIContainerObject.cpp
```diff
@@ -4889,6 +4889,35 @@ void SystemObject::loadParameterValuesFromFile(QString parameterFile)
...
+        values.append(parameter.mValue);
+        dataTypes.append(parameter.mType);
+        quantities.append(parameter.mQuantity);
+        units.append(parameter.mUnit);
+    }
+
+    for(ModelObject *modelObject : getModelObjects()) {
+        if(modelObject->getTypeName() == "Subsystem") {
+            qobject_cast<SystemObject*>(modelObject)->getAllParametersAndValuesRecursively(prefix+modelObject->getName()+"|", names, values, dataTypes, quantities, units);
+            continue;
```
peterNordin, line 4907: Not a big deal, but I would have preferred a good old if ... else,  instead of continue to skip the "else part"

HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -698,16 +698,260 @@ void ModelWidget::saveAs()
...
-void ModelWidget::exportModelParameters()
+void ModelWidget::exportModelParametersToHpf()
 {
     saveModel(NewFile, ParametersOnly);
 }
 
-void ModelWidget::importModelParameters(QString parameterFile)
+void ModelWidget::exportModelParametersToSsv()
+{
+    QMap<QString, QString> dataTypeTranslator{{"double", "Real"}, {"integer", "Integer"}, {"bool", "Boolean"}, {"string", "String"}, {"textblock", "String"}, {"filepath", "Strig"}, {"conditional", "Integer"}};
```
peterNordin, line 708: "filepath", "Strig"

HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -698,16 +698,260 @@ void ModelWidget::saveAs()
...
+    }
+    QString parameterFile = QFileDialog::getOpenFileName(gpMainWindowWidget, tr("Load Parameters from SSV File"),
+                                                 openLocation,
+                                                 tr("System Structure Parameter Values (*.ssv)"));
+    if(!parameterFile.isEmpty()) {
+        gpConfig->setStringSetting(CFG_PARAMETERIMPORTDIR,  QFileInfo(parameterFile).absolutePath());
+    }
+
+    //This map translates data type names from Hopsan to SSV
+    QMap<QString, QString> dataTypeTranslator{{"double", "Real"}, {"integer", "Integer"}, {"bool", "Boolean"}, {"string", "String"}, {"textblock", "String"}, {"filepath", "Strig"}, {"conditional", "Integer"}};
```
peterNordin, line 810: Again {"filepath", "Strig"}
Maybe this translation map could be one global map variable in an anonymous namespace (int this file). But not sure if you can create QMap like that. But it would be better to have one, instead of two copies. Maybe a std::uniqu_ptr or qsharedpointer could be used. Or maybe its simpler this way.

HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -698,16 +698,260 @@ void ModelWidget::saveAs()
...
+                        hopsanUnit = parameter.mUnit;
+                        if(parameter.mQuantity.isEmpty()) {
+                            hopsanQuantity = gpConfig->getQuantitiesForUnit(parameter.mUnit)[0];
+                        }
+                        else {
+                            hopsanQuantity = parameter.mQuantity;
+                        }
+                        UnitConverter uc1 = gpConfig->getUnitScaleUC(ssvQuantity, ssvUnit);         //Converts between SSV unit and base unit
+                        UnitConverter uc2 = gpConfig->getUnitScaleUC(hopsanQuantity, parameter.mUnit); //Converts between Hopsan unit and base unit
+                        pComponent->setParameterValue(parameterName, uc2.convertFromBase(uc1.convertToBase(ssvValue)));
```
peterNordin, line 942: I am unsure about this, but should we really convert from base, should we not always set parameters in the base unit.
Or is the conversion handled inside pComponent->setParameterValue? In this case I believe that pComponent is a GUIComponent. So maybe that is the case.

In the core I think the idea is that we always set parameter values in base unit. But I do not remember really.

peterNordin (changes requested):
Looks fairly OK, I do not have time to test though.
Some minor issues that needs fixing, some optional.

HopsanGUI/Configuration.cpp
```diff
@@ -985,6 +1009,19 @@ QString Configuration::getBaseUnit(const QString &rQuantity)
     return  mUnitScales.value(rQuantity, QuantityUnitScale()).baseunit;
 }
 
+void Configuration::getBaseUnitSIExponents(const QString &rQuantity, int &kg, int &m, int &s, int &A, int &K, int &mol, int &cd, int &rad)
```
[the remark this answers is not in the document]
robbr48: I changed it just to be consistent.

HopsanGUI/Configuration.h
```diff
@@ -112,6 +112,15 @@
 #define CFG_UNITSETTINGS "unitsettings"
 #define CFG_QUANTITY "quantity"
 
+#define CFG_KG "kg"
```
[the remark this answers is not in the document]
robbr48: Good point, didn't think about that. I will also add namespaced constexpr values for the SSV XML strings.

HopsanGUI/Configuration.cpp
```diff
@@ -249,6 +249,14 @@ void Configuration::saveToXml()
             {
                 xmlQuantity.setAttribute(CFG_BASEUNIT, baseunit);
             }
+            xmlQuantity.setAttribute(CFG_KG, qit.value().kg);
```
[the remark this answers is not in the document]
robbr48: Fixed.

HopsanGUI/GUIObjects/GUIContainerObject.cpp
robbr48: You are right. I like to use "continue" in this way sometimes to avoid indenting very long sections of code, but in this case there is no point.

HopsanGUI/Widgets/ModelWidget.cpp
robbr48: I moved the map to `XMLUtilities.h` and made it static (as part of the `ssv`  namesapce)

HopsanGUI/Widgets/ModelWidget.cpp
robbr48: We need to set variable in the unit specified in the GUIComponent. This value is then converted again in the core. There is a slight risk that we lose precision, but I guess that is unavoidable.

peterNordin:
Did you forget to push the changes?

robbr48:
Sorry, I thought I pushed it by my internet connection was a little unstable yesterday. Now all changes should be up-to-date.

HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -698,16 +698,254 @@ void ModelWidget::saveAs()
...
+            QStringList unitQuantities = gpConfig->getQuantitiesForUnit(units[i]);
+            if(!unitQuantities.isEmpty()) {
+                quantity = unitQuantities.first();  //Picking the first quantity is the best we can do without more knowledge
+            }
+        }
+        if(usedUnits.contains(qMakePair(units[i], quantity)) || quantity.isEmpty()) {
+            continue;
+        }
+        usedUnits.append(qMakePair(units[i], quantity));
+        QString baseUnit = gpConfig->getBaseUnit(quantity);
```
peterNordin, line 752: qtcreator reports this line as unused

HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -698,16 +698,254 @@ void ModelWidget::saveAs()
...
+                    if(quantities.isEmpty()) {
+                        gpMessageHandler->addWarningMessage("Unit is not supported by Hopsan: "+ssvUnit+". Setting parameter without unit conversion.");
+                        pSystem->setParameterValue(parameterName, ssvValue);
+                    }
+                    else {
+                        QString ssvQuantity, hopsanUnit, hopsanQuantity;
+                        ssvQuantity = quantities[0];    //Guessing for first quantity is the best we can do
+                        hopsanUnit = parameter.mUnit;
+                        if(parameter.mQuantity.isEmpty()) {
+                            hopsanQuantity = gpConfig->getQuantitiesForUnit(parameter.mUnit)[0];
```
peterNordin, line 887: Must make sure returned list is not empty before taking [0], could be that someone creates a component and make up some custom unit

HopsanGUI/Widgets/ModelWidget.cpp
peterNordin: I think I know why I still feel this is wrong (it probably is correct).
As I remember it, in the core (component c++ implementation)  units "must" be in base units. 

When you add a parameter or in/out variable you either give a unit or a quantity. 
The old way is to give it a unit, but then you can not be sure how to convert it unless there is a unique match among one of the quantities. That is why we have that odd code allover the place that tries to reverse lookup the quantity.

The "new" way was to give it a quantity, then the value is assumed to be in base unit. Because of interoperability with other components. If you add an output variable with quantity "position" than any receiver of that variable must expect "position" and use the agreed upon unit (in this case base-unit) because we use precompiled components so we can not adapt if someone uses for instance mm instead of m.

So in this case if a component is made correctly, the core unit "must" be the base unit of its quantity (if a quantity was specified instead of unit). Therefor the uc2 converter looked wrong to me, because when you set the parameter value here it goes directly into core without conversion. (Core does not know about unit converters (but it would be nice if it did)).

But in this case, the uc2 converter will most often be 1:1 so it does not matter.
And in the case that someone has actually chosen a different unit (using the old method of directly specifying unit), for instance "mm" for a position variable then this uc2 is actually needed.

Does my reasoning make sense?

Extra:
One can also argue whether choosing a variable (start-value) or constant parameter to have a non base unit, is a good or bad idea. For constant parameters it make more sense since they are never connected to anything else.

I think we came to the conclusion a long time ago that it would be nice if you could in the core choose quantity and unit.
At least for constant parameters, like "diameter = Length:mm" to specify both quantity and unit, but then core must handle all conversion (which would be nice) and it would be possible to discover if someone connects "mm" to "m" for instance. But it would be a headache for users if different precompiled components use different units in variable data (signal/power connectors) . 
That is why I think my idea of requiring base units in the core makes sense. You only do the conversion in the GUI when you set variables or analyze results, as a user convenience.

peterNordin (changes requested):
Does my reasoning about the units make sense?
I believe you were right, but need to verify I remember the code correctly.

But you need to fix the potential crash if custom units (without registered quantities) are used

HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -698,16 +698,254 @@ void ModelWidget::saveAs()
...
+                    if(quantities.isEmpty()) {
+                        gpMessageHandler->addWarningMessage("Unit is not supported by Hopsan: "+ssvUnit+". Setting parameter without unit conversion.");
+                        pComponent->setParameterValue(parameterName, ssvValue);
+                    }
+                    else {
+                        QString ssvQuantity, hopsanUnit, hopsanQuantity;
+                        ssvQuantity = quantities[0];    //Guessing for first quantity is the best we can do
+                        hopsanUnit = parameter.mUnit;
+                        if(parameter.mQuantity.isEmpty()) {
+                            hopsanQuantity = gpConfig->getQuantitiesForUnit(parameter.mUnit)[0];
```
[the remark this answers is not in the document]
robbr48: That is already handled by the if statement above.

HopsanGUI/Widgets/ModelWidget.cpp
robbr48: Same as above, we never get here if quantities is empty.

HopsanGUI/Widgets/ModelWidget.cpp
robbr48: I believe reasoning is correct. There could be cases where you really want to have a non-SI unit as a value, for example both Hz and rad/s are valid units for frequency. There are also certain domains, e.g. aerospace engineering, where different units are sometimes used as a way to avoid mixing up different parameters. In such cases the solution with both quantity and unit would make sense. A problem is that units are stored in the configuration file at the moment, which is not read by the core.

HopsanGUI/Widgets/ModelWidget.cpp
robbr48: Ok, I removed it.

HopsanGUI/Widgets/ModelWidget.cpp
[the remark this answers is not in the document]
peterNordin: but in this case we retrieve quantities for parameter.mUnit, as far as I can see above, we get quantities for ssvUnit
They are likely not the same in this case.
Also I noticed now that hopsanUnit is assigned but not used, (as far as I can see in the diff)

HopsanGUI/Widgets/ModelWidget.cpp
peterNordin: but in this case we retrieve quantities for parameter.mUnit, as far as I can see above, we get quantities for ssvUnit
They are not the same in this case due to the if check at line 873. (This check did not exist in the other place)
Also I noticed now that hopsanUnit is assigned but not used, (as far as I can see in the diff)

HopsanGUI/Widgets/ModelWidget.cpp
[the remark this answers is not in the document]
robbr48: Aha, I thought the comment was about line 926. I added a non-empty check for the hopsan quantities as well.

HopsanGUI/Widgets/ModelWidget.cpp
robbr48: Added a check here too.

HopsanGUI/Widgets/ModelWidget.cpp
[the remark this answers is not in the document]
peterNordin: We dont need to do it now, but maybe gpConfig should have a function getFirstMatchingQuantity() that returns the first or empty string. We have these checks all over the place. Something to add in the future.

peterNordin (approved):

Issue #2117, opened 2022-01-20
Open HMF for subsystem components in library


robbr48:
**Is your improvement request related to a problem? Please describe.**
It is difficult to find the HMF file for subsystem components. Right now you have to open the XML file, figure out the HMF file name, and then open it manually from the library folder.

**Describe the solution you'd like**
1. "View source code" in right click menu should be replaced with "Open model file" for such components.
2. Double-clicking the component should open the HMF file (now it opens an empty text editor)

PR #2118, opened 2022-01-21
Open HMF for subsystem components from library widget

robbr48:
- Added "Open model file" option in right-click menu for subsystem components in library widget
- Hide "Open source code" option for components that have no source code specified
- Double-clicking on subsystem components opens model file

### DIFF
--- a/HopsanGUI/Widgets/LibraryWidget.cpp
+++ b/HopsanGUI/Widgets/LibraryWidget.cpp
@@ -423,12 +423,14 @@ void LibraryWidget::handleItemClick(QTreeWidgetItem *item, int column)
         QAction *pOpenFolderAction = contextMenu.addAction("Open Containing Folder");
         QAction *pEditXMLAction = contextMenu.addAction("Edit XML Description");
         QAction *pEditCodeAction = contextMenu.addAction("Edit Source Code");
+        QAction *pOpenModelAction = contextMenu.addAction("Open Model File");
 
         pUnloadAllAction->setVisible(false);
         pUnloadAction->setVisible(false);
         pOpenFolderAction->setVisible(false);
         pEditXMLAction->setVisible(false);
         pEditCodeAction->setVisible(false);
+        pOpenModelAction->setVisible(false);
         pRecompileAction->setVisible(false);
         pReloadAction->setVisible(false);
         pCheckConsistenceAction->setVisible(false);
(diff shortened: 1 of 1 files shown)


HopsanGUI/Widgets/LibraryWidget.cpp
```diff
@@ -461,7 +463,12 @@ void LibraryWidget::handleItemClick(QTreeWidgetItem *item, int column)
         {
             pRecompileAction->setVisible(true);
             pEditXMLAction->setVisible(true);
-            pEditCodeAction->setVisible(true);
+            if(!gpLibraryHandler->getEntry(mItemToTypeNameMap.find(item).value()).pAppearance->getSourceCodeFile().isEmpty()) {
```
peterNordin, line 466: since this long line of code is reused below, cosider saving the result in a bool.
"itemHasSourceCode" and use that bool in the setVisible call

Same for  hmfFile, even though its only used once.

But I could not test this menu item, it never showed up for me. I took the subsystem example model and saved Servo1 as a component inside the defaultLibrary/Special
I can not get this menu option, not sure why, have not had time to debug

HopsanGUI/Widgets/LibraryWidget.cpp
```diff
@@ -696,7 +720,13 @@ void LibraryWidget::handleItemDoubleClick(QTreeWidgetItem *item, int column)
                 basePath.append("/");
             }
             QString sourceFile = appearance->getSourceCodeFile();
-            gpModelHandler->loadTextFile(basePath+sourceFile);
+            if(!sourceFile.isEmpty()) {
```
peterNordin, line 723: I am not able to trigger these double click events on any library tree items. (I dont think they are necessary so no big deal)

peterNordin:
Slightly related issue I noticed, when I saved a subsystem as a component I got to choose the "unique" typename (subtypename)
But i could not chose the "display name" so I saved Servo1, gave it the typename "Monkey" but then it showed up in the library as Servo1, which was the instance name in my original model. (Not sure if it deserves its own issue)
Anyway user should be able to chose default (display)name as well.

HopsanGUI/Widgets/LibraryWidget.cpp
robbr48: They only work for external libraries. I can change it so that they also work for default components. It makes no sense why it should not.

robbr48:
Yes, that deserves its own issue. This one only covers importing such components. There are probably other errors with the export code as well.

peterNordin:
Resolves #2117

PR #2120, opened 2022-02-28
Support SSV parameter sets for components

robbr48:
Parameter sets are specified in component XML files:
```
<?xml version='1.0' encoding='UTF-8'?>
<hopsanobjectappearance version="0.3">
    <modelobject [...]>
        <parametersets>
            <parameterset name="Small Piston" path="smallpiston.ssv"/>
            <parameterset name="Big Piston" path="bigpiston.ssv"/>
        </parametersets>
        [...]
    </modelobject>
</hopsanobjectappearance>
```

The SSV file must be a System Structure Parameter Values file according to the [SSP standard](https://ssp-standard.org/), where values to some or all parameters in the component are stored. 

If parameter sets are specified, they appear in the component properties dialog, where they can be selected from a drop-down menu and then applied with a button. Units are automatically converted depending on what is currently selected in the drop-down menus in the component properties dialog.

This PR also moves the reading of the SSV file to a separate utilty, and adds an SSV parameter specification class.

### DIFF
--- a/HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
+++ b/HopsanGUI/Dialogs/ComponentPropertiesDialog3.cpp
@@ -279,6 +279,51 @@ void ComponentPropertiesDialog3::openDescription()
     pDescriptionWindow->show();
 }
 
+
+//! @brief Applies the parameter set selected in the drop-down menu by reading its SSV file
+void ComponentPropertiesDialog3::applyParameterSet()
+{
+    //Read from SSV file (XML)
+    QString fileName = mpModelObject->getParameterSets().value(mpSetsComboBox->currentText());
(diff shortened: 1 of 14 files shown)


peterNordin (commented):
Code looks OK

But please add the new CAF xml to the CAF xml description into some file in the documentation.
There is a doxygen file with and xml format description, dont remember which one tight now.

Could we use the existing hpf files as well ? Since we alrady have them for importing or exporting parameter sets? (But maybe SSV should replace the old format eventually)

(for later) Could SSV be embedded directly in the Caf XML ? So that external files would not be needed?

(for later) I guess right now you gave to manually add them to the caf xml file. It would be nice if you could add sets automatically from inside hopsan gui somehow.

Issue #2122, opened 2022-03-04
Model changes behaviour after disconnecting and reconnecting


robbr48:
**Describe the bug**
A SignalSum is connected to two AngleTransducers. It gives the wrong output in one model, but correct in an identical reference model. Removing and re-connecting one of the connections makes the output correct again.

**To Reproduce**
Not reproducable. Seems to be different behaviours on different computers.

**Hopsan version (please complete the following information):**
 - OS: Windos 10
 - Version: 2.19.1
 - Build: Confirmed both on official build and development build

robbr48:
I found the error in the sort order of signal components. It seems like ReadMultiPorts are not sorted properly. It only uses the first connected port, and ignores the rest.

peterNordin:
I was looking at the code and trying to understand what happened, did it completely ignore one of the components or did it just sort it wrongly randomly ?

robbr48:
It only considered the first connection to the multi port. The rest of the connections were ignored. If you were lucky those component were sorted b efore the multiport component, but it depends on the order the components and/or the connections are added to the model.

peterNordin:
Closing manually, ("Resolvs" was misspelled)

PR #2123, opened 2022-03-04
Fix sorting of ReadMultiPort with multiple connections

robbr48:
Take all connections into account when sorting signal components with ReadMultiPort.

Resolvs #2122.

### DIFF
--- a/HopsanCore/src/ComponentSystem.cpp
+++ b/HopsanCore/src/ComponentSystem.cpp
@@ -701,66 +701,63 @@ bool ComponentSystem::sortComponentVector(std::vector<Component*> &rComponentVec
     }
 #else
     bool didSomething = true;
-    while(didSomething)
-    {
+    while(didSomething) {
         didSomething = false;
 
         // Loop through the unsorted component vector
-        for(size_t c=0; c<rComponentVector.size(); ++c)
-        {
+        for(size_t c=0; c<rComponentVector.size(); ++c) {
             // A pointer to an unsorted component
             Component* pUnsrtComp = rComponentVector[c];
             //Ignore components that are already added to the new vector
-            if(!vectorContains<Component*>(newComponentVector, pUnsrtComp))
-            {
+            if(!vectorContains<Component*>(newComponentVector, pUnsrtComp)) {
                 bool readyToAdd=true;
                 std::vector<Port*> portVector = pUnsrtComp->getPortPtrVector();
                 // Ask each read port for its node, then ask the node for its write port component
-                for(size_t p=0; p<portVector.size(); ++p)
-                {
+                for(size_t p=0; p<portVector.size(); ++p) {
                     // Take the port pointer (To make code easier to read)
                     Port *pPort = portVector[p];
-                    Component* pRequiredComponent=0;
+                    std::vector<Component*> pRequiredComponents;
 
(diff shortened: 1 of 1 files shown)


HopsanCore/src/ComponentSystem.cpp
```diff
@@ -701,66 +701,67 @@ bool ComponentSystem::sortComponentVector(std::vector<Component*> &rComponentVec
...
-                        {
-                            if(!vectorContains<Component*>(newComponentVector, pRequiredComponent->mpSystemParent) &&
-                                    (pRequiredComponent->mpSystemParent->getTypeCQS() == pPort->getComponent()->getTypeCQS()) &&
-                                    vectorContains<Component*>(rComponentVector,pRequiredComponent->mpSystemParent))
-                            {
-                                readyToAdd = false;     //Depending on subsystem component which has not yet been added
+                    }
+                    if(!pRequiredComponents.empty()) {
+                        for(size_t r=0; r<pRequiredComponents.size(); ++r) {
+                            if(pRequiredComponents[r]->mpSystemParent == this) {
```
peterNordin, line 745: Not a big deal in this case, but I would prefer adding `Component* pRequiredComponent=pRequiredComponents[r];` and then using pRequiredComponent (like before). Since only the pointer at location r is checked every loop.

HopsanCore/src/ComponentSystem.cpp
```diff
@@ -701,66 +701,67 @@ bool ComponentSystem::sortComponentVector(std::vector<Component*> &rComponentVec
...
 
-                    if ( (sortHint == Destination) && pPort->isConnected() )
-                    {
-                        Port *pSourcePort = pPort->getNodePtr()->getSortOrderSourcePort();
-                        if (pSourcePort && pSourcePort->getComponent())
-                        {
-                            pRequiredComponent = pSourcePort->getComponent();
+                    if ( (sortHint == Destination) && pPort->isConnected() ) {
+                        std::vector<Port *> sourcePorts;
+                        if(pPort->isMultiPort()) {
```
peterNordin, line 729: I think that the for loop would be enough, and not need to check if multiport or not. The getNumPorts returns 1 for ordinary ports. But no need to change, I think this is more clear.

peterNordin:
It would be nice with some sort of test. Preferably "unit test" but we do not have any for that code so maybe in this case add it to the ComponentTest math functions, and create a validation data set. Not sure how much work is needed though.

peterNordin:
Consider my comments, otherwise OK

HopsanCore/src/ComponentSystem.cpp
robbr48: True, it will work without the check. It is a bit ugly though that `getNodePtr(const size_t subPortIdx)` will ignore the index argument for non-multiports. This means that `getNodePtr(42) = getNodePtr(0) = getNodePtr()`. But perhaps this was the intended behaviour?

robbr48:
I would like a test for this, but I have not been able to create a non-working example model. I have a model from a user that fails, but have not yet been able to find what actually triggers the bug.

HopsanCore/src/ComponentSystem.cpp
peterNordin: yes it was intended by its not very nice behavior, should redesign at some point

PR #2125, opened 2022-03-16
Steady-state identification

robbr48:
- New component: Steady-state identifier
- New HCOM function: `ssi()`
- New HCOM function: `cut()`

The ssi() function is used for offline steady-state identification of a vector, while the component can be used for online steady-state identification. 

The cut function can remove samples from a vector depending on another vector (of zeros and ones). This is useful for displaying steady-state results in a plot, e.g. by plotting the original vector and the cut vector in the same plot to visualize where it is a steady-state.

### DIFF
--- a/HopsanCore/include/ComponentUtilities/WhiteGaussianNoise.h
+++ b/HopsanCore/include/ComponentUtilities/WhiteGaussianNoise.h
@@ -42,7 +42,7 @@ namespace hopsan {
     class HOPSANCORE_DLLAPI WhiteGaussianNoise
     {
     public:
-        double getValue();
+        static double getValue();
     };
 }
 
--- a/HopsanGUI/HcomHandler.cpp
+++ b/HopsanGUI/HcomHandler.cpp
@@ -397,6 +397,8 @@ HcomHandler::HcomHandler(TerminalConsole *pConsole) : QObject(pConsole)
     registerInternalFunction("bdf1", "Transforms derivatives in a symbolic expression using the first order backward differentiation formula (implicit Euler)", "Usage: e = bdf1(f)");
     registerInternalFunction("bdf2", "Transforms derivatives in a symbolic expression using the second order backward differentiation formula", "Usage: e = bdf2(f)");
     registerInternalFunction("latex", "Prints a symbolic expression with LaTeX syntax", "Usage: latex(e)");
+    registerInternalFunction("cut", "Removes all samples x[i] for which y[i]<0.5", "Usage: cut(x, y)");
+    registerInternalFunction("ssi", "Identifies steady-state for specified variable", "Usage: ssi(vector, method, arguments)\n       Method 0 (rectangular window):\n         ssi(vector, 0, tolerance, windowlength)\n       Method 1 (ratio of differently estimated variances):\n         ssi(vector, 1, tolerance, windowlength, noiseamplitude\n       Method 2 (ratio of differently estimated variances using weighted moving average):\n         ssi(vector, 2, tolerance, lambda1, lambda2, lambda3, noiseamplitude)");
 
     //Setup local function pointers (used to evaluate expressions in SymHop)
     registerFunctionoid("aver", new HcomFunctionoidAver(this), "Calculate average value of vector", "Usage: aver(vector)");
@@ -7381,6 +7383,16 @@ void HcomHandler::evaluateExpression(QString expr, VariableType desiredType)
             return;
         }
     }
+    else if(desiredType != Scalar && isHcomFunctionCall("cut", expr))
+    {
+        executeCutBuiltInFunction(expr);
+        return;
+    }
+    else if(desiredType != Scalar && isHcomFunctionCall("ssi", expr))
+    {
+        executeSsiBuiltInFunction(expr);
+        return;
+    }
     //timer.toc("Vector functions", 1);
 
     //Evaluate expression using SymHop
@@ -8755,9 +8767,9 @@ bool HcomHandler::evaluateArithmeticExpression(QString cmd)
     return false;
 }
 
-void HcomHandler::executeGtBuiltInFunction(QString fnc_call)
+void HcomHandler::executeGtBuiltInFunction(QString functionCall)
 {
-    QStringList args = extractFunctionCallExpressionArguments(fnc_call);
+    QStringList args = extractFunctionCallExpressionArguments(functionCall);
     if(args.size()==2)
     {
         const QString &arg1 = args[0];
@@ -8862,9 +8874,9 @@ void HcomHandler::executeGtBuiltInFunction(QString fnc_call)
     }
 }
 
-void HcomHandler::executeLtBuiltInFunction(QString fnc_call)
+void HcomHandler::executeLtBuiltInFunction(QString functionCall)
 {
-    QStringList args = extractFunctionCallExpressionArguments(fnc_call);
+    QStringList args = extractFunctionCallExpressionArguments(functionCall);
     if(args.size()==2)
     {
         const QString &arg1 = args[0];
@@ -8969,10 +8981,10 @@ void HcomHandler::executeLtBuiltInFunction(QString fnc_call)
     }
 }
 
-void HcomHandler::executeEqBuiltInFunction(QString fnc_call)
+void HcomHandler::executeEqBuiltInFunction(QString functionCall)
 {
     double eps = 1e-3;
-    QStringList args = extractFunctionCallExpressionArguments(fnc_call);
+    QStringList args = extractFunctionCallExpressionArguments(functionCall);
 
     if (args.size() == 3)
     {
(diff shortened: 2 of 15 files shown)


HopsanCore/include/ComponentUtilities/WhiteGaussianNoise.h
```diff
@@ -42,7 +42,7 @@ namespace hopsan {
     class HOPSANCORE_DLLAPI WhiteGaussianNoise
     {
     public:
-        double getValue();
+        static double getValue();
```
peterNordin, line 45: Good, but this should not even be a class, just a function would do. But that could break stuff so lets keep it this way for now

HopsanGUI/HcomHandler.cpp
```diff
@@ -9091,6 +9103,197 @@ void HcomHandler::executeEqBuiltInFunction(QString fnc_call)
...
+
+    VariableTypeT type1 = pVar1->getVariableType();
+    VariableTypeT type2 = pVar2->getVariableType();
+    if(type2 == ComplexType) {
+        HCOMERR("Second variable must not be of complex type.");
+        mAnsType = Undefined;
+        return;
+    }
+
+    mAnsVector = mpModel->getLogDataHandler()->createOrphanVariable(pVar2->getFullVariableName()+"_cut", type1);
```
peterNordin, line 9149: Does not matter since temporary variable, but would not pVar1->fullname make more sense

HopsanGUI/LogVariable.cpp
```diff
@@ -805,6 +805,130 @@ SharedVectorVariableT VectorVariable::toFrequencySpectrum(const SharedVectorVari
...
+            if(time[i] < win) {
+                pRetVar->append(time[i], 0);
+                nw = i+1;  //Update number of samples for window
+                continue;
+            }
+
+            //Compute first variance
+            double xmax = data[i-nw];
+            double xmin = data[i-nw];
+            for(int j=i-nw+1; j<i; ++j) {
```
peterNordin, line 828: I am not exactly sure what is going on here, maybe it must be this way, but this loop in a loop risk becoming very slow for very large time vectors. Especially if i  nw is o low number.

HopsanGUI/LogVariable.cpp
```diff
@@ -805,6 +805,130 @@ SharedVectorVariableT VectorVariable::toFrequencySpectrum(const SharedVectorVari
...
+            if(time[i] < win) {
+                pRetVar->append(time[i], 0);
+                nw = i+1;  //Update number of samples for window
+                continue;
+            }
+
+
+            //Compute average of window
+            double mean = 0;
+            for(int j=i-nw; j<i; ++j) {
```
peterNordin, line 865: Same here for this and the following for loop. They will become slower and slower as j grows.
Maybe it has to be this way?

HopsanGUI/LogVariable.cpp
```diff
@@ -805,6 +805,130 @@ SharedVectorVariableT VectorVariable::toFrequencySpectrum(const SharedVectorVari
...
+                mean += x[j];
+            }
+            mean /= nw;
+
+            //Compute first variance
+            double s1 = 0;
+            for(int j=i-nw; j<i; ++j) {
+                s1 += (x[j]-mean)*(x[j]-mean);
+            }
+            s1 /= (nw/*-1.0*/);
```
peterNordin, line 875: here -1.o is commented, but below on s2 it is not ?

HopsanGUI/LogVariable.cpp
```diff
@@ -805,6 +805,130 @@ SharedVectorVariableT VectorVariable::toFrequencySpectrum(const SharedVectorVari
...
+
+            if(s1/s2 < tol) {
+                pRetVar->append(time[i], 1);
+            }
+            else {
+                pRetVar->append(time[i], 0);
+            }
+            xold = x;
+        }
+    }
```
peterNordin, line 925: I know it would be time consuming, but for all of these mathematical algorithms it would be nice with some, comments explaining the algorithms, possibly with embedded equations (maybe latex). I dont know the math behind these algorithms so I can not properly review them. And in the future if someone looks at this code some explanation would be nice so that the intended behavior would be clear. If you can add something like that it would be good. (but not required). I guess an alternative would be to look up the methods on Wikipedia or similar.

HopsanGUI/Utilities/GUIUtilities.cpp
```diff
@@ -688,8 +688,14 @@ double firstTimeAt(QVector<double> vData, QVector<double> vTime, double value)
...
 double normalDistribution(double average, double sigma)
 {
-    double U1 = (double)rand() / (double)RAND_MAX;
-    double U2 = (double)rand() / (double)RAND_MAX;
+    double U1 = 0;
+    double U2 = 0;
+    while(U1 == 0) {
+        U1 = (double)rand() / (double)RAND_MAX;
+    }
+    while(U2 == 0) {
```
peterNordin, line 696: OK U1 should not be zero, but why not allow U2 to be zero? not sure if it matters.

HopsanGUI/LogVariable.cpp
```diff
@@ -805,6 +805,130 @@ SharedVectorVariableT VectorVariable::toFrequencySpectrum(const SharedVectorVari
...
+    SharedVectorVariableT pRetVar = mpParentLogDataHandler->createOrphanVariable(this->getFullVariableName()+"_ss", TimeDomainType);
+
+    SharedVectorVariableT pTime = this->getSharedTimeOrFrequencyVector();
+    QVector<double> time = pTime->getDataVectorCopy();
+    QVector<double> data = this->getDataVectorCopy();
+
+    if(method == RectangularWindowTest) {
+        int nw = 0;
+        for(int i=0; i<time.size(); ++i) {
+            if(time[i] < win) {
```
peterNordin, line 819: Is win the start of the window or end of the window, maybe clarify variable names to make code easier to understand

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.hpp
```diff
@@ -0,0 +1,217 @@
...
+            addInputVariable("sd", "Standard deviation of white noise (for methods 2 and 3)", "", 0.01, &mpSd);
+            addInputVariable("l1", "Filter factor 1 (for method 3)", "", 0.1, &mpL1);
+            addInputVariable("l2", "Filter factor 2 (for method 3)", "", 0.1, &mpL2);
+            addInputVariable("l3", "Filter factor 3 (for method 3)", "", 0.1, &mpL3);
+        }
+
+
+        void initialize()
+        {
+            mWindowId = 0;
```
peterNordin, line 87: Does this mean that the window always begins from the beginning

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.hpp
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+
+        void simulateOneTimestep()
+        {
+            if(mMethod == 0) {  //Rectangular sliding window
+                mWindow[mWindowId] = *mpIn;
+                mWindowId++;
+                if(mWindowId > mWindow.size()-1) {
+                    mWindowId = 0;
```
peterNordin, line 109: Does this mean that the window repeats itself ? 
(I am trying to understand how it works)

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.hpp
```diff
@@ -0,0 +1,217 @@
...
+                }
+
+                double xmax = mWindow[0];
+                double xmin = mWindow[0];
+                for(int i=0; i<mWindow.size(); ++i) {
```
peterNordin, line 114: Risk for slow simulation, loops over entire window every timestep, maybe this is needed, or maybe this could be made some other way. Maybe update min/max value in window every time a new sample is added/ removed. 
But OK a see the problem, when a value slides out of the window, how should we then update min/max values.

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.hpp
```diff
@@ -0,0 +1,217 @@
...
+                //This requires iterating in correct order
+                double s2 = 0;
+                size_t i = mWindowId;
+                size_t j = (i == mWindow.size()-1) ? 0 : mWindowId+1;
+                for(size_t k=0; k<mWindow.size()-1; ++k) {
+                    s2 += (randWindow[j]-randWindow[i])*(randWindow[j]-randWindow[i]);
+                    i = (i == mWindow.size()-1) ? 0 : i+1;
+                    j = (i == mWindow.size()-1) ? 0 : i+1;
+                }
+                s2 /= (double)randWindow.size();
```
peterNordin, line 167: here there is no -1 as was the case in the gui equivalent code

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.md
```diff
@@ -0,0 +1,52 @@
...
+* **wl** - Length of sliding window (for methods 1 and 2) [-]
+* **sd** - Standard deviation of white noise (for methods 2 and 3) [-]
+* **l1** - Filter factor 1 (for method 3) [-]     
+* **l2** - Filter factor 2 (for method 3) [-]       
+* **l3** - Filter factor 3 (for method 3) [-]       
+
+#### Output Variables
+* **out** - Output value (1 if steady-state, else 0) [-]
+
+### Theory
```
peterNordin, line 19: Aha, here are the equations and explanation, should have looked here first

HopsanGUI/HcomHandler.cpp
robbr48: Of course, that was what was intended.

robbr48:
General comment on performance: Sliding window methods will always be slow, since they need to loop over all variables in Window for every sample. This is why the third method (moving average) was invented. The drawback is that it is harder to parameterize and less intuitive.

HopsanCore/include/ComponentUtilities/WhiteGaussianNoise.h
robbr48: I agree. The problem is if someone is using this in their own libraries. We could keep it as a class and add standard deviation as a member, since you normally want to rescale the noise anyway. But that does not belong in this PR.

HopsanGUI/LogVariable.cpp
robbr48: I checked the references, and it should not be commented. I will fix it.

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.hpp
robbr48: The window is a circular buffer, where we add a new value and remove the oldest one at every step (for performance reasons).

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.hpp
robbr48: This only means that we start on index 0 in the circular buffer (see below).

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.hpp
robbr48: Yes, it would not work if the discarded value is the maximum. Perhaps we could optimize this by checking if the max/min needs to be updated or not, but it would risk making the code overcomplicated.

componentLibraries/defaultLibrary/Signal/Logic/SignalSteadyStateIdentifier.hpp
robbr48: I added it, according to the references.

HopsanGUI/Utilities/GUIUtilities.cpp
robbr48: You are right, I will fix it. The equivalent function in HopsanCore only checked U1.

HopsanGUI/LogVariable.cpp
robbr48: `win` is the length of the sliding window. This check makes sure we wait until the entire window is filled with values before we start looking for steady-state. I added a comment block for the function which explains the arguments.

HopsanGUI/LogVariable.cpp
robbr48: The problem is that the even if the theory is in the markdown file for the component, the user will not know where to look for it when trying to understand the function in HCOM. But that problem can't be easily solved I think.

peterNordin:
Maybe add a ComponentTest model, that shows the three methods in action

PR #2126, opened 2022-03-18
Real-time simulation in HopsanCore

robbr48:
- Functions in HopsanCore for starting and stopping detached realtime simulations
- Relative realtime simulation speed can be controlled using a `realtimeFactor` argument
- Access functions in HopsanGUI CoreAccess classes
- Replaces the old (very inefficient) simulation for real-time animation

This feature only works if HopsanCore is compiled with C++11 or above. I could not find any easy way of achieving the same functionality using old C++. 

Performance of realtime animation is greatly improved.

### DIFF
--- a/HopsanCore/include/CoreUtilities/MultiThreadingUtilities.h
+++ b/HopsanCore/include/CoreUtilities/MultiThreadingUtilities.h
@@ -107,6 +107,8 @@ HOPSANCORE_DLLAPI void simSlave(ComponentSystem *pSystem, std::vector<Component*
                                 double timeStep, size_t numSimSteps, BarrierLock *pBarrier_S,
                                 BarrierLock *pBarrier_C, BarrierLock *pBarrier_Q, BarrierLock *pBarrier_N);
 
+HOPSANCORE_DLLAPI void simWholeSystemInRealtime(double realTimeFactor, volatile bool *pStopSimulation, double *pTime, double timeStep, std::vector<Component *> signalComponentPtrs, std::vector<Component *> cComponentPtrs, std::vector<Component *> qComponentPtrs);
+
 HOPSANCORE_DLLAPI void simWholeSystems(std::vector<ComponentSystem *> systemPtrs, double stopTime);
 
 
--- a/HopsanCore/src/ComponentSystem.cpp
+++ b/HopsanCore/src/ComponentSystem.cpp
@@ -37,6 +37,7 @@
 #include <iostream>
 #include <algorithm>
 #include <map>
+#include <time.h>
 
 #include "ComponentSystem.h"
 #include "HopsanEssentials.h"
@@ -54,6 +55,8 @@ using namespace std;
 #if (__cplusplus >= 201103L)
 #ifdef _WIN32
 #include <windows.h>
+#else
+#include <unistd.h>
 #endif
 #include <functional>
 #include <chrono>
@@ -3345,8 +3348,7 @@ void ComponentSystem::simulate(const double stopT)
     //Simulate
     for (size_t i=0; i<numSimulationSteps; ++i)
     {
-        if (mStopSimulation)
-        {
+        if (mStopSimulation) {
             break;
         }
 
@@ -3376,6 +3378,18 @@ void ComponentSystem::simulate(const double stopT)
 
         logTimeAndNodes(mTotalTakenSimulationSteps);
     }
+}
+
+bool ComponentSystem::startRealtimeSimulation(double realTimeFactor)
+{
+#if (__cplusplus >= 201103L)
+    std::thread rtThread(simWholeSystemInRealtime, realTimeFactor, &mStopSimulation, &mTime, mTimestep, mComponentSignalptrs, mComponentCptrs, mComponentQptrs);
+    rtThread.detach();
+    return true;
+#else
+    stopSimulation("Real-time simulation requires C++11 or above.");
+    return false;
+#endif
 }
 
 
--- a/HopsanCore/src/CoreUtilities/MultiThreadingUtilities.cpp
+++ b/HopsanCore/src/CoreUtilities/MultiThreadingUtilities.cpp
@@ -36,6 +36,13 @@
 #include <unistd.h>
 #endif
 
+#if __cplusplus >= 201103L
+#include <mutex>
+#include <chrono>
+#include <ctime>
+#include <thread>
+#endif
+
 #include "CoreUtilities/MultiThreadingUtilities.h"
 #include "ComponentSystem.h"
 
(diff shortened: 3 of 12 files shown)


HopsanCore/include/CoreUtilities/MultiThreadingUtilities.h
```diff
@@ -38,6 +38,13 @@
 #include <algorithm>
 #include "win32dll.h"
 
+#if __cplusplus >= 201103L
+#include <mutex>
+#include <chrono>
+#include <ctime>
+#include <thread>
+#endif
```
peterNordin, line 46: This block of code does not seem to affect anything inside this header file. Unneeded or maybe should be placed someplace else ? (preferably inside a cpp file)

HopsanCore/src/ComponentSystem.cpp
```diff
@@ -3345,11 +3348,10 @@ void ComponentSystem::simulate(const double stopT)
     // Round to nearest, we may not get exactly the stop time that we want
     size_t numSimulationSteps = calcNumSimSteps(mTime, stopT); //Here mTime is the last time step since it is not updated yet
 
-    //Simulate
+     //Simulate
```
peterNordin, line 3351: Leading space ?

HopsanCore/src/ComponentSystem.cpp
```diff
@@ -3381,6 +3383,18 @@ void ComponentSystem::simulate(const double stopT)
...
 }
 
+bool ComponentSystem::startRealtimeSimulation(double realTimeFactor)
+{
+#if (__cplusplus >= 201103L)
+    std::thread rtThread(simWholeSystemInRealtime, realTimeFactor, &mStopSimulation, &mTime, mTimestep, mComponentSignalptrs, mComponentCptrs, mComponentQptrs);
+    rtThread.detach();
+    return true;
+#else
+    stopSimulation("Real-time simulation requires C++11 or above.");
```
peterNordin, line 3393: Yes all of our threading requires c++11 or above so no problem

HopsanCore/src/CoreUtilities/MultiThreadingUtilities.cpp
```diff
@@ -617,6 +617,41 @@ void simWholeSystems(std::vector<ComponentSystem *> systemPtrs, double stopTime)
...
+{
+#if (__cplusplus >= 201103L)
+    auto wallTime = std::chrono::steady_clock::now();
+
+    while(!(*pStopSimulation)) {
+        *pTime += timeStep; //mTime is updated here before the simulation,
+        //mTime is the current time during the simulateOneTimestep
+
+        //Signal components
+        for (const auto sComponent : signalComponentPtrs) {
```
peterNordin, line 633: not sure if it matters since pointers, but maybe const auto&

HopsanCore/src/CoreUtilities/SimulationHandler.cpp
```diff
@@ -102,6 +102,16 @@ bool SimulationHandler::simulateSystem(const double startT, const double stopT,
     return false;
 }
 
+bool SimulationHandler::startRealtimeSimulation(const double startT, ComponentSystem *pSystem, double realtimeFactor)
```
peterNordin, line 105: startT not used ?, also no need to make it const, functions taking value by copy should be allowed to do what they want with it

HopsanGUI/CoreAccess.cpp
```diff
@@ -200,6 +200,16 @@ bool CoreSimulationHandler::simulate(const double startTime, const double stopTi
     return gHopsanCore.getSimulationHandler()->simulateSystem(startTime, stopTime, nThreads, coreSystems, modelHasNotChanged, algorithm);
 }
 
+bool CoreSimulationHandler::startRealtimeSimumlation(const double startTime, CoreSystemAccess *pCoreSystemAccess, double realtimeFactor)
+{
+    return gHopsanCore.getSimulationHandler()->startRealtimeSimulation(startTime, pCoreSystemAccess->getCoreSystemPtr(), realtimeFactor);
```
peterNordin, line 205: Startime was never used, (as far as I could see)

HopsanCore/src/ComponentSystem.cpp
```diff
@@ -3381,6 +3383,18 @@ void ComponentSystem::simulate(const double stopT)
     }
 }
 
+bool ComponentSystem::startRealtimeSimulation(double realTimeFactor)
+{
+#if (__cplusplus >= 201103L)
+    std::thread rtThread(simWholeSystemInRealtime, realTimeFactor, &mStopSimulation, &mTime, mTimestep, mComponentSignalptrs, mComponentCptrs, mComponentQptrs);
+    rtThread.detach();
```
peterNordin, line 3390: Minor risk, but maybe up to application, when we start a detached simulation like this we do not know if it is running or what it is doing. We could still start ordinary simulation or another copy of detached rt simulation. That would use the same model at the same time.
And what happens if a thread locks up, we can not be sure that it is complete. (Except look at end time) But I dont know if it would make much difference if we had a handle to the thread, we cant do anything with it anyway,
So I guess its OK

HopsanCore/src/CoreUtilities/MultiThreadingUtilities.cpp
robbr48: Of course, it was already used for qComponent. I don't think it matters for pointers, but it should be consistent for s, c and q.

HopsanCore/include/CoreUtilities/MultiThreadingUtilities.h
robbr48: You are right, I moved it to the cpp file.

HopsanCore/src/ComponentSystem.cpp
robbr48: Fixed.

HopsanCore/src/CoreUtilities/SimulationHandler.cpp
robbr48: Ok, I will remove it. The `startT` argument in `simulateSystem` is needed only for multi-threaded simulations, because they need to reinitialize the model after measuring simulation time. If we want parallel realtime simulation in the future we could need to add it to `startRealtimeSimulation` again.

HopsanGUI/CoreAccess.cpp
robbr48: See above. I removed it here as well.

HopsanCore/src/ComponentSystem.cpp
robbr48: Similar problems exist with normal simulations in nonblocking mode. If a component e.g. gets stuck in a loop you need to restart Hopsan, because the simualte mutex will never be unlocked.

I will move the start and stop functions to ModelWidget though, so that we can at least lock and unlock the simualte mutex to avoid starting a second simulation while the first is still running.

PR #2129, opened 2022-03-30
Fix add custom node

peterNordin:


peterNordin:
@robbr48 I have not had time to rebase this properly, so I only need review of the two last commits (Excluding the test commit).
If you have the time maybe test, and do you thing there are more places we need to add hacks for custom nodes, I know we have some places with hard-coded node type checks. Maybe just wait and see.

HopsanGUI/LibraryHandler.cpp
```diff
@@ -82,6 +82,15 @@
 
 namespace  {
 
+QString visibleFilePath(const QFileInfo & fi) {
```
robbr48, line 85: Perhaps add a description comment to this function. I didn't understand the purpose of it until I read the commit message.

robbr48 (commented):
Looks good I think. As far as I can find, node types are hard-coded in losses calculation, component code generator dialog, and when importing FMUs with a TLM specification file. We should fix this later, but it is not urgent. It should always use the registered types of course.

HopsanGUI/LibraryHandler.cpp
peterNordin: Yes, will do

peterNordin:
@robbr48 I did some additional fixes to make it pass CI
And make sure that temp-files from CSV parser are removed, at least on linux.
Could not make it work on Windows, but at least they are always placed in the temp directory so they will disappear eventually.
On Linux they sometimes (or always, not sure) show up in the working directory.

PR #2131, opened 2022-05-02
Fix missing finalize of real-time animation

robbr48:
Just add a missing call to `finalize()`.

Related to #2126.


HopsanGUI/Widgets/ModelWidget.cpp
```diff
@@ -696,6 +696,7 @@ void ModelWidget::stopRealtimeSimulation()
 {
     CoreSimulationHandler mCoreSimulationHandler;
     mCoreSimulationHandler.stopRealtimeSimulation(mpToplevelSystem->getCoreSystemAccessPtr());
+    mCoreSimulationHandler.finalize(mpToplevelSystem->getCoreSystemAccessPtr());
     mSimulateMutex.unlock();
```
peterNordin, line 700: Just a thought, where is the system initialized when running rt simualtion ?

peterNordin (approved):

PR #2133, opened 2022-05-23
Limit time vector length if simulation was stopped before reaching stop time

### LINKED ISSUE #2132: Spectrum/FFT dialog Crashes GUI if simulation was stopped by stop component
**Describe the bug**
FFT/Spectrum dialog causes program to crash if simulation run was stopped by component.

**To Reproduce**

1. Add component and connect as per screenshot:
image
4. Simulate the model
5. View the scope graph, use the FFT button to view spectrum.
7. The program will crash

**Expected behavior**
I would expect a spectrum of the signal up to the point the simulation was stopped to be displayed, as with the signal graph

1. Add component and connect as per screenshot:
image
4. Simulate the model
5. View the scope graph, use the FFT button to view spectrum.
**Screenshots, Model or Data files**
*If applicable, add screenshots, models or data files, that display or cause the bug.*

*Version information is shown in the message widget when HopsanGUI is started*
 - OS: Windows 10
 - Version: HopsanCore: 2.19.1.20211105.2150
                HopsanGUI : 2.19.1.20211105.2150
 - Build: Official build

**Additional information**
*Add any other context/information about the problem here.*
[min_crash_stopped_sim.zip](https://github.com/Hopsan/hopsan/files/8714505/min_crash_stopped_sim.zip)

robbr48:
Restrict length of time vector in log data hadnler to actually number of log samples. This is already done for data variables. 

Having different length of time and data vector causes FFT analysis (and perhaps other features) to crash.

Resolves #2132.

### DIFF
--- a/HopsanGUI/LogDataHandler2.cpp
+++ b/HopsanGUI/LogDataHandler2.cpp
@@ -1185,6 +1185,7 @@ bool LogDataHandler2::collectLogDataFromSystem(SystemObject *pCurrentSystem, con
 
     // Store the systems own time vector
     auto pCoreSysTimeVector = pCurrentSystem->getCoreSystemAccessPtr()->getLogTimeData();
+    pCoreSysTimeVector->resize(pCurrentSystem->getCoreSystemAccessPtr()->getCoreSystemPtr()->getNumActuallyLoggedSamples());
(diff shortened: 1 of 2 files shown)


peterNordin:
I made some additional changes, its better to limit the copies of the time vector data instead of the actual storage vector

PR #2137, opened 2022-08-12
Only expand powers for integer exponents in SymHop

robbr48:
`x^3` should be expanded to `x*x*x` to improve performance, but `x^2.5` should obviously not be expanded.

### DIFF
--- a/SymHop/src/SymHop.cpp
+++ b/SymHop/src/SymHop.cpp
@@ -1619,6 +1619,10 @@ bool Expression::isNumericalSymbol() const
     return (!mString.isEmpty() && (mString[0].isNumber() || mString == "-1.0"));
 }
 
+bool Expression::isInteger() const
+{
+    return (this->isNumericalSymbol()) && (trunc(this->toDouble()) == this->toDouble());
+}
 
 //! @brief Tells whether or not this is a variable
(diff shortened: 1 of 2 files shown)


SymHop/src/SymHop.cpp
```diff
@@ -2624,7 +2628,7 @@ void Expression::expand(const ExpressionSimplificationT simplifications)
 //Expand power functions from e.g. "pow(x,3)" to "x*x*x". This will improve performance in generated code.
 void Expression::expandPowers()
 {
-    if(this->isPower() && this->getPower()->isNumericalSymbol()) {
+    if(this->isPower() && this->getPower()->isInteger()) {
         int nFactors = int(mpPower->toDouble());
```
peterNordin, line 2632: This probably works `int(mpPower->toDouble());` but i would feel better if it was `int(mpPower->toDouble()+0.5);` but I am not sure if it matters. At least for small doubles they probably convert cleanly to int.

peterNordin (approved):

SymHop/src/SymHop.cpp
robbr48: Good point. It's not very common to use very large integer exponents, but better be safe.

PR #2152, opened 2022-11-14
Do not add closing parenthesis when it has already been added

robbr48:
When typeing a left parenthesis in text editor, the right parenthesis is automatically added and the cursor is placed between them. If user types an extra end parenthesis it should be ignored and the cursor should be moved to after the closing parenthesis.

### DIFF
--- a/HopsanGUI/Widgets/TextEditorWidget.cpp
+++ b/HopsanGUI/Widgets/TextEditorWidget.cpp
@@ -794,6 +794,22 @@ void TextEditor::keyPressEvent(QKeyEvent* event)
             setTextCursor(cursor);
         }
     }
+    else if(event->key() == Qt::Key_ParenRight)
+    {
+        auto cursor = textCursor();
+        auto text = cursor.block().text();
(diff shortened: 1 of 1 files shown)


peterNordin:
Code looks OK, but why do we want this behavior?
I think I recognize it from some other software, but I cant remember which one.  

If I press ) I would expect a ) to be added. I can see the use if the code would count the number of ( and figure out if the ) is needed or not, but that will become complicated.

robbr48:
This is the default behaviour in both Qt Creator and Visual Studio.

Say for example that I want to type "y=f(x)". I would naturally press `y`+`=`+ `f`+`(`+`x`+`)`. But since the end parenthesis was already added automatically this would give "y=f(x))". If I forget to press backspace I get a compilation error later. The problem is that we don't normally look at the screen when we type, we use the muscle memory.

I have also compared the Hopsan behaviour to Qt Creator and I have not found any difference in behaviour.

peterNordin (approved):
OK, I noticed now that there was actually a description in the PR that mentioned the reason.

PR #2153, opened 2022-11-14
Remember original data when saving models with missing components

### LINKED ISSUE #1009: You should not be allowed to save a model where components are missing
---

Original Redmine Issue: 1009, https://flumes.iei.liu.se/redmine/issues/1009
Original Date: 2013/11/25

---

You should not be allowed to save a model where components are missing
An alternative would be to avoid overwriting the missing component data with nothing (since a dummy component does not have parameters and startvalues)
An other alternative would be to giva a warning message box, telling the user that all info from the following components will be lost

robbr48:
When loading a model with components from a missing library, the original XML data is stored in "fallback dom elements" for both the model object and the dangling connectors. These are then used when saving the model. This both elliminates the problem of accidentally saving a model with missing components and thereby losing data, and with the backup models being damaged during a failed library recompilation.

Resolves #1009.

### DIFF
--- a/HopsanGUI/loadFunctions.cpp
+++ b/HopsanGUI/loadFunctions.cpp
@@ -115,40 +115,35 @@ bool loadConnector(QDomElement &rDomElement, SystemObject* pContainer, UndoStatu
     Port *startPort = pContainer->getModelObjectPort(startComponentName, startPortName);
     Port *endPort = pContainer->getModelObjectPort(endComponentName, endPortName);
 
-    if (startPort && endPort)
+    Connector* pConn = pContainer->createConnector(startPort, endPort, NoUndo);
+    if (pConn)
     {
-        Connector* pConn = pContainer->createConnector(startPort, endPort, NoUndo);
-        if (pConn)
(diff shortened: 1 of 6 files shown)


peterNordin:
Wow, this is an old issue, finally getting a fix. Good!
I will need to look at this a bit more thoroughly, hopefully tomorrow.

HopsanGUI/loadFunctions.cpp
```diff
@@ -115,41 +115,38 @@ bool loadConnector(QDomElement &rDomElement, SystemObject* pContainer, UndoStatu
     Port *startPort = pContainer->getModelObjectPort(startComponentName, startPortName);
     Port *endPort = pContainer->getModelObjectPort(endComponentName, endPortName);
 
-    if (startPort && endPort)
+    //if (startPort && endPort)
```
peterNordin, line 118: leftover comment

peterNordin (approved):
Works great!

robbr48:
The only potential problem is if we change the HMF format in the future, and someone loads a model in the old format with a missing component. Then the old XML code will be stored and used for saving. I doubt this will cause too much trouble though, and it can be solved with some format conversion utility.

PR #2157, opened 2022-11-21
Remove duplicate activity diagram source files

robbr48:
Remove duplicate component source files. We keep the ones that were specified in [ActivityDiagramLib.xml](https://github.com/robbr48/hopsan/blob/remove_duplicate_components_activity_diagram/componentLibraries/defaultLibrary/Signal/ActivityDiagram/ActivityDiagramLib.xml).


peterNordin (changes requested):
I dont think that the files, Makefile.Debug, Makefile.Release, ActivityDiagramLib.cc , ActivitiyDiagramLib.pro and compile.bat should be added.

I recommend `git add -u` instead of `git add .` which I guess is the reason these files were added.

peterNordin:
Other than that, OK

robbr48:
My bad. I have removed the extra files.

peterNordin:
ActivityDiagramLib.cc , ActivitiyDiagramLib.pro and compile.bat still remain, remove them also

Issue #1924, opened 2020-04-23
Long save times when saving to network drive


robbr48:
**Describe the bug**
On some systems saving a model to a network drive can be two orders of magnitude slower compared to saving locally. 

**Possible solutions**
It could work to always save the file locally and then copy it to the final location. Perhaps this can be automated from Hopsan. But I have also seen reports of QFile::copy [being slow](https://stackoverflow.com/questions/39485537/qt-qfile-copy-operation-extremely-slow).

**Additional information**
There is a [bug report](https://bugreports.qt.io/browse/QTBUG-16715) for Qt from 2011 regarding this. It is still unresolved.

PR #2169, opened 2023-04-25
Replaced xml saving using QTextStream with QFile writing byte array

peterNordin:
This seems to significantly speed up saving files on a Windows network drive


peterNordin:
Related to #1924

robbr48 (approved):
Works really well and I have not noticed any problems!

PR #2174, opened 2023-05-31
Support dot before exponential in HCOM

robbr48:
Fixes issue with HCOM commands like
```
x=14.e-3
```
It used to assume the "e" should have a number  before it. This fix extends this to also allow a dot.


SymHop/src/SymHop.cpp
```diff
@@ -384,7 +384,12 @@ void Expression::commonConstructorCode(QStringList symbols, bool &ok, const Expr
                              (i>1 && str.size() > i+2 && str.at(i-2).isNumber() && str.at(i-1) == 'e' && str.at(i) == '-' && str.at(i+1) == '+' && str.at(i+2).isNumber()) ||
                              (i>1 && str.size() > i+1 && str.at(i-2).isNumber() && str.at(i-1) == 'e' && str.at(i) == '+' && str.at(i+1).isNumber()) ||
                              (i>2 && str.size() > i+1 && str.at(i-3).isNumber() && str.at(i-2) == "e" && str.at(i-1) == '+' && str.at(i) == '-' && str.at(i+1).isNumber()) ||
-                             (i>2 && str.size() > i+1 && str.at(i-3).isNumber() && str.at(i-2) == "e" && str.at(i-1) == '-' && str.at(i) == '+' && str.at(i+1).isNumber())))     //End of variable, append it to symbols (last two checks makes sure that Xe+Y and Xe-Y are treated as one symbol)
+                             (i>2 && str.size() > i+1 && str.at(i-3).isNumber() && str.at(i-2) == "e" && str.at(i-1) == '-' && str.at(i) == '+' && str.at(i+1).isNumber()) ||
```
peterNordin, line 387: I am not a fan of this code, but I guess this is OK. Could you add a test case that test all of these cases.
That way it would be more clear which sort of expressions are supported in this case

Issue #2115, opened 2022-01-11
HCOM: Crash if trying to change non-existing parameter with same name as variable


robbr48:
**Describe the bug**
Program will crash if user tries to change a model parameter that does not exist, if a variable of the same name exists locally in HCOM.

**To Reproduce**
1. Open a new model without a system parameter named "x"
2. > x = 5
3. > chpa x 3
4. Program will crash

**Expected behavior**
Show some clever error message.

robbr48:
Seems like it crashes because the code tries to evaluate the parameter name by something called "component index". I have no idea what this code is supposed to do.
https://github.com/Hopsan/hopsan/blob/master/HopsanGUI/HcomHandler.cpp#L2119
```
//If no parameters found, try if component name can be evaluated by component index instead
if(parameterNames.isEmpty())
{
    evaluateExpression(splitCmd[0].section(".",0,0));
    if(mAnsType == Scalar)
    {
        QString compName = pSystem->getModelObjects().at(mAnsScalar)->getName();
        splitCmd[0].replace(0, splitCmd[0].section(".",0,0).size(), compName);
        getParameters(splitCmd[0], parameterNames);
    }
}
```

peterNordin:
I dont know either, get rid of it. 
Probably some ancient stuff, it is not possible for the user to expect at which index in the pointer vector a particular component resides.

peterNordin:
A git blame shows you as the author :)

robbr48:
Yes, I noticed :) It might have been related to my multicore benchmark models, which were autogenerated by HCOM and had very many components of the same type. But it is useless today, so I will remove it.

robbr48:
Resolved by #2225.

PR #2225, opened 2024-05-27
Remove local variables with same name when adding system parameters

robbr48:
If they are not removed, the following problem will occur:
```
>> y=3
Assigning scalar y with 3
>> adpa y 4
>> y
3
```
The value in the system parameters widget, however, was actually 4.


robbr48:
Resolves #2115.

PR #2251, opened 2024-12-10
Switch to cmake part1

peterNordin:
This is part one of switch over to CMake, has been in progress for more than three years.
After adding DCP additional work is needed to be able to build release, but i will make a part2

This PR enables me to build all of Hopsan on Linux GCC, Windows MinGW and Windows MSVC 2019.
Including with DCP. But I do not know if it actually works in all these cases. Hopsan at least starts.

Deb build has been migrated to CMake. But flatpak/snap and Windows builds remain.


hopsandcp/CMakeLists.txt
```diff
@@ -1,19 +1,19 @@
-cmake_minimum_required(VERSION 3.1)
 project(hopsandcp)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_DEBUG_POSTFIX _d)
 if(WIN32)
   set(CMAKE_SHARED_LIBRARY_PREFIX "")
 endif()
 
-set(CMAKE_AUTOMOC ON)
+include(${CMAKE_CURRENT_LIST_DIR}/../dependencies/dcplib.cmake)
```
robbr48, line 8: We should not include this, since DCPLib is header-only and does no longer require linking.

hopsandcp/CMakeLists.txt
```diff
@@ -1,19 +1,19 @@
...
+#include(${CMAKE_CURRENT_LIST_DIR}/../dependencies/xerces.cmake)
 
-# include(${CMAKE_CURRENT_LIST_DIR}/../dependencies/xerces.cmake)
 # include(${CMAKE_CURRENT_LIST_DIR}/../dependencies/libzip.cmake)
```
robbr48, line 11: We need to include libzip.cmake, however, because we must link to libzip directly below.

hopsandcp/CMakeLists.txt
```diff
@@ -25,22 +25,23 @@ add_library(hopsandcp STATIC ${srcfiles} ${headerfiles})
...
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/dcplib-code/include/core>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/dcplib-code/include/ethernet>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/dcplib-code/include/master>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/dcplib-code/include/slave>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/dcplib-code/include/xml>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/dcplib-code/include/zip>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/libzip/include>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/asio-code/include>
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/xerces/include>
+  # $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/dcplib-code/include/bluetooth>
```
robbr48, line 28: These include directories are needed, because we need to link to the headers in DCPLib, libzip, Asio and Xerces. But perhaps they should can be in dcplib.cmake if we remove all stuff with linking to the library.

hopsandcp/CMakeLists.txt
```diff
@@ -25,22 +25,23 @@ add_library(hopsandcp STATIC ${srcfiles} ${headerfiles})
...
+  #$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/libzip/include>
+  #$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/asio-code/include>
+  #$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../dependencies/xerces/include>
   $<INSTALL_INTERFACE:include>)
 
-if(WIN32)
-    target_link_libraries(hopsandcp hopsancore libzip -lws2_32)
-else()
-    target_link_libraries(hopsandcp hopsancore libzip)
+target_link_libraries(hopsandcp hopsancore DCPLib::Master DCPLib::Slave DCPLib::Zip DCPLib::Ethernet)
```
robbr48, line 40: Remove: DCPLib::Master, DCPLib::Slave and DCPLib::Zip
Add: libzip

robbr48 (changes requested):
Seems to work except for the DCPLib. It no is no longer compiled and linked since it is header-only. See separate comments in hopsandcp/CMakeLists.txt.

The conflict is just because I fixed the same typo, so it is very easy to fix.

hopsandcp/CMakeLists.txt
peterNordin: I moved the libzip setup into dcplib.cmake since it was only a dependency to dcplib, ..... I think.

hopsandcp/CMakeLists.txt
peterNordin: Those are the targets defined by the dcplib project, so I wanted to "link" (depend) on them instead of explicitly pointing to the header directories. There may be more properties in these targets (compiler definitions and such things) that may be required. I think that DCPLib::Zip depends on libzip

robbr48:
@peterNordin I would want to make a release before summer. Is this ready to be merged?

robbr48 (commented):
It works on Ubuntu. On Windows it compiles, but when I try to run from within Qt Creator it loops for the executables in the wrong folder. I was able to run Hopsan outside of Qt, but for some reason all default components failed to load.

peterNordin:
@robbr48  It did seem to work last time a was working on it, I tried it on linux and windows mingw / msvc, But I cant be sure that it works in all cases. Release package builds also need to be verified and tested. They looked fine, but its hard to know for sure, I cant test all functions. I just dont have time.

But I will at least compare the contents of the new installer packages with the latest release to make sure that all files are included.


Your problem with running in QtCreator,  which compiler. Mingw or MSVC  build?


Anyway I rebased it on your latest release, and now everything is broken again. Mostly due to ssp4c build and config needing the same changes as the other dependencies, but it takes a while to work through. And ofcourse CI systems have changed the images again which causes additional problems....

I am trying to make it build for the flathub kde runtime update. After that we can (must) merge this branch.
It kind of builds now, but seem to be very broken on Windows due to zlib lookup, but I think its the same problem as the other dependencies had. I just have to remember how to fix it.

Snap will be updated later in part2

PR #2312, opened 2026-03-15
Update some thirdparty libraries

peterNordin:
So that Hopsan builds on flathub


robbr48 (approved):
Seems to work. I was able to compile and run without problems. It fails on MSVC for some reason though, perhaps the filename of the xerces dll is wrong?

peterNordin:
merged even though msvc fails

